PR #230, opened 2014-12-15
Add eventlet reactor

justinsb:
Apache licensed, from MagnetoDB project

https://github.com/stackforge/magnetodb/blob/master/magnetodb/common/cassandra/io/eventletreactor.py

### DIFF
--- a/cassandra/io/eventletreactor.py
+++ b/cassandra/io/eventletreactor.py
@@ -1,0 +1,195 @@
+# Copyright 2014 Symantec Corporation
+# Copyright 2013-2014 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import collections
+import eventlet
+from eventlet.green import select
+from eventlet.green import socket
+from eventlet import queue
+import errno
+import functools
+from six import moves
+import threading
+
+import logging
+import os
+
+import cassandra
+from cassandra import connection as cassandra_connection
+from cassandra import protocol as cassandra_protocol
+
+
+log = logging.getLogger(__name__)
+
+
+def is_timeout(err):
+    return (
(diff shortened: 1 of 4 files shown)


cassandra/io/eventletreactor.py
```diff
@@ -0,0 +1,213 @@
...
+                buf = self._socket.recv(self.in_buffer_size)
+                self._iobuf.write(buf)
+            except green.socket.error as err:
+                if not is_timeout(err):
+                    log.debug(
+                        "Exception during socket recv for %s: %s", self, err)
+                    self.defunct(err)
+                    return  # leave the read loop
+
+            if self._iobuf.tell():
```
aholmberg: Is there a reason this logic is here, in handle_read instead of left to connection.process_io_buffer?

cassandra/io/eventletreactor.py
```diff
@@ -0,0 +1,213 @@
...
+
+import eventlet
+from eventlet import green
+from eventlet import queue
+
+
+try:
+    from cStringIO import StringIO  # ignore flake8 warning: # NOQA
+except ImportError:
+    from StringIO import StringIO  # ignore flake8 warning: # NOQA
```
aholmberg: This doesn't work for python3. Suggest six.StringIO.
However, I'm also not sure why this connection uses an iobuffer distinct from connection._iobuf

cassandra/io/eventletreactor.py
```diff
@@ -0,0 +1,213 @@
...
+                            self._total_reqd_bytes = body_len + 8
+                            break
+            else:
+                log.debug("connection closed by server")
+                self.close()
+                return
+
+    def push(self, data):
+        chunk_size = self.out_buffer_size
+        for i in xrange(0, len(data), chunk_size):
```
aholmberg: Does not work in python3. Suggest six.moves xrange

aholmberg:
It looks like the original source may have been updated. Would it be worth incorporating that here?

cassandra/io/eventletreactor.py
ccstolley: Or better, consider io.BytesIO. six.StringIO maps to StringIO.StringIO in python2.x, which is slow. See also https://github.com/datastax/python-driver/commit/38306470bf0fc3f4499bd7e1c2d1eba987325647

aholmberg: Good note. I was actually trying that myself when the second point occurred to me.

justinsb:
Going to the latest version from magnetodb was a great idea - thanks!  I think they did indeed resolve the issues you identified (and thanks for the code review).

I added the following to the version found there:

* I changed send -> sendall
* I added eventlet_nosetests, which works the same way as gevent_nosetests.  eventlet, like gevent, needs early monkey-patching.

aholmberg:
Thanks for the quick turnaround, and for including the test mechanism.
Looking at this now.

PR #234, opened 2015-01-05
Properly handle unhexlify argument type and return type in Python 3

kandul:
When the schema contains user defined types, binascii.uhexlify gets called.
In Python 3 unhexlify accepts "only bytestring or bytearray objects as input"
and returns bytes too. This caused apply_parameters to crash.
Now for Python 3 there'a a wrapper than accepts str or bytes and returns str.

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -59,6 +59,18 @@ else:
     _number_types = frozenset((int, long, float))
 
 
+if six.PY3:
+    def _unhexlify(s):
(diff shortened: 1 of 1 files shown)


aholmberg:
Thanks for the PR. Created for planning and test: https://datastax-oss.atlassian.net/browse/PYTHON-208

aholmberg:
Do you happen to have an example schema and script that showed this happening?

aholmberg:
Nevermind. It was easy enough to reproduce.

PR #237, opened 2015-01-07
PYTHON-205 - Build Schema on Disagreement

aholmberg:


cassandra/cluster.py
```diff
@@ -1044,14 +1045,26 @@ def _ensure_core_connections(self):
...
+        """
+        Synchronously refresh the schema metadata.
+        By default timeout for this operation is governed by :attr:`~.Cluster.max_schema_agreement_wait`
+        and :attr:`~.Cluster.control_connection_timeout`.
+        Passing schema_agreement_wait here overrides :attr:`~.Cluster.max_schema_agreement_wait`.
+        Setting schema_agreement_wait <= 0 will bypass schema agreement and refresh schema immediately.
+        RuntimeWarning is raised if schema refresh fails for any reason.
+        """
+        if not self.control_connection.refresh_schema(keyspace, table, usertype, schema_agreement_wait):
+            raise RuntimeWarning("Schema was not refreshed. See log for details.")
```
thobbs: ``RuntimeWarning`` (or really, any subclass of ``Warning``) isn't really intended for this type of use.  I would either raise a generic ``Exception`` or make a special exception class for this.

thobbs: Also, ``ControlConnection.refresh_schema()`` can raise other Exceptions as well, so the docstring isn't quite correct.

aholmberg: I don't think it raises Exceptions. They are all caught and logged.
Regardless, good input on the warning. Updated.

PR #240, opened 2015-01-13
PYTHON-197 - Idle connection heartbeat

aholmberg:


cassandra/connection.py
```diff
@@ -732,3 +743,90 @@ def deliver(self, timeout=None):
...
+        Thread.__init__(self, name="Connection heartbeat")
+        self._interval = interval_sec
+        self._get_connection_holders = get_connection_holders
+        self._shutdown_event = Event()
+        self.daemon = True
+        self.start()
+
+    def run(self):
+        elapsed = 0
+        while not self._shutdown_event.wait(self._interval - elapsed):
```
thobbs: In Python 2.6, Event.wait() always returns None (even if set() is called on the Event).

cassandra/connection.py
```diff
@@ -732,3 +743,90 @@ def deliver(self, timeout=None):
...
+    def __init__(self, connection, owner):
+        self._exception = None
+        self._event = Event()
+        self.connection = connection
+        self.owner = owner
+        log.debug("Sending options message heartbeat on idle connection %s %s",
+                  id(connection), connection.host)
+        with connection.lock:
+            connection.send_msg(OptionsMessage(), connection.get_request_id(), self._options_callback)
+            connection.in_flight += 1
```
thobbs: You need to check that ``in_flight`` is less than ``max_request_id`` before getting a new request ID.  (See ``wait_for_responses`` for one method of doing this.)

cassandra/cluster.py
```diff
@@ -721,18 +741,17 @@ def shutdown(self):
...
-                session.shutdown()
+        for session in self.sessions:
+            session.shutdown()
+
+        self.executor.shutdown()
 
-        if self.executor:
-            self.executor.shutdown()
+        if self._idle_heartbeat:
+            self._idle_heartbeat.stop()
```
thobbs: It shouldn't make a big difference, but I would stop this before closing the control conn and sessions.

cassandra/cluster.py
```diff
@@ -372,6 +373,14 @@ def auth_provider(self, value):
     If set to :const:`None`, there will be no timeout for these queries.
     """
 
+    idle_heartbeat_interval = 30
```
thobbs: I don't know if there has been discussion of the default value for this somewhere, but 30s feels a little high to me, especially for the control connection (since the driver may miss DOWN notifications while the control conn has failed).

cassandra/connection.py
```diff
@@ -732,3 +743,90 @@ def deliver(self, timeout=None):
...
+
+    def wait(self, timeout):
+        if self._event.wait(timeout):
```
thobbs: Same note on Event.wait() always returning None in Python 2.6.  See ResponseWaiter.deliver() for an example of handling this.

cassandra/connection.py
thobbs: Also, ``self._interval - elapsed`` could be a negative number.  I'm not sure if ``wait()`` handles that correctly, so you may want to do max(0.01, ...).

cassandra/connection.py
```diff
@@ -732,3 +743,90 @@ def deliver(self, timeout=None):
...
+                    except Exception:
+                        log.warning("Heartbeat failed for connection (%s) to %s",
+                                    id(connection), connection.host, exc_info=True)
+                        failed_connections.append((f.connection, f.owner))
+
+                for connection, owner in failed_connections:
+                    connection.defunct(Exception('Connection heartbeat failure'))
+                    owner.return_connection(connection)
+            except Exception:
+                log.warning("Failed connection heartbeat", exc_info=True)
```
thobbs: This seems like an unexpected error (unless I'm missing something), so I would do log.error() here.

cassandra/connection.py
```diff
@@ -673,6 +677,13 @@ def process_result(result):
 
         self.send_msg(query, request_id, process_result)
 
+    @property
+    def is_idle(self):
+        return self.in_flight == 0 and not self.msg_received
```
thobbs: Hmm, I'm not completely sure that the heartbeat should only be issued if self.in_flight is 0.  For example, if a switch started dropping packets, an operation could hang (with in_flight > 0) for quite some time.  The user would get an OperationTimedOut() for an operation, but the connection wouldn't necessarily be defuncted right away.  The heartbeat timeing out could help to defunct the connection more quickly in that case if the in_flight check is removed.

cassandra/connection.py
aholmberg: Thanks for the note on 2.6. Was not aware of that. I did test negative numbers -- works as if timeout == 0.

cassandra/connection.py
aholmberg: I was relying on the fact that an 'idle' connection would not have a bunch in_flight. Based on your later comment about the definition of is_idle, that could change. Will update.

cassandra/cluster.py
aholmberg: I don't think there's a lot of science behind it, but the general notion that it should be sufficiently high that there is little to no chance of false negatives (this value taken to align with other drivers' default settings). We don't want a bunch of connection thrashing 'by default'. 
Many modes of connection failure would detect the problem on send, which would be much more responsive. 
WRT missing events, we can at least rely on the new connection to refresh topology when it comes up.
Of course users are free to tighten this up if it suits their environment.

cassandra/connection.py
aholmberg: Good point. Updated.

cassandra/cluster.py
thobbs: > WRT missing events, we can at least rely on the new connection to refresh topology when it comes up.

That's true, but since the control connection doesn't regularly execute queries, it will in general be down until the next heartbeat fails.  It might be worth using a different heartbeat period for the control connection.  It probably wouldn't need a config option, either, since "accidentally" closing the control conn shouldn't have much of an effect on the application.  Unfortunately, this would slightly complicate the current heartbeat design.

But, overall, I think this default value would be reasonable to change in a future release if we see that it's not optimal in practice, so getting it perfect right now isn't crucial.

aholmberg: I understand the desire to make the CC more responsive. It would actually be pretty easy to refactor and put a dedicated thread with its own interval on CC. That would be the cleanest, design-wise. I didn't do that in this iteration because I wasn't convinced it was worth the overhead.
Another option would be to run the loop at a higher frequency and only hit connections with is_control_connection set.
Since it's a late in the dev cycle I might propose this as a follow-on improvement.

cassandra/connection.py
```diff
@@ -732,3 +743,96 @@ def deliver(self, timeout=None):
...
+        self.connection = connection
+        self.owner = owner
+        log.debug("Sending options message heartbeat on idle connection %s %s",
+                  id(connection), connection.host)
+        with connection.lock:
+            if connection.in_flight < connection.max_request_id:
+                connection.in_flight += 1
+                connection.send_msg(OptionsMessage(), connection.get_request_id(), self._options_callback)
+            else:
+                self._exception = Exception("Failed to send heartbeat because connection 'in_flight' exceeds threshold")
```
thobbs: I think this may be problematic for the v1/v2 protocol under load, because connections can get maxed out pretty easily.  Here are a few alternatives:
* Wait for in_flight to drop in a bounded loop with sleeps.  This is kind of expensive, but should happen pretty quickly with a healthy connection.
* Skip the heartbeat for this round.  If in_flight is maxed out because the connection is healthy, this is fine.  If it's maxed out because none of the operations are completing, then, well ... skipping the round sucks.  So I guess this option isn't great.
* Always reserve a spare request ID for heartbeats (by lowering the max for other connection users).  This is probably the best option, but requires a bit more work.

aholmberg: Considering that we only heartbeat if a connection appears idle, arriving here and finding no request id would indicate that there are max_request_id requests out at this very instant, while no responses have been seen on this connection for the last idle_heartbeat_interval seconds.
I think in a healthy, high load situation we would not be here. I can imagine wanting to wait is if the idle_heartbeat_interval is < normal request latency, but that would be a silly configuration.

cassandra/cluster.py
thobbs: +1, either of those would be a good follow-up.

cassandra/connection.py
thobbs: Oh, good point, I forgot about normal messages reseting is_idle.  Whoops :)

cassandra/cluster.py
aholmberg: https://datastax-oss.atlassian.net/browse/PYTHON-214

PR #242, opened 2015-01-15
PYTHON-186 - Support for nested collections

aholmberg:


cassandra/util.py
```diff
@@ -555,3 +555,76 @@ def _intersect(self, other):
...
+    def __eq__(self, other):
+        if isinstance(other, OrderedMap):
+            return self._items == other._items
+        try:
+            d = dict(other)
+            return len(d) == len(self._items) and all(i[1] == d[i[0]] for i in self._items)
+        except KeyError:
+            return False
+        except TypeError:
+            pass
```
thobbs: Is this primarily to handle ``dict(other)`` failing?  If so, I think it would be better to just isolate the ``TypeError`` try/except to that and return ``False``.  As for ``TypeError``s during comparison, I'm not exactly sure what to do, but I think letting it be raised is better than hiding it behind ``NotImplemented``.

cassandra/util.py
```diff
@@ -555,3 +555,76 @@ def _intersect(self, other):
...
+        return NotImplemented
+
+    def __repr__(self):
+        return '%s[%s]' % (
+            self.__class__.__name__,
+            ', '.join("(%r, %r)" % (k, v) for k, v in self._items))
+
+    def __str__(self):
+        return '{%s}' % ', '.join("%r: %r" % (k, v) for k, v in self._items)
```
thobbs: For ``__str__()`` I would use ``%s`` instead of ``%r``.

cassandra/util.py
```diff
@@ -555,3 +555,76 @@ def _intersect(self, other):
...
+        except KeyError:
+            return False
+        except TypeError:
+            pass
+        return NotImplemented
+
+    def __repr__(self):
+        return '%s[%s]' % (
+            self.__class__.__name__,
+            ', '.join("(%r, %r)" % (k, v) for k, v in self._items))
```
thobbs: So, this isn't a strict rule, but in general ``__repr__`` should provide a string that could recreate the object with ``eval()``.  So maybe matching the constructor syntax (perhaps with kwargs) is a better option.

cassandra/util.py
```diff
@@ -555,3 +555,76 @@ def _intersect(self, other):
...
                         isect.add(item)
             return isect
+
+from collections import Mapping
+import six
+from six.moves import cPickle
+
+class OrderedMap(Mapping):
+    '''
+    An ordered map that accepts non-hashable types for keys.
```
thobbs: I would mention that it's not mutable as well.  Also, it would be good to include this class in the API docs.

cassandra/util.py
aholmberg: I was basically following the idea that if there is not a sane comparison, then "NotImplemented". It seems to be a common idiom ([here, for example](https://github.com/datastax/python-driver/blob/2.1.3/cassandra/util.py#L320-L323)). The reason I'm handling a typeerror instead of testing 'isinstance' is because I wanted to let the dict construct from anything possible (e.g., a sequence of pairs...).

cassandra/util.py
aholmberg: I designed it with that exact idea in mind. The list of pairs seemed to be the most robust for arbitrary types. Maybe there are some cases I didn't think of?

cassandra/util.py
aholmberg: Good thought. I've not added any new entities before. Is it just a matter of adding it to docs/api/index.rst and docs/api/cassandra/util.rst?

cassandra/util.py
aholmberg: Added

cassandra/util.py
thobbs: I agree that ``isinstance()`` isn't the right way to go.

I'll admit that I hadn't noticed ``NotImplemented`` being used in ``__eq__()`` anywhere before.  I'm not sure why WeakSet does that, honestly.  The main problem I see with that is that ``NotImplemented`` is truthy.  Simply returning ``False`` if there's no sane way to compare the types seems like a better option for all of the cases I can think of.

cassandra/util.py
thobbs: Oh, I think you're just missing the parens in ``%s[%s]``, then.  The list of pairs is a good way to go.

cassandra/util.py
aholmberg: This explains one rationale: http://stackoverflow.com/a/879005/20688
However, knowing what we know about this type, I could see giving that up and assuming other.__eq__(self) would not work either.

cassandra/util.py
aholmberg: Now I see what you mean. Thanks.

PR #241, opened 2015-01-13
Support new 'date' and 'time' types in Cassandra

aholmberg:
CASSANDRA-7523
PYTHON-190

PR #243, opened 2015-01-16
PYTHON-190 Updated date/time support

aholmberg:
Superset of https://github.com/datastax/python-driver/pull/241

Updated to map native types to cql types.

The integration test will probably not be merged until we find out what versions of C* will have these types to test against.

I had some general difficulty testing against this branch due to some apparently unrelated server-side errors.
https://github.com/josh-mckenzie/cassandra/compare/7523_squashed

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -75,12 +75,12 @@ def unix_time_from_uuid1(u):
     return (u.time - 0x01B21DD213814000) / 10000000.0
 
 
+DATETIME_EPOC = datetime.datetime(1970, 1, 1)
+
+
 def datetime_from_timestamp(timestamp):
-    if timestamp >= 0:
-        dt = datetime.datetime.utcfromtimestamp(timestamp)
-    else:
-        # PYTHON-119: workaround for Windows
-        dt = datetime.datetime(1970, 1, 1) + datetime.timedelta(seconds=timestamp)
+    # PYTHON-119: workaround for Windows
+    dt = DATETIME_EPOC + datetime.timedelta(seconds=timestamp)
     return dt
 
 
@@ -652,7 +652,7 @@ class SimpleDateType(_CassandraType):
     @staticmethod
     def deserialize(byts, protocol_version):
         timestamp = SimpleDateType.seconds_per_day * (uint32_unpack(byts) - 2 ** 31)
-        dt = datetime.datetime.utcfromtimestamp(timestamp)
+        dt = datetime_from_timestamp(timestamp)
         return datetime.date(dt.year, dt.month, dt.day)
 
 
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -48,15 +48,13 @@ from cassandra.marshal import (int8_pack, int8_unpack,
                                int32_pack, int32_unpack, int64_pack, int64_unpack,
                                float_pack, float_unpack, double_pack, double_unpack,
                                varint_pack, varint_unpack)
-from cassandra.util import OrderedMap, sortedset
+from cassandra.util import OrderedMap, sortedset, Time
 
 apache_cassandra_type_prefix = 'org.apache.cassandra.db.marshal.'
 
 
 if six.PY3:
     _number_types = frozenset((int, float))
-    _time_types = frozenset((int,))
-    _date_types = frozenset((int,))
     long = int
 
     def _name_from_hex_string(encoded_name):
@@ -64,8 +62,6 @@ if six.PY3:
         return bin_str.decode('ascii')
 else:
     _number_types = frozenset((int, long, float))
-    _time_types = frozenset((int, long))
-    _date_types = frozenset((int, long))
     _name_from_hex_string = unhexlify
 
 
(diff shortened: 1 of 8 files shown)


aholmberg:
@thobbs, hate to bother, but I would like to discuss this one a bit with you since it changes some encoder mappings.

josh-mckenzie:
Code changes on the python side look reasonable to me.  I rebased 7523_squashed yesterday so the -1 ref counts you were getting should be addressed.

Changing to using native types rather than namedtuple means I'm going to need to modify cqlshlib and cqlsh to use the table metadata and pass some more metadata information around while formatting (to differentiate if a long is a cql long or a hi-res time for instance).  While I was fond of the 1-1 cql type mapping namedtuple gave me, I can see why you'd want to keep the driver decoupled in that respect.  I assume that's why the change to native type from namedtuple?

thobbs:
> Changing to using native types rather than namedtuple means I'm going to need to modify cqlshlib and cqlsh to use the table metadata and pass some more metadata information around while formatting (to differentiate if a long is a cql long or a hi-res time for instance)

Really, we should put some effort into making result metadata available though the driver so that cqlsh can stop guessing or mucking with the Metadata.  Looking at the Metadata is problematic when aliases are used for selectors, in any case.  There's really nothing in the driver for that now, unfortunately, but I think it would be okay for cqlsh to have less-than-optimal formatting of a new type for a while until we have better options.

cassandra/cqltypes.py
```diff
@@ -625,6 +633,95 @@ def serialize(timeuuid, protocol_version):
...
+        date_time = datetime.datetime.strptime(v, SimpleDateType.date_format)
+        return datetime.date(date_time.year, date_time.month, date_time.day)
+
+    @staticmethod
+    def serialize(val, protocol_version):
+        # Values of the 'date'` type are encoded as 32-bit unsigned integers
+        # representing a number of days with "the epoch" at the center of the
+        # range (2^31). Epoch is January 1st, 1970
+        try:
+            shifted = (calendar.timegm(val.timetuple()) // SimpleDateType.seconds_per_day) + 2 ** 31
```
thobbs: Docs should note that datetimes that don't correspond exactly to a date are truncated (as opposed to being rounded).  We don't have great docs around the handling of various types yet, though.

thobbs:
Overall this looks pretty good to me.  No obvious problems stick out.

josh-mckenzie:
> it would be okay for cqlsh to have less-than-optimal formatting of a new type for a while until we have better options

In its current incarnation, dates will read something like "datetime.date(2014, 1, 1)" and time's will show as raw longs.  I'm fine with that as well so we don't hold up getting the types in - I'll update the cassandra branch to remove formatters for the new types until we get things ironed out on the driver side.

josh-mckenzie:
@aholmberg Cassandra branch should be updated now if you need to do more testing (was updated on 22nd - forgot to update this PR w/that).

Do you need anything more from me on the C* side or with the driver PR?

aholmberg:
Nothing further required at this point. Just need to get the time to work on it.
Do you have a 'fix in' version target for C* at this point? That would be useful for prioritization on the driver side. We have some competing priorities.

josh-mckenzie:
I'd prioritize this on the low-mid level of things.  There was talk of targeting 3.0 at some point but we settled on 2.1.X given the safety of the inclusion.  Right now we have it targeted in the 2.1 line and I'd prefer to get it in then but that's mostly just due to not wanting the code to atrophy too much.

aholmberg:
Since there is no built-in type that represents time-of-day to nanoseconds, I introduced a new type. This will make it easier to work with, and easy to discriminate in cqlsh.

lmkwyt, and I will update the tests and docs.

I have a patch to cqlshlib I can post, too.

josh-mckenzie:
Hm - I'm not getting notifications for updates on this thread.  I'll look into that.

Back to the topic at hand - the Time type inside cassandra/util.py looks solid.  If your cqlshlib patch applies to the github branch [here](https://github.com/josh-mckenzie/cassandra/compare/7523_squashed) I can just apply it with a rebase.

aholmberg:
Yes, I was working from your branch. Patch here: https://gist.github.com/aholmberg/76c534d6e0ba86e3dc46

josh-mckenzie:
Current method of deserialization + formatting doesn't display dates before epoch due to restrictions on [utcfromtimestamp](https://docs.python.org/2/library/datetime.html#datetime.datetime.utcfromtimestamp).  Are we comfortable with that?

Your patch for cqlsh has been merged with the C* branch and I've rebased to current 2.1 head.

aholmberg:
We should be able to make that work. This mess should take care on the
deserialization side:
https://github.com/datastax/python-driver/blob/PYTHON-190/cassandra/cqltypes.py#L78-L84

Can you shoot me an example?

josh-mckenzie:
Need to change cqltypes.py:SimpleDateType.deserialize to read:
```python
    @staticmethod
    def deserialize(byts, protocol_version):
        timestamp = SimpleDateType.seconds_per_day * (uint32_unpack(byts) - 2 ** 31)
        dt = datetime_from_timestamp(timestamp)
        return datetime.date(dt.year, dt.month, dt.day)
```
Currently it's using utcfromtimestamp, which bombs on pre-epoch values.

aholmberg:
@thobbs did you care to provide further input here? If not, I think this is ready to go in.

thobbs:
+1, looks good to me

PR #244, opened 2015-01-16
PYTHON-195 - Return connection to pool after set keyspace

aholmberg:
PYTHON-195
Fixes a problem where connection.in_flight count leaks when keyspace
changes due to session.set_keyspace or 'USE' statements.


mpenick:
+1 Good find. Can't believe this wasn't found before now.

PR #247, opened 2015-01-20
PYTHON-194 - Eventlet reactor/connection

aholmberg:
Builds on #230


justinsb:
Awesome - thank you for the speedy merge and for the additional fixes!

justinsb:
Question - any guesses on when the next release will be made, i.e. when this will be available in PyPI?

We're trying to decide whether to keep a copy of the eventlet driver in the Ceilometer project (if the next release is not imminent), or whether we should just wait for the next release.

(Per comment here https://review.openstack.org/#/c/135630/)

It's not a big problem to keep a copy of the eventlet driver until this does get released, but if a release is coming any day now then we'd rather not!

aholmberg:
@justinsb I'm currently working on testing. Planning to roll the release next Monday, 1/26.

PR #249, opened 2015-01-30
enable collections inside UDTs via freezing Tuple

kishkaru:
Collection datatypes are allowed inside UDT's as the UDT itself is frozen, which freezes the collections inside. This is true for all collection types except for Tuple, which must be declared as frozen<> explicitly.

### DIFF
--- a/tests/integration/standard/test_udts.py
+++ b/tests/integration/standard/test_udts.py
@@ -580,7 +580,6 @@ class TypeTests(unittest.TestCase):
         """
         Test for inserting various types of DATA_TYPE_NON_PRIMITIVE into UDT's
         """
-        raise unittest.SkipTest("Collections are not allowed in UDTs")
         c = Cluster(protocol_version=PROTOCOL_VERSION)
         s = c.connect()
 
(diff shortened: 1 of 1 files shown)


tests/integration/standard/test_udts.py
```diff
@@ -580,7 +580,6 @@ def test_nonprimitive_datatypes(self):
         """
         Test for inserting various types of DATA_TYPE_NON_PRIMITIVE into UDT's
         """
-        raise unittest.SkipTest("Collections are not allowed in UDTs")
```
aholmberg: I think this still needs to skip for earlier versions.

tests/integration/standard/test_udts.py
```diff
@@ -599,6 +598,9 @@ def test_nonprimitive_datatypes(self):
                 if nonprim_datatype == "map":
                     type_string = "{0}_{1} {2}<{3}, {3}>".format(chr(start_index + i), chr(start_index + j),
                                                                  nonprim_datatype, datatype)
+                elif nonprim_datatype == "tuple":
+                    type_string = "{0}_{1} frozen<{2}<{3}>>".format(chr(start_index + i), chr(start_index + j),
+                                                            nonprim_datatype, datatype)
```
aholmberg: > [Jan-30 3:28 PM] Tyler Hobbs: yes, although requiring ``frozen<tuple<...>>`` inside a UDT was essentially a bug, and should be fixed in 2.1.3

Do I want this in here if it's going to be 'fixed' in the next rev?

tests/integration/standard/test_udts.py
kishkaru: There is a check for protocol v3 in the setUp: https://github.com/kishkaru/python-driver/blob/enable_collections_in_udt/tests/integration/standard/test_udts.py#L41 which applies to all tests under the TypeTests class.

tests/integration/standard/test_udts.py
kishkaru: FWIW, this test is currently running on the java-driver, and will be running on the ruby-driver. Even if it will be fixed in 2.1.3, explicitly having "frozen" is forwards-compatible for 2.1.3+

aholmberg:
Thanks for the update!

PR #253, opened 2015-03-06
PYTHON-231 - Normalized string encoding for OrderedMap Indexing

aholmberg:
Original problem statement in https://datastax-oss.atlassian.net/browse/PYTHON-231.

This change uses the internal cassandra type (when materializing a map from native protocol) to serialize map keys for the OrderedMap index.

Some notes about how I arrived here:
- JSON serialization had okay performance, but not very good support for serializing arbitrary types, and even some built-in types (set, for example)
- I looked at using cassandra.encoder.Encoder to serialize keys as CQL literals
-- Performance with quoting was not great, and removing quoting/escapes would lead to ambiguous serialized form for collections
-- This introduced a cyclic dependency Encoder-->util.OrderedMap-->Encoder that was awkward to extract
- Using the cassandra type
-- Makes materializing the index on protocol deserialization much more performant
-- Makes key serialization on lookups normalized, fully recursive, and efficient
-- Introduces an asymmetry between OrderedMaps created in the application context, and those materialized internally. However, the normalized serialization of the latter means comparison and indexing work as expected.
--- OrderedMaps created in the application context still have the ascii/unicode discrepancy, but I wouldn't expect lookup types to change in that context -- more concerned about normalized lookup for types returned from the server.

I welcome input from the reviewer.

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -780,8 +780,8 @@ class MapType(_ParameterizedType):
 
     @classmethod
     def validate(cls, val):
-        key_type, subvaltype = cls.subtypes
-        return dict((key_type.validate(k), subvaltype.validate(v)) for (k, v) in six.iteritems(val))
+        key_type, value_type = cls.subtypes
+        return dict((key_type.validate(k), value_type.validate(v)) for (k, v) in six.iteritems(val))
 
     @classmethod
     def deserialize_safe(cls, byts, protocol_version):
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -48,7 +48,7 @@ from cassandra.marshal import (int8_pack, int8_unpack,
                                int32_pack, int32_unpack, int64_pack, int64_unpack,
                                float_pack, float_unpack, double_pack, double_unpack,
                                varint_pack, varint_unpack)
-from cassandra.util import OrderedMap, sortedset, Time
+from cassandra.util import OrderedMapSerializedKey, sortedset, Time
 
 apache_cassandra_type_prefix = 'org.apache.cassandra.db.marshal.'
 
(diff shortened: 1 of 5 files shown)


cassandra/cqltypes.py
```diff
@@ -780,12 +780,12 @@ class MapType(_ParameterizedType):
 
     @classmethod
     def validate(cls, val):
-        subkeytype, subvaltype = cls.subtypes
-        return dict((subkeytype.validate(k), subvaltype.validate(v)) for (k, v) in six.iteritems(val))
+        key_type, subvaltype = cls.subtypes
```
mpenick: Should "subvaltype" be "value_type"?

tests/unit/test_policies.py
```diff
@@ -140,20 +140,21 @@ def host_down():
 
         # make the GIL switch after every instruction, maximizing
         # the chace of race conditions
```
mpenick: "chace"

cassandra/util.py
```diff
@@ -653,9 +674,8 @@ def __repr__(self):
     def __str__(self):
         return '{%s}' % ', '.join("%s: %s" % (k, v) for k, v in self._items)
 
-    @staticmethod
-    def _serialize_key(key):
-        return cPickle.dumps(key)
+    def _serialize_key(self, key):
+        return self.cass_key_type.serialize(key, self.protocol_version) if self.cass_key_type else cPickle.dumps(key)
```
mpenick: What about doing something like this?

```python
class OrderedMapKeyTypeSerialized(OrderedMap):
    def __init__(cass_key_type, protocol_version):
        self.cass_key_type = cass_key_type
        self.protocol_version = protocol_version

    def _insert_unchecked(self, key, flat_key, value):
        self._items.append((key, value))
        self._index[flat_key] = len(self._items) - 1

    def _serialize_key(self, key):
        return self.cass_key_type.serialize(key, self.protocol_version)
```

aholmberg: I like it. Thanks.

PR #254, opened 2015-03-09
PYTHON-99 - Time/TimeUUID Utility Functions

aholmberg:


cassandra/util.py
```diff
@@ -1,4 +1,128 @@
...
+    Converts a version 1 :class:`uuid.UUID` to a timestamp with the same precision
+    as :meth:`time.time()` returns.  This is useful for examining the
+    results of queries returning a v1 :class:`~uuid.UUID`.
+
+    :param uuid_arg: a version 1 :class:`~uuid.UUID`
+
+    :rtype: timestamp
+
+    """
+    return (uuid_arg.time - 0x01B21DD213814000) / 10000000.0
```
mpenick: why not use the original "1e7"? (easier to read IMO)

aholmberg: What do you mean by 'original'? I just moved it verbatim from cqltypes. 
(shrug)
In any event, I like 1e7. Will change before merging.

PR #257, opened 2015-03-11
PYTHON-213 - CQL export refinement for legacy tables

aholmberg:
Treat DCT like other unrecognized types
Improve CQL export for DCT tables
Do not print columns with invalid names
PYTHON-213

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -291,11 +291,7 @@ class _CassandraType(object):
             cname = apache_cassandra_type_prefix + cname
         if not subtypes:
             return cname
-        fieldnames = getattr(cls, 'fieldnames', None)
-        if fieldnames and any(fieldnames):
-            sublist = ', '.join('%s=>%s' % (field_name, subtype.cass_parameterized_type(full=full)) for (field_name, subtype) in zip(fieldnames, subtypes))
-        else:
-            sublist = ', '.join(styp.cass_parameterized_type(full=full) for styp in subtypes)
+        sublist = ', '.join(styp.cass_parameterized_type(full=full) for styp in subtypes)
         return '%s(%s)' % (cname, sublist)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


cassandra/cqltypes.py
```diff
@@ -291,7 +291,11 @@ def cass_parameterized_type_with(cls, subtypes, full=False):
             cname = apache_cassandra_type_prefix + cname
         if not subtypes:
             return cname
-        sublist = ', '.join(styp.cass_parameterized_type(full=full) for styp in subtypes)
+        fieldnames = getattr(cls, 'fieldnames', None)
+        if fieldnames and any(fieldnames):
```
thobbs: I'm guessing this is meant to only apply to DynamicCompositeType, but UserType also uses ``fieldnames``, so it seems like the wrong thing might happen here.

thobbs: That said, I'm not sure if that will ever actually matter.

cassandra/metadata.py
```diff
@@ -416,8 +420,8 @@ def _build_table_metadata(self, keyspace_metadata, row, col_rows, trigger_rows):
                 if value_alias_rows:  # CASSANDRA-8487
                     validator = types.lookup_casstype(value_alias_rows[0].get('validator'))
 
-            col = ColumnMetadata(table_meta, value_alias, validator)
-            if value_alias:  # CASSANDRA-8487
+            if value_alias and validator:  # CASSANDRA-8487
```
thobbs: Seems like ``validator`` could be uninitialized here, although that was also true in the old code.

aholmberg: Good catch. Luckily this was just a workaround for a C* bug, now resolved. Adding initialization.

cassandra/cqltypes.py
aholmberg: I'm a little cold on this change. I thought I had reasoned that fieldnames would only be set on this path for DCT, but it's a little hazy now. Instead of this, I can just specialize the method for DCT. I think that should be more clear.

aholmberg:
Metadata has moved on a great deal since this was created. I'm closing the PR and retaining the branch.

PR #262, opened 2015-04-01
Tag the version with the git hash if --tag-build is specified on the command line

stef1927:
A change to optionally tag release artifacts with the git hash as discussed in https://issues.apache.org/jira/browse/CASSANDRA-7814?focusedCommentId=14378205&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14378205.

Upon specifying --tag-build, the HEAD git hash would be appended to the version. This would be useful when we build internal versions of the driver that are committed to the Cassandra repository.

Conflicts:
	setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,9 +35,10 @@ from distutils.core import Extension
 from distutils.errors import (CCompilerError, DistutilsPlatformError,
                               DistutilsExecError)
 from distutils.cmd import Command
-
+from subprocess import check_output
 
 import os
+import shlex
 import warnings
 
 try:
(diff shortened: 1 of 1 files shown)


aholmberg:
Is this just for convenience? I'm not sure I like the overriding the --tag-build command line option. Would you be opposed to just using egg_info?:

    python setup.py egg_info -b-`git rev-parse --short HEAD` sdist --formats=zip

stef1927:
I wasn't aware of egg_info at all, that does the job nicely so we can ignore this. Thanks!

aholmberg:
Cool. I was not aware of it either, until you brought this to my attention. I happened upon it in the middle of this wall of text: http://svn.python.org/projects/sandbox/trunk/setuptools/setuptools.txt

PR #263, opened 2015-04-01
CASSANDRA-7814: Added indexes to table and ks metadata 

stef1927:
This is the suggested change for https://datastax-oss.atlassian.net/browse/PYTHON-241.

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -122,10 +122,9 @@ class Metadata(object):
             keyspace_col_rows = col_def_rows.get(keyspace_meta.name, {})
             keyspace_trigger_rows = trigger_rows.get(keyspace_meta.name, {})
             for table_row in cf_def_rows.get(keyspace_meta.name, []):
-                table_meta = self._build_table_metadata(
+                self._add_table_metadata_to_ks(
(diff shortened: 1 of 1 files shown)


cassandra/metadata.py
```diff
@@ -215,6 +216,20 @@ def _build_usertype(self, keyspace, usertype_row):
         return UserType(usertype_row['keyspace_name'], usertype_row['type_name'],
                         usertype_row['field_names'], type_classes)
 
+    def _add_table_metadata_to_ks(self, keyspace_metadata, row, col_rows, trigger_rows):
```
aholmberg: Wondering if this functionality would be better implemented in ``KeyspaceMetadata``, where the index is held (KeyspaceMetadata.add_table(self, table_meta), or something like that).

stef1927: Yes you are quite right, I moved both methods to the ks metadata, see latest commit.

aholmberg:
Please see updates in #284. Let me know what you think of those changes.

stef1927:
Changes look good and tests for CASSANDRA-7814 still pass, thanks!

PR #265, opened 2015-04-08
PYTHON-238 - New request exceptions in protocol v4

aholmberg:


mpenick:
Is there any value in adding additional unit tests for these new types in https://github.com/datastax/python-driver/blob/master/tests/unit/test_response_future.py?

aholmberg:
@mpenick I outlined techniques for integration tests [here](https://datastax-oss.atlassian.net/browse/PYTHON-238?focusedCommentId=21430&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21430). Those would probably cover the same paths, but it's a good idea to include as much as possible in unit tests. Will do.

PR #268, opened 2015-04-13
PYTHON-211 - UDF metadata in Cassandra 3.0

aholmberg:


aholmberg:
Not ready yet. Overlooked aggregates.

cassandra/metadata.py
```diff
@@ -140,7 +160,7 @@ def rebuild_schema(self, ks_results, type_results, cf_results, col_results, trig
                 self._keyspace_added(keyspace_meta.name)
 
         # remove not-just-added keyspaces
-        removed_keyspaces = [ksname for ksname in self.keyspaces.keys()
+        removed_keyspaces = [name for name in self.keyspaces.keys()
                              if ksname not in current_keyspaces]
```
mpenick: should "ksname" be "name" here?

cassandra/cluster.py
```diff
@@ -2073,6 +2097,24 @@ def _handle_results(success, result):
             log.debug("[control connection] Fetched user type info for %s.%s, rebuilding metadata", keyspace, usertype)
             types_result = dict_factory(*types_result.results) if types_result.results else {}
             self._cluster.metadata.usertype_changed(keyspace, usertype, types_result)
+        elif function:
+            # user defined function within this keyspace changed
+            where_clause = " WHERE keyspace_name = '%s' AND function_name = '%s' AND signature = [%s]" \
+                           % (keyspace, function.name, ','.join("'%s'" % t for t in function.type_signature))
```
mpenick: Would it be a useful feature to allow refreshing by function/aggregate "name" alone? if the "type_signature" is None it would refresh all functions/aggregates of the same name? Maybe this would collide with a function with no parameters. I say this because building the type signature could be difficult and providing the name allow would be more user friendly.

cassandra/metadata.py
aholmberg: Yes. Good catch. I'm not even sure why that changed. Reverting.

aholmberg: Remembered -- I think I was trying to silence a flake8 warning about a name collision

cassandra/cluster.py
aholmberg: I was mostly just using the information that comes in the schema_change event. I think it's a fine idea to make it optional for the ``Cluster.refresh_schema`` API. I've added a note to the ticket I have to split this up: https://datastax-oss.atlassian.net/browse/PYTHON-291

tests/integration/standard/test_metadata.py
```diff
@@ -928,3 +930,311 @@ def test_keyspace_alter(self):
...
+            super(FunctionTest.VerifiedFunction, self).__init__(test_case, Function, test_case.keyspace_function_meta, **kwargs)
+
+    class VerifiedAggregate(Verified):
+        def __init__(self, test_case, **kwargs):
+            super(FunctionTest.VerifiedAggregate, self).__init__(test_case, Aggregate, test_case.keyspace_aggregate_meta, **kwargs)
+
+
+class FunctionMetadata(FunctionTest):
+
+    def make_function_kwargs(self, deterministic=True, called_on_null=True):
```
GregBestland: Every integration test needs doxygen comments.

tests/integration/standard/test_metadata.py
```diff
@@ -928,3 +930,311 @@ def test_keyspace_alter(self):
...
+                'type_signature': ['double', 'int'],
+                'argument_names': ['d', 'i'],
+                'return_type': DoubleType,
+                'language': 'java',
+                'body': 'return new Double(0.0);',
+                'is_deterministic': deterministic,
+                'called_on_null_input': called_on_null}
+
+    def test_functions_after_udt(self):
+        self.assertNotIn(self.function_name, self.keyspace_function_meta)
```
GregBestland: Every test case need doxygen comments.

PR #274, opened 2015-04-17
Stabilize Jenkins on Cassandra 1.2 - 2.0

kishkaru:
Some extra fixes to stabilize older C* versions.


kishkaru:
The code was cleaner this way, and some other tests were already doing it this way (create a new keyspace per test, to avoid name conflicts). Another benefit of a keyspace per test is that for tests that are skipped, we don't have to do extra checks to see if a table exists before dropping it.

PR #275, opened 2015-04-17
added panda_factory for data retrieval. Uses DataFrame from pandas

aaronbenz:
Adds pandas to the stack for data retrieval purposes

### DIFF
--- a/cassandra/query.py
+++ b/cassandra/query.py
@@ -24,7 +24,7 @@ import re
 import struct
 import time
 import six
-
+from pandas import DataFrame
 from cassandra import ConsistencyLevel, OperationTimedOut
 from cassandra.util import unix_time_from_uuid1
 from cassandra.encoder import Encoder
@@ -50,6 +50,23 @@ def _clean_column_name(name):
         _clean_name_cache[name] = clean
         return clean
 
+def panda_factory(colnames, rows):
+    """
+    Returns all the rows as a singular panda DataFrame
+
+    Example::
+
+        >>> from cassandra.query import panda_factory
+        >>> session = cluster.connect("mykeyspace")
+        >>> session.row_factory = tuple_factory
+        >>> df = session.execute("SELECT name, age FROM users LIMIT 2")
+        >>>print df
+            name    age
+        0   Bob     42
+
+    """
+    clean_column_names = map(_clean_column_name, colnames)
+    return DataFrame(rows, columns=clean_column_names)
 
 def tuple_factory(colnames, rows):
     """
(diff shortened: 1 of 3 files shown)


aaronbenz:
any suggestions on using travis ci to get a successful build? Locally I created a brand new empty virtual envr and built successfully without any issues...

cassandra/query.py
```diff
@@ -50,6 +50,23 @@ def _clean_column_name(name):
...
 
+def panda_factory(colnames, rows):
+    """
+    Returns all the rows as a singular panda DataFrame
+
+    Example::
+
+        >>> from cassandra.query import panda_factory
+        >>> session = cluster.connect("mykeyspace")
+        >>> session.row_factory = tuple_factory
```
aholmberg: session.row_factory = panda_factory?

aholmberg:
Is there a reason you need this to be in the driver itself? I'm hesitant to include such a heavy library as a dependency, and it doesn't work for all the runtimes we support (pypy, for instance).

aaronbenz:
yeah, so that is really up to you guys, but here's where I'm coming from: My intention is to build support for pandas dataframes on the inbound and outbound. For example, to input into cassandra, you could have a dataframe with column names as the column names of the cassandra table, and then simply specify a keyspace and table name (or create an object that has those as attributes). If the column names were correct (verified through some prepared statement to also insert faster), then it inserts everything behind the scenes with a simple insert call. This little panda factory code was really just the start of that effort. Overall the goal would be to provide more direct support to data scientists as many of them heavily use pandas

StuartAxelOwen:
This sounds like something that would be better for a wrapper than being built in to the base.

aholmberg:
Interesting. Perhaps you could publish these utilities in a package as a
standalone project that depends on this one. (?) This work seems worthy of
sharing, but I would prefer not to introduce that dependency in the core.
I'm definitely interested in seeing your work progress. I would also
continue to encourage contributions of generic features enabling that
tooling.

aaronbenz:
I think you all are right. Ill start creating a separate cassandra python data science package. As a side bar, I am a heavy R user and would love to re-create the current rcassandra package that is available (as it was built on the old api, not cql)

StuartAxelOwen:
Been thinking more about this, and in looking at our own usages of the driver, I'm seeing a consistent pattern: passing the result rows into constructors to make them into model instances (with their methods and other bells/whistles).

Having an optional constructor parameter somewhere in the query process seems like a great interface for the kind of extensions @aaronbenz is pursuing, and would be very extensible for general purpose use.  

Having it built into the driver would be good, as it could ensure consistent results across each query pattern (async, sync, paged, etc).  The code change would probably be pretty small too.

Is this in line with the generic features you're talking about, @aholmberg?

PR #276, opened 2015-04-17
PYTHON-280 - Custom payload for protocol v4

aholmberg:


cassandra/query.py
```diff
@@ -701,6 +720,16 @@ def _maybe_set_routing_attributes(self, statement):
                 self.routing_key = statement.routing_key
                 self.keyspace = statement.keyspace
 
+    def _update_custom_payload(self, statement):
+        if statement.custom_payload:
+            if self.custom_payload is None:
+                self.custom_payload = {}
+            self.custom_payload.update(statement.custom_payload)
```
mpenick: Should this behavior be explicitly enabled? This seems like unexpected behavior to me.

mpenick:
Should "Cluster.protocol_version = 4" by default?

cassandra/cluster.py
```diff
@@ -2654,6 +2675,12 @@ def has_more_pages(self):
         """
         return self._paging_state is not None
 
+    @property
+    def custom_payload(self):
+        if not self._event.is_set():
+            raise Exception("custom_payload cannot be retrieved before ResponseFuture is finalized")
```
mpenick: Should this wait for the future to be set instead of throwing if not set? I shouldn't be that much more overhead. Maybe all I care about is the custom payload. :)

cassandra/query.py
aholmberg: My thinking was that statements would only have payloads on them if they were explicitly set. And if they're being added to the batch, they want those values (the sub-statements don't get to carry individual payloads).

This felt like the most flexible and easy to use. I don't feel too strongly about it since I doubt it will ever be used,

cassandra/cluster.py
aholmberg: Good question. If we wait for the future, then there has to be a timeout, which doesn't feel right for this property. This might be a good candidate for change once I get timers introduced and allow true request timeouts.

In the mean time, do you think it should just return what's there, regardless of whether it's been set yet (instead of throwing)? lmkwyt
Regardless, I should add a docstring that describes the semantics.

aholmberg:
Regarding protocol_version: it could probably change, but I'm also hoping to get protocol version negotiation in there for 3.0. This will probably go away in favor of a max protocol version at that point.

cassandra/query.py
mpenick: This makes sense.

PR #278, opened 2015-04-20
PYTHON-241 - Deprecate cqlengine.Float(double_precision=True)

aholmberg:
Overload deprecated in favor of distinct Float, Double types

PYTHON-241


aholmberg:
created with wrong issue branch

PR #279, opened 2015-04-20
PYTHON-246 - Deprecate cqlengine.Float(double_precision=True)

aholmberg:
Overload deprecated in favor of distinct Float, Double types

PYTHON-246


cassandra/cqlengine/columns.py
```diff
@@ -617,6 +608,31 @@ def to_database(self, value):
...
 
 
+class Float(BaseFloat):
+    """
+    Stores a single-precision floating-point value
+    """
+    db_type = 'float'
+
+    def __init__(self, double_precision=None, **kwargs):
+        if double_precision is None or bool(double_precision):
```
mpenick: Won't this is always print if double_precision is not provided? "is not None"?

aholmberg: Yes? I think that's doing what was intended. I'm trying to emulate the previous default (double_precision=True if omitted), and tell users to get away from ``Float`` if they really should be using ``Double``. If they had double_precision=False, they are (and should) be using float, and the warn is not emitted.

PR #286, opened 2015-04-22
PYTHON-297 - Make DCAware LBP tolerate DC changes during query plan

aholmberg:
Fixes a RuntimeError that would raise if the DCAwareRoundRobinPolicy
DC:host map changed during generation.

PYTHON-297


aholmberg:
This became d0ee3bc5470aea7fa67d5b80189e51cb68ec64b6 in the 2.5.1 release.

PR #287, opened 2015-04-23
PYTHON-298 - Remove race when adding new host during node/token rebuild

aholmberg:
Fixes an issue where a race could cause None return from
Cluster.add_host during node/token list rebuild.

This would cause None to be inserted into the Token map,
and subsequently blow up TokenAware LBP creating a query plan.

PYTHON-298


aholmberg:
This became https://github.com/datastax/python-driver/compare/d0ee3bc5470aea7fa67d5b80189e51cb68ec64b6...25f10a574c1cbb789eaeb19c16b30dfd92bb7d90 in the 2.5.1 release.

PR #292, opened 2015-04-27
Add trace_id to statement

carlyeks:


cassandra/cluster.py
```diff
@@ -2698,9 +2699,11 @@ def _set_result(self, response):
             if self._current_pool and self._connection:
                 self._current_pool.return_connection(self._connection)
 
-            trace_id = getattr(response, 'trace_id', None)
-            if trace_id:
-                self._query_trace = QueryTrace(trace_id, self.session)
+            self._query_trace_id = getattr(response, 'trace_id', None)
```
aholmberg: Curious why this was changed from local to class variable.

carlyeks: So that we have access to the trace id, even if we fail to get the trace events back before the timeout. That way, we can query any trace events we have for that trace_id.

aholmberg: I'm asking because you also assign it to the query, and it looks like that's the instance that gets used in your cqlsh patch.

carlyeks: Not really sure; pushed a new version which gets rid of the class variable.

aholmberg:
Thanks for the update.

PR #294, opened 2015-05-02
[PYTHON-246] tests for float/double updates in cqlengine

kishkaru:
Tests for PYTHON-246

### DIFF
--- a/tests/integration/cqlengine/model/test_model_io.py
+++ b/tests/integration/cqlengine/model/test_model_io.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from uuid import uuid4
+from uuid import uuid4, UUID
 import random
-from datetime import date
+from datetime import date, datetime
+from decimal import Decimal
 from operator import itemgetter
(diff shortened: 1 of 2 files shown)


tests/integration/cqlengine/model/test_model_io.py
```diff
@@ -124,6 +125,101 @@ def test_a_sensical_error_is_raised_if_you_try_to_create_a_table_twice(self):
...
+        class FloatingPointModel(Model):
+            id = columns.Integer(primary_key=True)
+            a = columns.Float(double_precision=False)
+            b = columns.Float(double_precision=True)
+            c = columns.Float()
+            d = columns.Double()
+
+        sync_table(FloatingPointModel)
+
+        with self.assertRaises(AttributeError):
```
aholmberg: It looks like this is failing in Travis. Why do you expect an attribute error here? 'a' is a valid attribute, and 'float' is double in Python. I think the only thing we can really validate here is that the DB type is correct. (?)

kishkaru:
Addressed the issue above, and squashed the commits.

PR #295, opened 2015-05-04
Limit the amount of iterations we take over rows

sontek:
This reduces the amount of times we iterate over rows and hopefully speeds up the deserialization process.

This was the before and after profiler with this change:

```
time python single.py 
102326

real	0m4.778s
user	0m2.045s
sys	0m0.049s
```

before

and after:

```
time python single.py 
102326

real	0m4.302s
user	0m1.881s
sys	0m0.050s
```

after

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2366,9 +2366,11 @@ class ControlConnection(object):
 
         versions = defaultdict(set)
         if local_result.results:
-            local_row = dict_factory(*local_result.results)[0]
-            if local_row.get("schema_version"):
-                versions[local_row.get("schema_version")].add(local_address)
+            local_row = dict_factory(*local_result.results)
+            if local_row:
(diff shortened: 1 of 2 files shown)


sontek:
Hmm, although it improved the profiler/callstack significantly I'm not seeing too much actual performance difference on it, here is my 2 million row dataset:

Master:

```
time python multi_hour.py 
real	0m9.363s
user	0m22.062s
sys	0m5.668s
```

With my change:
```
time python multi_hour.py 
real	0m9.449s
user	0m21.094s
sys	0m5.743s
```

aholmberg:
Interesting. That actual traversal of rows is most likely not too much overhead. Maybe the change to generator just skews the profile by changing *where* the work is done. Still, could be good to remove the extra iteration.

sontek:
I don't think this is necessary anymore

PR #298, opened 2015-05-06
PYTHON-108 - Timers

aholmberg:
Full request timers

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2600,7 +2600,8 @@ class ResponseFuture(object):
         self._start_timer()
 
     def _start_timer(self):
-        self._timer = self.session.cluster.connection_class.create_timer(self.timeout, self._on_timeout)
+        if self.timeout is not None:
+            self._timer = self.session.cluster.connection_class.create_timer(self.timeout, self._on_timeout)
 
     def _cancel_timer(self):
         if self._timer:
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3010,7 +3010,7 @@ class ResponseFuture(object):
             if self._paging_state is None:
                 return self._final_result
             else:
-                return PagedResult(self, self._final_result, timeout)
+                return PagedResult(self, self._final_result)
         else:
             raise self._final_exception
 
@@ -3162,10 +3162,9 @@ class PagedResult(object):
 
     response_future = None
 
-    def __init__(self, response_future, initial_response, timeout=_NOT_SET):
+    def __init__(self, response_future, initial_response):
         self.response_future = response_future
         self.current_response = iter(initial_response)
-        self.timeout = timeout
 
     def __iter__(self):
         return self
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1401,17 +1401,14 @@ class Session(object):
         trace details, the :attr:`~.Statement.trace` attribute will be left as
         :const:`None`.
         """
-        if timeout is _NOT_SET:
-            timeout = self.default_timeout
-
         if trace and not isinstance(query, Statement):
             raise TypeError(
                 "The query argument must be an instance of a subclass of "
                 "cassandra.query.Statement when trace=True")
 
-        future = self.execute_async(query, parameters, trace)
+        future = self.execute_async(query, parameters, trace, timeout)
         try:
-            result = future.result(timeout)
+            result = future.result()
         finally:
             if trace:
                 try:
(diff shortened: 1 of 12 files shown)


cassandra/cluster.py
```diff
@@ -2596,6 +2598,18 @@ def __init__(self, session, message, query, default_timeout=None, metrics=None,
...
+    def _start_timer(self):
+        if self.timeout is not None:
+            self._timer = self.session.cluster.connection_class.create_timer(self.timeout, self._on_timeout)
+
+    def _cancel_timer(self):
+        if self._timer:
+            self._timer.cancel()
+
+    def _on_timeout(self):
+        self._set_final_exception(OperationTimedOut())
```
thobbs: The ``OperationTimedOut()`` should be populated with ``_errors`` and ``_current_host``, I think.

cassandra/connection.py
```diff
@@ -871,3 +892,69 @@ def stop(self):
...
+
+    def __lt__(self, other):
+        return self.end < other.end
+
+    def cancel(self):
+        self.callback = self._noop
+        self.cancelled = True
+
+    def finish(self, time_now):
+        if self.callback is self._noop or time_now >= self.end:
```
thobbs: Why not immediately return ``True`` if ``self.callback is self._noop``?

cassandra/io/asyncorereactor.py
```diff
@@ -86,24 +88,22 @@ def maybe_start(self):
...
         log.debug("Starting asyncore event loop")
         with self._loop_lock:
-            while True:
+            while not self._shutdown:
                 try:
-                    asyncore.loop(timeout=0.001, use_poll=True, count=1000)
+                    asyncore.loop(timeout=0.001, use_poll=True, count=100)
+                    self._timers.service_timeouts()
+                    if not asyncore.socket_map:
+                        time.sleep(0.005)
```
thobbs: I presume this is to avoid burning CPU while there are no conns?  Do you plan to switch to a non-global map instead of ``socket_map``?

cassandra/io/eventletreactor.py
```diff
@@ -53,19 +52,39 @@ class EventletConnection(Connection):
...
+        cls._new_timer.set()
+        return timer
+
+    @classmethod
+    def service_timeouts(cls):
+        timer_manager = cls._timers
+        while True:
+            next_end = timer_manager.service_timeouts()
+            sleep_time = max(next_end - time.time(), 0) if next_end else 10000
+            cls._new_timer.wait(sleep_time)
```
thobbs: A comment to explain roughly how this loop works would be good -- took me a minute to figure out.

cassandra/io/twistedreactor.py
```diff
@@ -133,6 +135,25 @@ def _cleanup(self):
                             "Cluster.shutdown() to avoid this.")
             log.debug("Event loop thread was joined")
 
+    def add_timer(self, timer):
+        self._timers.add_timer(timer)
+        reactor.callFromThread(self._schedule_timeout, timer.end)
```
thobbs: Why ``callFromThread()``?

thobbs:
Since this fixes calling of callbacks/errbacks on timeout, the ``ResponseFuture`` docs need to be updated.

cassandra/connection.py
aholmberg: Good question. Actually this noop concept is residual from an earlier iteration where the time comparison was in a different context. I think I can do away with noop and just use a flag
Will update.

cassandra/io/asyncorereactor.py
aholmberg: I think it would be an alright idea to use a non-global map. I wasn't going to update that in this changeset since I really just need this loop to stay alive, global connections or not. I was staying with this to avoid having to manage a map at this level. If you think it's worth removing the dependency on socket_map in that module, I can look into it more.

cassandra/io/twistedreactor.py
aholmberg: To my knowledge this is the only way to interact with the loop thread in a safe way (similar to async events in libev). I callFromThread to get into the scheduling function with the loop thread, where it is safe to modify the loop timer.
If there are better techniques I'm open to suggestions.

cassandra/io/asyncorereactor.py
thobbs: It's probably not worth a lot of effort to use a non-global map.  The only thing I would maybe check is whether that attr is present and used in Jython.  (I know that Jython isn't really supported right now, but 2.7 was just released, so it would be good to know if there's obvious breakage.)

cassandra/io/twistedreactor.py
thobbs: Okay, that makes sense to me and is what I expected.  I would just add a comment to make that clear.

aholmberg:
I believe all comments have been addressed. Pending reviewer concurrence.

PR #304, opened 2015-05-12
PYTHON-244 - cqle: Protect ks/cf identifiers that use keywords

aholmberg:
PYTHON-244

### DIFF
--- a/cassandra/cqlengine/management.py
+++ b/cassandra/cqlengine/management.py
@@ -523,7 +523,7 @@ def drop_table(model):
 
     try:
         meta.keyspaces[ks_name].tables[raw_cf_name]
-        execute('drop table {};'.format(model.column_family_name()))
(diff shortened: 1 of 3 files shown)


cassandra/cqlengine/management.py
```diff
@@ -520,11 +519,11 @@ def drop_table(model):
     meta = get_cluster().metadata
 
     ks_name = model._get_keyspace()
-    raw_cf_name = model.column_family_name(include_keyspace=False)
+    raw_cf_name = model._raw_column_family_name()
 
     try:
         meta.keyspaces[ks_name].tables[raw_cf_name]
-        execute('drop table {};'.format(model.column_family_name(include_keyspace=True)))
+        execute('drop table {};'.format(model.column_family_name()))
```
mpenick: "DROP TABLE" ? More consistent with the rest of the code.

PR #310, opened 2015-05-14
PYTHON-291 - Deprecate Cluster.refresh_schema and submit_refresh_schema

aholmberg:
PYTHON-291
Use Cluster.refresh_*_metadata instead

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1202,7 +1202,7 @@ class Cluster(object):
         if not self.control_connection.refresh_schema(keyspace, table, schema_agreement_wait=max_schema_agreement_wait):
             raise Exception("Table metadata was not refreshed. See log for details.")
 
-    def refresh_type_metadata(self, keyspace, user_type, max_schema_agreement_wait=None):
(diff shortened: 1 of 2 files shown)


cassandra/cluster.py
```diff
@@ -1142,18 +1145,72 @@ def refresh_schema(self, keyspace=None, table=None, usertype=None, max_schema_ag
...
+        """
+        Synchronously refresh table metadata. This applies to a table, and any triggers or indexes attached
+        to the table.
+
+        See :meth:`~.Cluster.refresh_schema_metadata` for description of ``max_schema_agreement_wait`` behavior
+        """
+        if not self.control_connection.refresh_schema(keyspace, table, schema_agreement_wait=max_schema_agreement_wait):
+            raise Exception("Table metadata was not refreshed. See log for details.")
+
+    def refresh_type_metadata(self, keyspace, user_type, max_schema_agreement_wait=None):
```
mpenick: Maybe this should be "refresh_user_type_metadata" as not to be confused with other type?

PR #311, opened 2015-05-15
Add support for tinyint and smallint types

blerer:
CASSANDRA-8951 add support on the server side for tinyint and smallint types. This patch add support on the driver side.


aholmberg:
Thanks for the update. Merged and tweaked unit tests: 9b940d2

PR #312, opened 2015-05-15
Python 280 test

GregBestland:
Tests cases for python-280 and formatting cleanup

### DIFF
--- a/tests/integration/long/test_custom_payload.py
+++ b/tests/integration/long/test_custom_payload.py
@@ -1,0 +1,201 @@
+# Copyright 2013-2015 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


tests/integration/long/test_custom_payload.py
```diff
@@ -0,0 +1,201 @@
...
+    def test_custom_query_basic(self):
+        """
+        Test to validate that custom payloads work with simple queries
+
+        creates a simple query and ensures that custom payloads are passed to C*. A custom
+        query provider is used with C* so we can validate that same custom payloads are sent back
+        with the results
+
+
+        @since 2.2
```
kishkaru: This should be 2.6, since that's the upcoming driver version.

GregBestland:
Issues addressed. Commits squashed.

PR #316, opened 2015-05-19
Suggestions for PYTHON-295

stef1927:
Whilst adding small and tiny ints to cqlsh I noticed that the DESCRIBE command was wrong, the column name was 'org.apache.cassandra.db.marshal.ByteType' instead of 'tinyint' and  'org.apache.cassandra.db.marshal.ShortType' instead of 'smallint'. I think the names of the types in cqltypes.py must match the names in org.apache.cassandra.db.marshal, see lookup_casstype_simple(). Therefore I changed them accordingly.

I also added the columns, but I did not test them.

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -377,9 +377,23 @@ class Integer(Column):
         return self.validate(value)
 
 
+class TinyInt(Integer):
+    """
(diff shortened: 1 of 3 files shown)


aholmberg:
Thanks for the correction.

stef1927:
Welcome, thanks for merging so quickly.

PR #319, opened 2015-05-20
Suggestions for PYTHON-315

stef1927:
We don't have access to the future in cqlsh because we call the synchronous Session.execute() method.

I've added the warnings to the statement to keep it in line with what's already done for tracing but feel free to propose an alternative, i.e. returning the warning directly from execute() or other.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1482,7 +1482,7 @@ class Session(object):
         for future in futures:
             future.result()
 
-    def execute(self, query, parameters=None, timeout=_NOT_SET, trace=False, custom_payload=None):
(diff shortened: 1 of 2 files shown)


thobbs:
@stef1927 in cqlsh, we could always just do:

~~~python
future = session.execute_async()
results = future.result()
warnings = future.warnings
~~~

So I don't think we need changes to the driver to support this.

aholmberg:
+1 Tyler that is the intended usage. I don't want to attach request-specific data to things like statements, because they could be used concurrently for different requests (of course not in cqlsh, but in general). 

Trace data (on which I assume this is suggestion is based) is a bit of an aberration at the moment -- something I would like to change in the future.

stef1927:
OK - I will change cqlsh to use async_execute() so that we have access to both warnings and trace data.

PR #321, opened 2015-05-20
PYTHON-317 - Distinguish NULL and UNSET in bound parameters, protocol v4+

aholmberg:


tests/integration/standard/test_prepared_statements.py
```diff
@@ -213,6 +227,54 @@ def test_none_values(self):
...
+            ({'k': 0, 'v0': UNSET_VALUE, 'v1': 3}, (0, 4, 3)),
+            # nulls still work
+            ((0, None, None),                      (0, None, None)),
+        ]
+
+        for params, expected in bind_expected:
+            session.execute(insert, params)
+            results = session.execute(select, (0,))
+            self.assertEqual(results[0], expected)
+
```
GregBestland: Maybe you do so in the unit tests, but it might be good to validate that trying to use an unset value  for the primary key, throws a ValueError.

tests/integration/standard/test_prepared_statements.py
```diff
@@ -213,6 +227,54 @@ def test_none_values(self):
...
+        """
+        if PROTOCOL_VERSION < 4:
+            raise unittest.SkipTest("Binding UNSET values is not supported in protocol version < 4")
+
+        cluster = Cluster(protocol_version=PROTOCOL_VERSION)
+        session = cluster.connect()
+
+        # table with at least two values so one can be used as a marker
+        session.execute("CREATE TABLE IF NOT EXISTS test1rf.test_unset_values (k int PRIMARY KEY, v0 int, v1 int)")
+        insert = session.prepare( "INSERT INTO test1rf.test_unset_values (k, v0, v1) VALUES  (?, ?, ?)")
```
GregBestland: nit: pep8 complains about whitespace between ( and "Insert INTO test1rf.....

tests/integration/standard/test_prepared_statements.py
```diff
@@ -213,6 +227,54 @@ def test_none_values(self):
...
+        select = session.prepare( "SELECT * FROM test1rf.test_unset_values WHERE k=?")
```
GregBestland: See above

tests/integration/standard/test_prepared_statements.py
aholmberg: It is covered in the unit tests [here](https://github.com/datastax/python-driver/blob/317/tests/unit/test_parameter_binding.py#L137-L141)

tests/unit/test_parameter_binding.py
```diff
@@ -128,28 +129,89 @@ def test_inherit_fetch_size(self):
...
 
-        prepared_statement = PreparedStatement(column_metadata=column_metadata,
+    def test_dict_missing_routing_key(self):
+        self.assertRaises(KeyError, self.bound.bind, {'rk0': 0, 'ck0': 0, 'v0': 0})
+        self.assertRaises(KeyError, self.bound.bind, {'rk1': 0, 'ck0': 0, 'v0': 0})
+
+    def test_missing_value(self):
+        self.assertRaises(KeyError, self.bound.bind, {'rk0': 0, 'rk1': 0, 'ck0': 0})
+        # legacy behavior: short parameters are allowed through, errored by server
+        #self.assertRaises(ValueError, self.bound.bind, (0, 0, 0))
```
GregBestland: nit: does this need to be here?

PR #324, opened 2015-05-21
[PYTHON-238] Added tests for protocol v4 Exceptions, and MISC test cl…

GregBestland:
…eanup

### DIFF
--- a/tests/integration/long/test_failure_types.py
+++ b/tests/integration/long/test_failure_types.py
@@ -1,0 +1,279 @@
+# Copyright 2013-2015 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+
+
+from cassandra.cluster import Cluster
+from cassandra import ConsistencyLevel
+from cassandra import WriteFailure, ReadFailure, FunctionFailure
+from cassandra.concurrent import execute_concurrent_with_args
+from cassandra.query import SimpleStatement
+from tests.integration import use_singledc, PROTOCOL_VERSION, get_cluster, setup_keyspace
+
+
+def setup_module():
+    """
+    We need some custom setup for this module. All unit tests in this module
+    require protocol >=4. We won't bother going through the setup required unless that is the
+    protocol version we are using.
+    """
+
+    # If we aren't at protocol v 4 or greater don't waste time setting anything up, all tests will be skipped
+    if PROTOCOL_VERSION >= 4:
+        use_singledc(start=False)
+        ccm_cluster = get_cluster()
+        ccm_cluster.stop()
+        config_options = {'tombstone_failure_threshold': 2000, 'tombstone_warn_threshold': 1000}
+        ccm_cluster.set_configuration_options(config_options)
+        ccm_cluster.start(wait_for_binary_proto=True, wait_other_notice=True)
+        setup_keyspace()
+
+
+def teardown_module():
+    """
+    The rest of the tests don't need custom tombstones
+    reset the config options so as to not mess with other tests.
+    """
+    if PROTOCOL_VERSION >= 4:
+        ccm_cluster = get_cluster()
+        config_options = {}
+        ccm_cluster.set_configuration_options(config_options)
+        if ccm_cluster is not None:
+            ccm_cluster.stop()
+
+
+class ClientExceptionTests(unittest.TestCase):
+
+    def setUp(self):
+        """
+        Test is skipped if run with native protocol version <4
+        """
+
+        if PROTOCOL_VERSION < 4:
+            raise unittest.SkipTest(
+                "Native protocol 4,0+ is required for custom payloads, currently using %r"
+                % (PROTOCOL_VERSION,))
+
+        self.cluster = Cluster(protocol_version=PROTOCOL_VERSION)
+        self.session = self.cluster.connect()
+        self.nodes_currently_failing = []
+        self.node1, self.node2, self.node3 = get_cluster().nodes.values()
+
+    def tearDown(self):
+
+        self.cluster.shutdown()
+        failing_nodes = []
+
+        # Restart the nodes to fully functional again
+        self.setFailingNodes(failing_nodes, "testksfail")
+
+    def setFailingNodes(self, failing_nodes, keyspace):
+        """
+        This method will take in a set of failing nodes, and toggle all of the nodes in the provided list to fail
+        writes.
+        @param failing_nodes A definitive list of nodes that should fail writes
+        @param keyspace The keyspace to enable failures on
+
+        """
+
+        # Ensure all of the nodes on the list have failures enabled
+        for node in failing_nodes:
+            if node not in self.nodes_currently_failing:
+                node.stop(wait_other_notice=True, gently=False)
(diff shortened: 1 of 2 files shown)


tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+
+from cassandra.query import (PreparedStatement, BoundStatement, SimpleStatement,
```
aholmberg: Unused

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+
+    # If we aren't at protocol v 4 or greater don't waste time setting anything up, all tests will be skipped
+    if PROTOCOL_VERSION >= 4:
+        use_singledc(start=False)
+        ccm_cluster = get_cluster()
+        ccm_cluster.stop()
+        config_options = {'tombstone_failure_threshold': 2000, 'tombstone_warn_threshold': 1000}
+        ccm_cluster.set_configuration_options(config_options)
+        ccm_cluster.start(wait_for_binary_proto=True, wait_other_notice=True)
+        # there seems to be some race, with some versions of C* taking longer to
```
aholmberg: Does this auth comment apply for these tests? Maybe we can remove the comment and sleep in this case.

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+        # get the auth (and default user) setup. Sleep here to give it a chance
+        time.sleep(10)
+        setup_keyspace()
+
+
+def teardown_module():
+    """
+    The rests of the tests don't need custom tombstones
```
aholmberg: s/rests/rest/ ?

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+class ClientExceptionTests(unittest.TestCase):
+
+    def setUp(self):
+        """
+        Test is skipped if run with native protocol version <4
+        """
+
+        if PROTOCOL_VERSION < 4:
```
aholmberg: Wondering if we should just throw the skip in module setup, and avoid the additional conditionals here and in module teardown.

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+
+        self.cluster.shutdown()
+        failing_nodes = []
+
+        # Restart the nodes to fully functional again
+        self.setFailingNodes(failing_nodes, "testksfail")
+
+    def setFailingNodes(self, failing_nodes, keyspace):
+        """
+        This method will take in a set of failing nodes, and toggle all of the nodes provided to fail writes. It will
```
aholmberg: That last sentence trails off...

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+        @param failing_nodes A definitive list of nodes that should fail writes
+        @param keyspace The keyspace to enable failures on
+
+        """
+
+        # Ensure all of the nodes on the list have failures enabled
+        for node in failing_nodes:
+            if node not in self.nodes_currently_failing:
+                print "Stopping node "+node.name
```
aholmberg: Maybe use logging instead of print.

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+                node.stop(wait_other_notice=True, gently=False)
+                print "starting crippled node "+node.name
+                node.start(jvm_args=[" -Dcassandra.test.fail_writes_ks=" + keyspace], wait_for_binary_proto=True,
+                           wait_other_notice=True)
+                self.nodes_currently_failing.append(node)
+
+        # Ensure all nodes not on the list, but that are currently set to failing are enabled
+        for node in self.nodes_currently_failing:
+            if node not in failing_nodes:
+                print "stopping node "+node.name
```
aholmberg: logging?

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+                self.nodes_currently_failing.remove(node)
+
+    def _perform_cql_statement(self, text, consistency_level, expected_exception):
+        """
+        Simple helper method to preform cql statements and check for expected exception
+        @param text CQl statement to execute
+        @param consistency_level Consistency level at which it is to be executed
+        @param expected_exception Exception expected to be throw or none
+        """
+        statement = self.session.prepare(text)
```
aholmberg: Should we just use SimpleStatement here and avoid the prepare phase every time?

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # create table
+        self._perform_cql_statement(
+            """
+            CREATE TABLE testksfail.test (
+                k int PRIMARY KEY,
+                v int )
+            """, consistency_level=ConsistencyLevel.ALL, expected_exception=None)
+
+        # Simple insert so we have something to query
```
aholmberg: Why do we need something to query for the write test?

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+            self.session.execute(statement)
+        else:
+            with self.assertRaises(expected_exception) as cm:
+                self.session.execute(statement)
+
+    def test_write_failures_from_coordinator(self):
+        """
+        Test to validate that write failures from the coordinator are surfaced appropriately.
+
+        test_write_failures_from_coordinator Enable write failures on the various nodes using a custom jvm flag,
```
aholmberg: I don't see anything guaranteeing that the write failure is generated on the coordinator. I do see that it's generated on node1, which might be the coordinator, depending on the query plan generated.
This is fine, but just making sure the test intent matches what is happening.

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+        # Setup table for "wide row"
+        self._perform_cql_statement(
+            """
+            CREATE TABLE test3rf.test2 (
+                k int,
+                v0 int,
+                v1 int, PRIMARY KEY (k,v0))
+            """, consistency_level=ConsistencyLevel.ALL, expected_exception=None)
+
+        for i in range(3000):
```
aholmberg: If you care, this could be done quite a bit faster with concurrent.execute_concurrent_with_args

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+        @param consistency_level Consistency level at which it is to be executed
+        @param expected_exception Exception expected to be throw or none
+        """
+        statement = self.session.prepare(text)
+        statement.consistency_level = consistency_level
+
+        if expected_exception is None:
+            self.session.execute(statement)
+        else:
+            with self.assertRaises(expected_exception) as cm:
```
aholmberg: Superfluous named variable (can probably remove 'as cm').

tests/integration/long/test_failure_types.py
```diff
@@ -0,0 +1,295 @@
...
+        @param keyspace The keyspace to enable failures on
+
+        """
+
+        # Ensure all of the nodes on the list have failures enabled
+        for node in failing_nodes:
+            if node not in self.nodes_currently_failing:
+                print "Stopping node "+node.name
+                node.stop(wait_other_notice=True, gently=False)
+                print "starting crippled node "+node.name
```
aholmberg: pep8 whitespace. Can probably change to formatting if you choose to use logging for these messages.

tests/integration/long/test_failure_types.py
GregBestland: it does not, removed

tests/integration/long/test_failure_types.py
GregBestland: The only reason I don't want to do this is that raising a skiptest exception in the setup module seems to error out rather than mark tests as skipped.

tests/integration/long/test_failure_types.py
GregBestland: Both of these are left over debug statements I forgot to remove.

tests/integration/long/test_failure_types.py
GregBestland: We shouldn't. This was left over from some debugging. Good catch.

PR #328, opened 2015-05-26
PYTHON-315 - Tests for client warnings

aholmberg:


tests/integration/standard/test_custom_payload.py
```diff
@@ -21,54 +21,21 @@
...
-        """
-        Test is skipped if run with cql version <4
-        """
-
+    @classmethod
+    def setUpClass(cls):
         if PROTOCOL_VERSION < 4:
             raise unittest.SkipTest(
                 "Native protocol 4,0+ is required for custom payloads, currently using %r"
                 % (PROTOCOL_VERSION,))
```
kishkaru: Can't skip test here. The raise() will raise an error and error out the suite, instead of skipping it gracefully. Need to raise() SkipTest in either setUp() or in each individual test.

tests/integration/standard/test_client_warnings.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class ClientWarningTests(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        if PROTOCOL_VERSION < 4:
+            raise unittest.SkipTest(
+                "Native protocol 4,0+ is required for client warnings, currently using %r"
+                % (PROTOCOL_VERSION,))
```
kishkaru: Can't skip test here. The raise() will raise an error and error out the suite, instead of skipping it gracefully. Need to raise() SkipTest in either setUp() or in each individual test.

tests/integration/standard/test_client_warnings.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.cluster.shutdown()
+
+    def test_warning_basic(self):
+        """
+        Test to validate that client warnings can be surfaced
+
+        @since 2.6
```
kishkaru: This should be 2.6.0

tests/integration/standard/test_client_warnings.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.cluster.shutdown()
+
+    def test_warning_basic(self):
+        """
+        Test to validate that client warnings can be surfaced
+
```
kishkaru: There should be a longer description here that explains exactly what the test is doing.

tests/integration/standard/test_client_warnings.py
```diff
@@ -0,0 +1,124 @@
...
+        @since 2.6
+        @jira_ticket PYTHON-315
+        @expected_result valid warnings returned
+        @test_assumptions
+            - batch_size_warn_threshold_in_kb: 5
+        @test_category queries:client_warning
```
kishkaru: I'd put this under category error_codes

tests/integration/standard/test_client_warnings.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        future = self.session.execute_async(self.warn_batch)
```
kishkaru: There should be a newline after the end of the comment block, and before the beginning of code.

kishkaru:
Tests themselves look good. Double check that the tests are skipped properly in earlier protocol versions.

tests/integration/standard/test_client_warnings.py
aholmberg: There are no error codes involved with client warnings. (?)

tests/integration/standard/test_client_warnings.py
aholmberg: Where are you getting these rules? I think it's mostly inconsequential and not worth a comment but, for example, [this convention](https://www.python.org/dev/peps/pep-0257/) specifically states "There's no blank line either before or after the docstring.".

tests/integration/standard/test_client_warnings.py
aholmberg: I'm not sure what you mean. In my experimentation this has the desired effect of skipping the entire TestCase class.

aholmberg: Following up, Kishan noticed that this errors in environments utilizing unittest2 instead of unittest. It's [reported fixed in 0.4.2](https://pypi.python.org/pypi/unittest2#id9), but still erroring in 1.0.1

kishkaru:
LGTM.

PR #332, opened 2015-05-26
PYTHON-295 - Update type tests

aholmberg:


kishkaru:
LGTM. Thanks for cleaning up the type and tuple tests.

PR #335, opened 2015-05-27
Add support for upper case uuid

kracekumar:
Systems like iOS creates uuid in uppercase, while storing to
cassandra one has to convert to lowercase before storing in cassandra.
This is quite difficult for developer to remember, it is better
handle at driver level.

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -15,7 +15,6 @@
 from copy import deepcopy, copy
 from datetime import date, datetime
 import logging
-import re
 import six
 import warnings
 
@@ -518,8 +517,6 @@ class UUID(Column):
     """
     db_type = 'uuid'
 
-    re_uuid = re.compile(r'[0-9a-f]{8}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{4}-?[0-9a-f]{12}')
-
     def validate(self, value):
         val = super(UUID, self).validate(value)
         if val is None:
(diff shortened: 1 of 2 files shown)


aholmberg:
Thanks for the suggestion. I think it would be more efficient to make the regex match upper-case hex. If you end up making changes, I also noticed the regex is missing an end anchor.

kracekumar:
@aholmberg Thanks. Fixed the regex. Normally Python libraries validate `uuid`, using `uuid.UUID` [1] and catch an exception if it is an  invalid `uuid`. Any specific reason to use regex ?

Also performance difference is in `us`.

```
In [24]: %timeit re_uuid = re.compile(r'[0-9a-fA-F]{8}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{12}$');re_uuid.match("a0ddb9c1-50cb-4453-85fb-229ff66191A7")
The slowest run took 1676.05 times longer than the fastest. This could mean that an intermediate result is being cached
1000000 loops, best of 3: 1.51 µs per loop

In [25]: import uuid

In [26]: %timeit uuid.UUID("a0ddb9c1-50cb-4453-85fb-229ff66191A7")
The slowest run took 33.52 times longer than the fastest. This could mean that an intermediate result is being cached
100000 loops, best of 3: 4.3 µs per loop
```

[1]: https://github.com/schematics/schematics/blob/development/schematics/types/base.py#L281

aholmberg:
@kracekumar honestly I'm not sure why it was done this way. I think the routine you linked seems straightforward and robust. You could incorporate that, or not. Either way, would you mind squashing intermediate commits?

kracekumar:
@aholmberg Sorry, I tried my best I couldn't squash properly. I can submit new pull request is that ok ?

aholmberg:
@kracekumar you should be able to ``git reset  1ba07f1`` (reset to the parent commit), commit, then ``git push -f origin master``. If that doesn't work, another pr is fine. Thanks!

kracekumar:
@aholmberg Thanks for the help!

aholmberg:
Thanks for the contribution!

PR #336, opened 2015-05-27
[Python-211] Adding documentation for metadata tests

GregBestland:


tests/integration/standard/test_metadata.py
```diff
@@ -1357,20 +1357,31 @@ def make_function_kwargs(self, called_on_null=True):
...
 
-        # Ideally we would make a function that takes a udt type, but
-        # this presently fails because C* c059a56 requires udt to be frozen to create, but does not store meta indicating frozen
-        # https://issues.apache.org/jira/browse/CASSANDRA-9186
-        # Maybe update this after release
-        #kwargs = self.make_function_kwargs()
-        #kwargs['type_signature'][0] = "frozen<%s>" % udt_name
-
-        #expected_meta = Function(**kwargs)
-        #with self.VerifiedFunction(self, **kwargs):
```
aholmberg: Moving this commented block into the header loses its original context in the code. Are we trying to have no inline comments?

PR #337, opened 2015-05-27
PYTHON-309, PYTHON-20 - Add windows inet_pton for python versions that do not have the socket…

aholmberg:
… routine

Fixes an issue with inet addresses in the protocol (PYTHON-309)
Adds IPv6 support for Windows (PYTHON-20)

### DIFF
--- a/cassandra/util.py
+++ b/cassandra/util.py
@@ -1011,3 +1011,86 @@ class Date(object):
         except:
             # If we overflow datetime.[MIN|MAX]
             return str(self.days_from_epoch)
+
+import socket
+if hasattr(socket, 'inet_pton'):
+    inet_pton = socket.inet_pton
+    inet_ntop = socket.inet_ntop
+else:
+    """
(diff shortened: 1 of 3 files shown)


cassandra/util.py
```diff
@@ -1011,3 +1011,86 @@ def __str__(self):
...
+    Windows doesn't have socket.inet_pton and socket.inet_ntop until Python 3.4
+    This is an alternative impl using ctypes, based on this win_inet_pton project:
+    https://github.com/hickeroar/win_inet_pton
+    """
+    import ctypes
+
+
+    class sockaddr(ctypes.Structure):
+        _fields_ = [("sa_family", ctypes.c_short),
+                    ("__pad1", ctypes.c_ushort),
```
mpenick: Might want to comment here that "__pad1" skips over the "u_short sin_port" for v4 and "ipv4_addr" skips over the "u_long sin6_flowinfo" for v6 (from https://msdn.microsoft.com/en-us/library/windows/desktop/ms740496(v=vs.85).aspx). This is a bit clever. :)

PR #340, opened 2015-05-28
Update CQL keywords, expose in API docs.

aholmberg:
PYTHON-319
PYTHON-324

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -37,19 +37,44 @@ from cassandra.util import OrderedDict
 
 log = logging.getLogger(__name__)
 
-_keywords = set((
-    'select', 'from', 'where', 'and', 'key', 'insert', 'update', 'with',
-    'limit', 'using', 'use', 'count', 'set',
-    'begin', 'apply', 'batch', 'truncate', 'delete', 'in', 'create',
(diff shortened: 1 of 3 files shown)


cassandra/metadata.py
```diff
@@ -37,19 +37,44 @@
...
+    'tinyint', 'to', 'token', 'trigger', 'truncate', 'ttl', 'tuple', 'type', 'unlogged', 'update', 'use', 'user',
+    'users', 'using', 'uuid', 'values', 'varchar', 'varint', 'where', 'with', 'writetime'
 ))
-
-_unreserved_keywords = set((
-    'key', 'clustering', 'ttl', 'compact', 'storage', 'type', 'values'
+"""
+Set of keywords in CQL.
+
+Derived from .../cassandra/src/java/org/apache/cassandra/cql3/Cql.g
```
mpenick: Missing? consistency, level, non

aholmberg: I'm not sure. I was basing these on the lists (here)[https://github.com/apache/cassandra/blob/6eea3ea395957d1e799952de3423cc29fd469e3e/src/java/org/apache/cassandra/cql3/Cql.g#L1503-L1559]. Where do these three come from?

PR #344, opened 2015-06-04
fix unicode encoding issue for query.bind_params

ashwinrajeev:
`bind_params` raises `UnicodeDecodeError` if query is unicode. Since `bind_params` is used by `Cluster.execute` and `BatchStatement.add` it is better to move type checking of query inside `bind_params`.


aholmberg:
Thanks for the contribution! I created a ticket to track and test: https://datastax-oss.atlassian.net/browse/PYTHON-334

PR #345, opened 2015-06-06
Write timeouts being ignored with 0 responses

sanchaz:
DowngradingConsistencyRetryPolicy was ignoring write timeouts for
WriteTypes SIMPLE, BATCH and COUNTER even if responses received
was equal to 0.
It is valid to ignore the failure only if at least one response was
received.

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -816,7 +816,7 @@ class DowngradingConsistencyRetryPolicy(RetryPolicy):
                          required_responses, received_responses, retry_num):
         if retry_num != 0:
             return (self.RETHROW, None)
-        elif write_type in (WriteType.SIMPLE, WriteType.BATCH, WriteType.COUNTER):
+        elif write_type in (WriteType.SIMPLE, WriteType.BATCH, WriteType.COUNTER) and received_responses > 0:
             return (self.IGNORE, None)
         elif write_type == WriteType.UNLOGGED_BATCH:
             return self._pick_consistency(received_responses)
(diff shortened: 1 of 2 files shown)


avalanche123:
There is no gurantee that the write hasn't completed on some nodes. 0 responses means no nodes could deliver an ack, but the coordinator still forwarded the write to those. So they either haven't received the write or could deliver an ack, either way it is not safe to retry automatically. The application can retry based on whether the request was idempotent or not, but not the driver.

-1

sanchaz:
so in that case shouldn't it at the very least rethrow?

avalanche123:
Yes, certainly!

sanchaz:
The doc states
    * For writes, ignore the exception (thus silently failing the
      consistency requirement) if we know the write has been persisted on at
      least one replica.
If we get 0 responses we have no guarantee of this.
As such if we get 0 responses we rethrow this.
Otherwise the current behaviour of ignoring the failure is used.

aholmberg:
I'm not sure I agree that we should rethrow. If we get this response, the write either made it, or is at least persisted as a hint. None of the other drivers consider received responses for write timeouts on these types of requests.

aholmberg:
@sanchaz I think I'm onboard with this understanding: even if a hint is persisted at the time of this request, we can't guarantee it is delivered inside the hints window. Thus, no guaranteed persistence.
Sorry for the misunderstanding.
Presently tracking in PYTHON-338

PR #349, opened 2015-06-10
PYTHON-303 - Don't wait for schema agreement during startup.

aholmberg:
Fixes an issue where connect will stall when connecting to mixed version
clusters.
Should not affect model consistency since we are already registered for
schema_change events.


GregBestland:
Adam +1 LGTM.

PR #352, opened 2015-06-12
Set models.DEFAULT_KEYSPACE when calling set_session()

mikeokner:
Sets the default keyspace for cqlengine models when passing in an existing session to `cqlengine.connection.set_session()`. Without this, the user will see an error when trying to perform Cassandra interactions with a Model instance like "Keyspace none does not exist" unless they explicitly set `models.DEFAULT_KEYSPACE` themselves ahead of time.

### DIFF
--- a/cassandra/cqlengine/connection.py
+++ b/cassandra/cqlengine/connection.py
@@ -82,6 +82,11 @@ def set_session(s):
     session = s
     cluster = s.cluster
 
+    # Set default keyspace from given session's keyspace if not already set
+    from cassandra.cqlengine import models
+    if not models.DEFAULT_KEYSPACE and session.keyspace:
+        models.DEFAULT_KEYSPACE = session.keyspace
+
     _register_known_types(cluster)
 
     log.debug("cqlengine connection initialized with %s", s)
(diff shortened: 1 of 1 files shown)


mikeokner:
Looks like all but one permutation passed, and the one that failed was due to an interrupted connection:
```ConnectionShutdown: Connection to localhost was closed```

File: cassandra/cqlengine/connection.py [hunk 1]
```diff
@@ -82,6 +82,13 @@ def set_session(s):
     session = s
     cluster = s.cluster
 
+    # Set default keyspace from given session's keyspace
+    if not session.keyspace:
+        raise UndefinedKeyspaceException()
```
aholmberg: I'm not sure we should blow this up. It's possible to use models without a session keyspace or default with {{__keyspace__}} set on them.

I could see setting models.DEFAULT_KEYSPACE to session.keyspace if default is not set and session is.

mikeokner: Ah ok, that makes sense. I'll push up another commit with those changes.

mikeokner:
Update pushed. Only setting `models.DEFAULT_KEYSPACE = session.keyspace` if the default isn't set and session.keyspace is. No longer raising the exception.

I should follow up with another question for you: should [this exception](https://github.com/mikeokner/python-driver/blob/master/cassandra/cqlengine/connection.py#L117) be raised? I based my original changes on that section of the `setup()` method, but it would seem the same argument you made could be applied there as well.

aholmberg:
> should this exception be raised?

Good question. I think not. 

If you end up removing that, would you mind squashing these commits before I merge?

mikeokner:
How's that?

aholmberg:
Thanks for contributing!

PR #354, opened 2015-06-12
Windows tests suppport and stabilization

kishkaru:
Windows tests suppport and stabilization

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -12,25 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import time
-import traceback
-
 try:
     import unittest2 as unittest
 except ImportError:
     import unittest  # noqa
 
-import logging
-log = logging.getLogger(__name__)
-
-import os
+import os, six, time, sys, logging, traceback
 from threading import Event
-import six
 from subprocess import call
-
 from itertools import groupby
 
+from cassandra import OperationTimedOut, ReadTimeout, ReadFailure, WriteTimeout, WriteFailure
 from cassandra.cluster import Cluster
+from cassandra.protocol import ConfigurationException
 
 try:
     from ccmlib.cluster import Cluster as CCMCluster
@@ -38,6 +32,8 @@ try:
     from ccmlib import common
 except ImportError as e:
     CCMClusterFactory = None
+
+log = logging.getLogger(__name__)
 
 CLUSTER_NAME = 'test_cluster'
 SINGLE_NODE_CLUSTER_NAME = 'single_node'
@@ -84,7 +80,7 @@ def _tuple_version(version_string):
 
 USE_CASS_EXTERNAL = bool(os.getenv('USE_CASS_EXTERNAL', False))
 
-default_cassandra_version = '2.1.3'
+default_cassandra_version = '2.1.5'
 
 if USE_CASS_EXTERNAL:
     if CCMClusterFactory:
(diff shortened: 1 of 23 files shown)


tests/integration/__init__.py
```diff
@@ -158,8 +154,16 @@ def remove_cluster():
...
+        while True:
+            try:
+                CCM_CLUSTER.remove()
+                CCM_CLUSTER = None
+                break
+            except WindowsError:
+                ex_type, ex, tb = sys.exc_info()
+                log.warn("{0}: {1} Backtrace: {2}".format(ex_type.__name__, ex, traceback.extract_tb(tb)))
+                del tb
+                time.sleep(1)
```
GregBestland: It may not be an issue here, just due to the nature of ccm in windows, but we may want to limit the retry attempts so it doesn't run indefinitely.

aholmberg: Curious what ``WindowsError``s show up here?

kishkaru: Specifically, `WindowsError: [Error 5] Access is denied`. This is caused due to the Windows filesystem holding on to files, even though we've released them. Typically the resources are released after a few seconds.

tests/integration/__init__.py
```diff
@@ -240,6 +250,19 @@ def teardown_package():
...
+            return session.execute(query)
+        except ConfigurationException:
+            # keyspace/table was already created/dropped
+            return
+        except (OperationTimedOut, ReadTimeout, ReadFailure, WriteTimeout, WriteFailure):
+            ex_type, ex, tb = sys.exc_info()
+            log.warn("{0}: {1} Backtrace: {2}".format(ex_type.__name__, ex, traceback.extract_tb(tb)))
+            del tb
+
+
```
GregBestland: We might want a sanity check here, for instance stop retrying after a specific number of iterations that we should never hit.

kishkaru: In this case, we do want to try indefinitely until the query passes. As we handle the expected exceptions, anything else will be raised to the caller. We could have a max_tries = 100 for example, but we need to do some testing to make sure it can handle the worst case scenario in the current test suite.

tests/integration/long/test_schema.py
```diff
@@ -86,28 +86,13 @@ def test_for_schema_disagreements_different_keyspaces(self):
...
-                        session.execute("DROP KEYSPACE test_{0}".format(i))
-                        break
-                    except OperationTimedOut:
-                        ex_type, ex, tb = sys.exc_info()
-                        log.warn("{0}: {1} Backtrace: {2}".format(ex_type.__name__, ex, traceback.extract_tb(tb)))
-                        del tb
-                    except ConfigurationException:
-                        # We're good, the keyspace was never created due to OperationTimedOut
-                        break
+            execute_until_pass(session, "CREATE KEYSPACE test_{0} WITH replication = {{'class': 'SimpleStrategy', 'replication_factor': 1}}".format(i))
```
GregBestland: So much cleaner now!

tests/integration/standard/test_concurrent.py
```diff
@@ -47,6 +50,24 @@ def setUpClass(cls):
...
+
+    def execute_concurrent_args_helper(self, session, query, params):
+        while True:
+            try:
+                return execute_concurrent_with_args(session, query, params)
+            except (ReadTimeout, WriteTimeout, OperationTimedOut, ReadFailure, WriteFailure):
+                ex_type, ex, tb = sys.exc_info()
+                log.warn("{0}: {1} Backtrace: {2}".format(ex_type.__name__, ex, traceback.extract_tb(tb)))
+                del tb
+
```
GregBestland: Sanity check for these timeouts maybe?

tests/unit/test_time_util.py
```diff
@@ -42,11 +42,11 @@ def test_times_from_uuid1(self):
...
         t = util.unix_time_from_uuid1(u)
-        self.assertAlmostEqual(now, t, 3)
+        self.assertAlmostEqual(now, t, 2)
 
         dt = util.datetime_from_uuid1(u)
         t = calendar.timegm(dt.timetuple()) + dt.microsecond / 1e6
-        self.assertAlmostEqual(now, t, 3)
+        self.assertAlmostEqual(now, t, 2)
 
     def test_uuid_from_time(self):
```
GregBestland: Nice Job +1 really cleaned a ton up here.

tests/integration/standard/test_concurrent.py
kishkaru: The title of the method is `execute_concurrent_args_helper`. Not sure how else I can specify that it's a helper.

GregBestland: I meant indicate that it retries in some way,

kishkaru:
Addressed PR comments and squashed commits.

PR #359, opened 2015-06-22
PYTHON-352 - For meta export, don't copy in default min/max compaction thresholds

aholmberg:
Fixes an issue where table meta export would contain min/max_threshold
settings, even for compaction strategies that do not accept those.
Resulting CQL would cause "ConfigurationException ... code=2300 ...
Properties specified [min_threshold, max_threshold] are not
understood by [some compaction strategy]"


GregBestland:
+1 LGTOM

PR #361, opened 2015-06-25
Enable C extensions on pypy and add pypy3 support

sontek:
I'm not sure the history behind disabling these on pypy but they work just fine.  I also added a quick fix in the murmur3.c code to support pypy3

### DIFF
--- a/cassandra/murmur3.c
+++ b/cassandra/murmur3.c
@@ -20,6 +20,13 @@ typedef int Py_ssize_t;
 #define PY_SSIZE_T_MIN INT_MIN
 #endif
 
+#ifdef PYPY_VERSION
+#define COMPILING_IN_PYPY 1
+#define COMPILING_IN_CPYTHON 0
+#else
+#define COMPILING_IN_PYPY 0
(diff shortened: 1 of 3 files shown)


aholmberg:
Thanks for the PR. I actually don't know the history on disabling. If I had to guess, it might be that there was not enough engineering bandwidth to go there at the time, or possibly that PyPy was at a different level of maturity. Either way, I'll look into it.

Created to track and test: https://datastax-oss.atlassian.net/browse/PYTHON-357

aholmberg:
I'm not finding the libevwrapper extension reliable in pypy. Seems to be failing an assertion in the RPython ref counting emulation. I don't think this is something worth digging into at this point. I'm thinking of changing to explicitly whtie-list only the murmur3 extension, which seems to work fine across the board.
Does this suit your needs?

aholmberg:
Tweaked and merged from alternate, rebased branch: https://github.com/datastax/python-driver/commit/f7c0bd312df6ac105173969c8b57ee4b8bd9c664

PR #362, opened 2015-06-29
SSL Testing

GregBestland:
Added a few test cases for SSL


tests/integration/long/test_ssl.py
```diff
@@ -56,11 +73,18 @@ def teardown_module():
...
+    ccm_cluster = get_cluster()
+    ccm_cluster.stop()
     remove_cluster()
 
 
 class SSLConnectionTests(unittest.TestCase):
 
+    @classmethod
+    def setUpClass(cls):
+        print("initializing cluster")
```
aholmberg: we usually don't print directly

tests/integration/long/test_ssl.py
```diff
@@ -41,12 +49,21 @@ def setup_module():
...
+
+    if(client_auth):
+        abs_path_server_truststore_path = os.path.abspath(SERVER_TRUSTSTORE_PATH)
+        client_encyrption_options = config_options['client_encryption_options']
+        client_encyrption_options['require_client_auth'] = True
+        client_encyrption_options['truststore'] = abs_path_server_truststore_path
+        client_encyrption_options['truststore_password'] = DEFAULT_PASSWORD
+
     ccm_cluster.set_configuration_options(config_options)
     ccm_cluster.start(wait_for_binary_proto=True, wait_other_notice=True)
```
aholmberg: Something is causing this to fail when run as follows:

    nosetests -s -v tests/integration/long/test_ssl.py

ccm (2.0.3.1), java version "1.8.0_45"

I haven't quite figured this out.

tests/integration/long/test_ssl.py
```diff
@@ -108,4 +132,109 @@ def test_can_connect_with_ssl_ca(self):
...
+        @expected_result The client will throw an exception on connect
+
+        @test_category connection:ssl
+        """
+
+        abs_path_ca_cert_path = os.path.abspath(CLIENT_CA_CERTS)
+        cluster = Cluster(protocol_version=PROTOCOL_VERSION, ssl_options={'ca_certs': abs_path_ca_cert_path,
+                                                                          'ssl_version': ssl.PROTOCOL_TLSv1})
+        # attempt to connect and expect an exception
+        with self.assertRaises(Exception) as context:
```
aholmberg: Is there a more specific exception we could verify?

tests/integration/long/test_ssl.py
```diff
@@ -108,4 +132,109 @@ def test_can_connect_with_ssl_ca(self):
...
+        # Setup absolute paths to key/cert files
+        abs_path_ca_cert_path = os.path.abspath(CLIENT_CA_CERTS)
+        abs_driver_keyfile = os.path.abspath(DRIVER_KEYFILE)
+        abs_driver_certfile = os.path.abspath(DRIVER_CERTFILE_BAD)
+
+        cluster = Cluster(protocol_version=PROTOCOL_VERSION, ssl_options={'ca_certs': abs_path_ca_cert_path,
+                                                                          'ssl_version': ssl.PROTOCOL_TLSv1,
+                                                                          'keyfile': abs_driver_keyfile,
+                                                                          'certfile': abs_driver_certfile})
+        with self.assertRaises(Exception) as context:
```
aholmberg: Specific exception?

tests/integration/long/test_ssl.py
```diff
@@ -108,4 +132,109 @@ def test_can_connect_with_ssl_ca(self):
...
+        @test_category connection:ssl
+        """
+
+        # Need to get absolute paths for certs/key
+        abs_path_ca_cert_path = os.path.abspath(CLIENT_CA_CERTS)
+        abs_driver_keyfile = os.path.abspath(DRIVER_KEYFILE)
+        abs_driver_certfile = os.path.abspath(DRIVER_CERTFILE)
+
+        tries = 0
+        while True:
```
aholmberg: Curious why we need timeouts here. Could we improve by making sure the server is up, and expecting this to work first time?

tests/integration/long/test_ssl.py
```diff
@@ -108,4 +132,109 @@ def test_can_connect_with_ssl_ca(self):
...
+
+        # attempt a few simple commands.
+        insert_keyspace = """CREATE KEYSPACE ssltest
+            WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '3'}
+            """
+        statement = SimpleStatement(insert_keyspace)
+        statement.consistency_level = 3
+        session.execute(statement)
+
+        drop_keyspace = "DROP KEYSPACE ssltest"
```
aholmberg: any reason we don't just:

    session.execute("<cql>")
?

tests/integration/long/test_ssl.py
```diff
@@ -108,4 +132,109 @@ def test_can_connect_with_ssl_ca(self):
...
+
+                session = cluster.connect()
+                break
+            except Exception:
+                ex_type, ex, tb = sys.exc_info()
+                log.warn("{0}: {1} Backtrace: {2}".format(ex_type.__name__, ex, traceback.extract_tb(tb)))
+                del tb
+                tries += 1
+
+        # attempt a few simple commands.
```
aholmberg: would it be sufficient to just insert into one of the common test tables created in init?
(avoids more schema changes while verifying request execution)

tests/integration/long/test_ssl.py
GregBestland: I've ben running in this manner. "CASSANDRA_VERSION=2.2.0-rc2 nosetests -s -v tests/integration/long/test_ssl.py". I've also tried with C* version 2.1.8, with no issues. I'm not sure this test can be run against a standing C* cluster as it needs to restart the CCM cluster with custom yaml configs for the various certs/keys.

aholmberg: JCE strikes again. Local machine upgraded to jdk8, Unlimited JCE did not come along for the ride.

tests/integration/long/test_ssl.py
GregBestland: I suppose in this case this case what I'm doing is more than sufficient.

tests/integration/long/test_ssl.py
GregBestland: We should come up with a standard policy for these types of tests. I was attempting to setup the ssl tests in such a way where they will have no affect on any other tests. we have standard keyspaces for various replication factors test1rf, test2rf, test3rf.. etc. but I don't see any standard tables, perhaps we should add some for simple validations such as the above.

GregBestland:
Made some changes based on feedback, everything is now squashed.

PR #366, opened 2015-07-10
Python 108 Test

GregBestland:
Unit/Integration tests for python 108.
Added improvements infrastructure changes for various reactor tests going forward. 
Cleaned up tox dependencies.

### DIFF
--- a/tests/integration/long/test_failure_types.py
+++ b/tests/integration/long/test_failure_types.py
@@ -12,14 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys, logging, traceback
+import sys,logging, traceback, time
 
 from cassandra import ConsistencyLevel, OperationTimedOut, ReadTimeout, WriteTimeout, ReadFailure, WriteFailure,\
     FunctionFailure
 from cassandra.cluster import Cluster
 from cassandra.concurrent import execute_concurrent_with_args
 from cassandra.query import SimpleStatement
-from tests.integration import use_singledc, PROTOCOL_VERSION, get_cluster, setup_keyspace, remove_cluster
+from tests.integration import use_singledc, PROTOCOL_VERSION, get_cluster, setup_keyspace, remove_cluster, get_node
+from mock import Mock
 
 try:
     import unittest2 as unittest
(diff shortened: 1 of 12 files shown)


tests/integration/long/test_failure_types.py
```diff
@@ -301,3 +302,75 @@ def test_user_function_failure(self):
...
+
+
+class TimeoutTimerTest(unittest.TestCase):
+    def setUp(self):
+        """
+        Setup sessions and pause node1
+        """
+        self.cluster = Cluster(protocol_version=PROTOCOL_VERSION)
+        self.session = self.cluster.connect()
+        self.node1, self.node2, self.node3 = get_cluster().nodes.values()
```
aholmberg: looks like ``self.node[23]`` are not used. Maybe just get node1 by name?

tests/integration/long/test_failure_types.py
```diff
@@ -301,3 +302,75 @@ def test_user_function_failure(self):
...
+        ss = SimpleStatement('SELECT * FROM test3rf.test', consistency_level=ConsistencyLevel.ALL)
+
+        # Test with default timeout (should be 10)
+        start_time = time.time()
+        future = self.session.execute_async(ss)
+        with self.assertRaises(OperationTimedOut):
+            future.result()
+        end_time = time.time()
+        total_time = end_time-start_time
+        expected_time = 10
```
aholmberg: possibly pull this from ``session.default_timeout``?

tests/integration/long/test_failure_types.py
```diff
@@ -301,3 +302,75 @@ def test_user_function_failure(self):
...
+        # check timeout and ensure it's within a reasonable range
+        self.assertTrue(expected_time-.05 <= total_time <= expected_time+.05)
```
aholmberg: ``self.assertAlmostEqual`` is a handy way of expressing "within a tolerance".

tests/integration/long/test_failure_types.py
```diff
@@ -301,3 +302,75 @@ def test_user_function_failure(self):
             """
             DROP TABLE test3rf.d;
             """, consistency_level=ConsistencyLevel.ALL, expected_exception=None)
+
+
+class TimeoutTimerTest(unittest.TestCase):
```
aholmberg: Might be good to do some requests with callbacks and make sure the errback is called on timeout (this is new behavior with this change).

This could possibly be done as a standard integration test with a really short timeout (standard/test_cluster, maybe?) to get some coverage there.

tests/unit/io/eventlet_utils.py
```diff
@@ -0,0 +1,37 @@
...
+import socket
+import thread
+import Queue
+import threading
+import __builtin__
+import ssl
+import time
+
+
+def eventlet_un_patch_all():
```
aholmberg: Now that we have these (eventlet and gevent unpatch), would it be possible to always run those tests (vs. requiring pre-patching)? Not saying it should be part of this task, but maybe something new for the backlog.

tests/unit/io/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    while len(pending_callbacks) is not 0:
+        for callback in pending_callbacks:
+            if callback.was_invoked():
+                pending_callbacks.remove(callback)
+                completed_callbacks.append(callback)
+        time.sleep(.1)
+
+    # ensure they are all called back in a timely fashion
+    for callback in completed_callbacks:
+        unit_test.assertTrue(callback.expected_wait-.1 <= callback.get_wait_time() <= callback.expected_wait+.1)
```
aholmberg: assertAlmostEqual?

tests/unit/io/test_asyncorereactor.py
```diff
@@ -300,3 +300,37 @@ def test_partial_message_read(self, *args):
...
+        """
+        Verify that timer timeouts are honored appropriately
+        """
+        c = self.make_connection()
+        # Tests timers submitted in order at various timeouts
+        submit_and_wait_for_completion(self, AsyncoreConnection, 0, 100, 1, 100)
+        # Tests timers submitted in reverse order at various timeouts
+        submit_and_wait_for_completion(self, AsyncoreConnection, 100, 0, -1, 100)
+        # Tests timers submitted in varying order at various timeouts
+        submit_and_wait_for_completion(self, AsyncoreConnection, 0, 100, 1, 100, True)
```
aholmberg: Seems like several of these tests are the same. Would there be any benefit to writing a common base and specializing setup/teardown?
Maybe not worth it?

tests/unit/io/test_eventletreactor.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class EventletTimerTest(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        if EventletConnection is None:
+            raise unittest.SkipTest("Eventlet libraries not available")
+        if not is_eventlet_monkey_patched():
+            raise unittest.SkipTest("Can't test eventlet without monkey patching")
```
aholmberg: Raising skip out of a class or module setup causes an error when using unittest2 in any runtime other than python 2.6. This is mostly solved by excluding unittest2 in other tox testenvs. Just thought I'd mention in case there was an easy way to do this by method.

tests/unit/io/eventlet_utils.py
aholmberg: This doesn't appear to be used. Did you find that eventlet unpatching didn't work as well as gevent?

If not, do you think we should add a ``python setup.py eventlet_nosetests -w tests/unit`` to tox commands?

tests/unit/io/test_eventletreactor.py
```diff
@@ -0,0 +1,73 @@
...
+from tests import is_eventlet_monkey_patched
+import time
+
+try:
+    from cassandra.io.eventletreactor import EventletConnection
+except ImportError:
+    EventletConnection = None  # noqa
+
+
+class EventletTimerTest(unittest.TestCase):
```
aholmberg: How do you run these? I'm getting a hang when running via setup.py.

aholmberg: I think it's because LibEv setup only tests gevent patching, and does not skip on eventlet patching:
https://github.com/datastax/python-driver/pull/366/files#diff-46f276e998c7473a33539b5025b17de5R54

tests/unit/io/eventlet_utils.py
GregBestland: So I was not able to get the unpatching for eventlet to work in all cases. I will look at adding this to our test runs in tox.

tests/unit/io/test_asyncorereactor.py
GregBestland: I've already broken out all the logic into common utility methods. You would save like 3 lines of code probably not worth it in this case.

tests/unit/io/test_eventletreactor.py
GregBestland: Looks like this is a legacy issue. I'll change it over to use the same monkey patch code I use in the libev timer tests. That should check for both kinds of monkey patching.

tests/unit/io/test_eventletreactor.py
GregBestland: Didn't know this. Thanks!

tests/integration/long/test_failure_types.py
GregBestland: My only concern with adding that sort of coverage without a node pausing, is that it might not be deterministic. Especially with ccm we often see very quick response times. I'll see what I can maybe I can make something work

GregBestland:
Everything in the code review should be addressed, other than the extra integration test for the standard runtime, which I will add to my todo. Thanks Adam.

PR #367, opened 2015-07-14
PYTHON-123 - Generator inputs/outputs for execute_concurrent

aholmberg:
Avoids materializing entire sequence before/after execution.

### DIFF
--- a/cassandra/concurrent.py
+++ b/cassandra/concurrent.py
@@ -24,7 +24,6 @@ from cassandra.cluster import PagedResult
 
 log = logging.getLogger(__name__)
 
-
 def execute_concurrent(session, statements_and_parameters, concurrency=100, raise_on_first_error=True):
     """
     Executes a sequence of (statement, parameters) tuples concurrently.  Each
@@ -74,11 +73,6 @@ def execute_concurrent(session, statements_and_parameters, concurrency=100, rais
     if not statements_and_parameters:
         return []
 
-    # TODO handle iterators and generators naturally without converting the
-    # whole thing to a list.  This would require not building a result
-    # list of Nones up front (we don't know how many results there will be),
-    # so a dict keyed by index should be used instead.  The tricky part is
-    # knowing when you're the final statement to finish.
     statements_and_parameters = list(statements_and_parameters)
 
     event = Event()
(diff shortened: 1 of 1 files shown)


cassandra/concurrent.py
```diff
@@ -74,32 +84,121 @@ def execute_concurrent(session, statements_and_parameters, concurrency=100, rais
...
+    def _put_result(self, result, idx, success):
+        with self._condition:
+            heappush(self._results_queue, (idx, (success, result)))
+            self._execute_next()
+            self._condition.notify()
+
+    def _results(self):
+        with self._condition:
+            while self._current < self._exec_count:
+                while not self._results_queue or self._results_queue[0][0] != self._current:
```
mpenick: I don't think the `self._results_queue[0][0] != self._current` and `self._results_queue[0][0] == self._current` checks are necessary. Because the result queue is a heap if it has items then the front will alway be current. If that's the case the code could look something like this:

```python
class ConcurrentExecutorGenResults(_ConcurrentExecutor):

    # ...

    def _results(self):
        with self._condition:
            while self._current < self._exec_count:
                if self._results_queue:
                    _, res = heappop(self._results_queue)
                    if self._fail_fast and not res[0]:
                        self._raise(res[1])
                    yield res
                    self._current += 1
                else:
                    self._condition.wait()
```

Or if you want the inner results loop to be a bit tighter you could do something like this (uglier though):

```python
class ConcurrentExecutorGenResults(_ConcurrentExecutor):

    # ...

    def _results(self):
        with self._condition:
            while True:
                while self._results_queue:
                    _, res = heappop(self._results_queue)
                    if self._fail_fast and not res[0]:
                        self._raise(res[1])
                    yield res
                    self._current += 1

                if self._current < self._exec_count:
                    self._condition.wait()
                else:
                    break
```

cassandra/concurrent.py
```diff
@@ -74,32 +84,121 @@ def execute_concurrent(session, statements_and_parameters, concurrency=100, rais
...
-        return results
+
+
+class ConcurrentExecutorGenResults(_ConcurrentExecutor):
+
+    def _put_result(self, result, idx, success):
+        with self._condition:
+            heappush(self._results_queue, (idx, (success, result)))
+            self._execute_next()
+            self._condition.notify()
```
mpenick: Does this need to notify every time or could it notify in batches (or when an error happens with fast fail)? Signaling a condition variable is not cheap because it results in multiple `futex` calls (on Linux) and results in several context switches. Because of that expense makes it's advantageous to amortize that cost over multiple request results.  Still thinking of a clean way to implement this...

cassandra/concurrent.py
aholmberg: I'm doing that in case I get later requests in first. I want to make sure to yield them in order.

cassandra/concurrent.py
aholmberg: Batching seems like a good idea. I did some experimentation with batching notifications and didn't come up with anything that performed better. If you think of something I can give it a try.

cassandra/concurrent.py
mpenick: Makes sense.

PR #368, opened 2015-07-15
PYTHON-357 - enable murmur3 for PyPy runtimes

aholmberg:
superset of #361
made against 2.7 instead of master
Based on #364 -- review/merge 364 first for simplified diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ tests/integration/ccm
 setuptools*.tar.gz
 setuptools*.egg
 
+cassandra/*.c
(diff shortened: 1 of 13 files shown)


iksaif:
There is still a 'if platform.python_implementation() == 'CPython':' in cluster.py. Is that expected ?

aboudreault:
@iksaif Good catch. This shouldn't be in the source code at this time, since there is a pure python implementation of murmur3. I'll make sure this is removed in the 3.7 release. Thanks!

PR #369, opened 2015-07-16
PYTHON-338 - DowngradingConsistencyRetry - only ignore write timeout if at least one response

aholmberg:


cassandra/policies.py
```diff
@@ -816,14 +816,19 @@ def on_write_timeout(self, query, consistency, write_type,
                          required_responses, received_responses, retry_num):
         if retry_num != 0:
             return (self.RETHROW, None)
-        elif write_type in (WriteType.SIMPLE, WriteType.BATCH, WriteType.COUNTER):
-            return (self.IGNORE, None)
+
+        if write_type in (WriteType.SIMPLE, WriteType.BATCH, WriteType.COUNTER):
```
mpenick: I don't think it's a good idea to retry a counter on write failure. It might be best to fail in this case.

aholmberg: I don't think this will retry counters. It's either ignore or rethrow.

mpenick: Ugh. Sorry, you're right.

PR #373, opened 2015-07-22
PYTHON-348 - Track previous value of columns at instantiation

kdeldycke:
The first unit-test (`test_previous_value_tracking_of_persisted_instance`) pass and the internal state of column's previous value and changed columns are properly tracked. The second (`test_previous_value_tracking_on_instanciation`), which run a scenario similar to the first, fail.

The difference ? The first test starts with a CQLengine instance that is already persisted to Cassandra (thanks to the use of `.create()`). The second test starts with an instance that hasn't been persisted yet.

I think we should change the way CQLengine operate on instantiation. So that both the internal `previous_value` property of each column and the object's `get_changed_columns()` method reflect the reality of an instance that is not yet persisted.

### DIFF
--- a/cassandra/cqlengine/models.py
+++ b/cassandra/cqlengine/models.py
@@ -357,25 +357,25 @@ class BaseModel(object):
     _table_name = None  # used internally to cache a derived table name
 
     def __init__(self, **values):
-        self._values = {}
         self._ttl = self.__default_ttl__
         self._timestamp = None
         self._transaction = None
+        self._batch = None
+        self._timeout = connection.NOT_SET
 
+        # A flag set by the deserializer to indicate that update should be used
+        # when persisting changes.
+        self._is_persisted = values.pop('_is_persisted', False)
+
+        # Initialize each volumn to its provided or default right at
+        # instanciation. See PYTHON-348.
+        self._values = {}
         for name, column in self._columns.items():
             value = values.get(name, None)
             if value is not None or isinstance(column, columns.BaseContainerColumn):
                 value = column.to_python(value)
             value_mngr = column.value_manager(self, column, value)
-            if name in values:
-                value_mngr.explicit = True
             self._values[name] = value_mngr
-
-        # a flag set by the deserializer to indicate
-        # that update should be used when persisting changes
-        self._is_persisted = False
-        self._batch = None
-        self._timeout = connection.NOT_SET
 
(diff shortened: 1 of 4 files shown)


NoKriK:
Any news? This looks like a good fix.

aholmberg:
I've been mostly wrapped up in supporting the Cassandra 3.0 release. A cqlengine-focused release will follow 3.0. Fortunately @kdeldycke has been patient as I have not had time to analyze the PR.
Ticket and discussion in JIRA: https://datastax-oss.atlassian.net/browse/PYTHON-348

cassandra/cqlengine/models.py
```diff
@@ -447,9 +447,7 @@ def _construct_instance(cls, values):
         else:
             klass = cls
 
-        instance = klass(**field_dict)
-        instance._is_persisted = True
-        return instance
+        return klass(_is_persisted=True, **field_dict)
```
aholmberg: Putting this in the initializer can collide with actual field names.

cassandra/cqlengine/models.py
```diff
@@ -357,26 +357,26 @@ class MultipleObjectsReturned(_MultipleObjectsReturned):
...
+        # Initialize each volumn to its provided or default right at
+        # instanciation. See PYTHON-348.
+        self._values = {}
         for name, column in self._columns.items():
             value = values.get(name, None)
             if value is not None or isinstance(column, columns.BaseContainerColumn):
                 value = column.to_python(value)
             value_mngr = column.value_manager(self, column, value)
-            if name in values:
-                value_mngr.explicit = True
```
aholmberg: I'm not sure about removing the explicit semantics. I think his removes the ability to see unset values when reading.

aholmberg: (perhaps more importantly, the ability to set a value to None even if it has a default)

kdeldycke: Yes, I also worry about all the None/empty/null edge cases but I was pretty sure I didn't messed that up. Anyway, #433 seems to address this very issue.

kdeldycke:
Should we close this PR now that you reworked it in #433 ?

aholmberg:
It closes implicitly when the commits are contained in another.

kdeldycke:
Good, didn't know GitHub handled cascading merge ! :)

PR #379, opened 2015-08-04
PYTHON-367 - Mitigate timeouts and resource expansion during full cluster outage

aholmberg:


cassandra/pool.py
```diff
@@ -572,14 +573,16 @@ def return_connection(self, connection):
             in_flight = connection.in_flight
 
         if connection.is_defunct or connection.is_closed:
```
mpenick: Seems like a lot of the same code as in `HostConnection` does it make sense to merge any of this into a `Connection` method?

cassandra/connection.py
```diff
@@ -360,6 +368,28 @@ def error_all_requests(self, exc):
...
+        # want to tie up the event thread when there are many requests
+        def err_all_callbacks():
+            for cb, _ in requests.values():
+                try_callback(cb)
+        if len(requests) < Connection.CALLBACK_ERR_THREAD_THRESHOLD:
+            err_all_callbacks()
+        else:
+            # daemon thread here because we want to stay decoupled from the cluster TPE
+            # TODO: would it make sense to just have a driver-global TPE?
+            t = Thread(target=err_all_callbacks)
```
mpenick: Any concern about a flood of new threads if a large number of nodes go down or become unreachable?

aholmberg: It's a good thing to consider. I thought about it and here is my reasoning:
- nodes down should be relatively infrequent, and thread overhead is not a concern
- if there is a sudden loss of connection, the connection errors will be processed serially by the event thread
- only <node count> threads *may* be created
- clusters with higher node count are very unlikely to have high concurrent requests per connection (thus the threshold for starting a thread will not be breached)

aholmberg: I guess another option would be a lazily-constructed TPE for connections. I don't think that's worth the complexity based on the above reasoning. Interested to hear what you think.

cassandra/pool.py
aholmberg: Yeah. That's unfortunate. I put this flagging and logic at the pool level to avoid Connection having to know about Cluster and pooling logic. They are not quite the same because of the difference around replacing connections.

cassandra/connection.py
mpenick: Good points. It would have to be a conspiracy of perfectly aligned circumstances for this to be an issue.

PR #380, opened 2015-08-05
Update README.rst

jeremyschlatter:
This asterisk is no longer needed. It was for a footnote that was deleted in b47cbd627924fc574660fa81c67b175f733b98d1


aholmberg:
Thanks!

PR #381, opened 2015-08-06
PYTHON-363 - Pure Python Murmur3 Implementation

aholmberg:


cassandra/murmur3.py
```diff
@@ -0,0 +1,110 @@
...
+    l = len(data)
+    nblocks = l // 16
+    tail = l % 16
+    if nblocks:
+        return struct.unpack_from('qq' * nblocks, data), struct.unpack_from('b' * tail, data, -tail), l
+    else:
+        return tuple(), struct.unpack_from('b' * tail, data, -tail), l
+
+
+def rotl64(x, r):
```
mpenick: Probably not important because nothing "rotl"s anything more than 33. Anything that >= 64 is incorrect.
>>> hex(rotl64(1, 64))
'0x10000000000000001L'

cassandra/murmur3.py
```diff
@@ -0,0 +1,110 @@
...
+    mask = 2 ** r - 1
+    rotated = (x << r) | ((x >> 64 - r) & mask)
```
mpenick: The left shift doesn't mask off bits that extend beyond 64 bits. Example:

>>> hex(rotl64(0x8000000000000000, 1))
'0x10000000000000001L'
should be: 
'0x1L'

Does this present a problem? Or does the final truncation take care of this?

mpenick: I don't think this matters.

aholmberg: You are right. I do the minimal work to make it correct for this use. I will add a note that these are not be general-purpose routines.

PR #382, opened 2015-08-10
Cython-based message deserialization (or, how to speed up my python)

markflorisson:
This PR adds accelerated versions of ProtocolHandler for message deserialization using Cython. By default the CythonProtocolHandler will be used if python-driver is compiled with Cython.

This work has been generously supported by SurveyMonkey.


sontek:
Here is a profile of a standard app we have:


Before:
```
time python single.py 
2119498

real	1m32.361s
user	1m28.183s
sys	0m1.950s
```

After:
```
$ time python single.py 
2148066

real	0m11.417s
user	0m5.469s
sys	0m1.910s
```

Screenshot of the profiler after this change, it looks beautiful:

screenshot from 2015-08-10 10 05 07

sontek:
@aholmberg @thobbs What you guys think about this?

aholmberg:
@sontek Looks pretty good to me. I've been looking at the work in progress so there's not too much surprising. Will be reviewing and testing here today.

cassandra/buffer.pxd
```diff
@@ -0,0 +1,45 @@
...
+buffer is in use.
+"""
+
+from cpython.bytes cimport PyBytes_AS_STRING
+    # char* PyBytes_AS_STRING(object string)
+    # Macro form of PyBytes_AsString() but without error
+    # checking. Only string objects are supported; no Unicode objects
+    # should be passed.
+
+from cassandra.buffer cimport Buffer
```
aholmberg: What does this do when we're already in the module?

cassandra/cython_utils.pyx
```diff
@@ -0,0 +1,29 @@
...
+        # You have to call it before any usage of DateTime CAPI functions.
+    )
+
+import datetime
+import sys
+
+DATETIME_EPOC = datetime.datetime(1970, 1, 1)
+
+assert sys.byteorder in ('little', 'big')
+is_little_endian = sys.byteorder == 'little'
```
aholmberg: Is this redundant with cassandra.util.is_little_endian?

tests/unit/cython/utils.py
```diff
@@ -0,0 +1,24 @@
...
+except ImportError:
+    import unittest  # noqa
+
+def cyimport(import_path):
+    """
+    Import a Cython module if available, otherwise return None
+    (and skip any relevant tests).
+    """
+    try:
+        return __import__(import_path, fromlist=True)
```
aholmberg: This does not work in PyPy. Would ``return __import__(import_path, fromlist=[True])`` have the desired effect? This seems to work in all runtimes I tested.

cassandra/cython_marshal.pyx
```diff
@@ -0,0 +1,221 @@
...
+
+# Special case for cassandra header
+header_struct = struct.Struct('>BBbB')
+header_pack = header_struct.pack
+header_unpack = header_struct.unpack
+
+# in protocol version 3 and higher, the stream ID is two bytes
+v3_header_struct = struct.Struct('>BBhB')
+v3_header_pack = v3_header_struct.pack
+v3_header_unpack = v3_header_struct.unpack
```
aholmberg: Are these actually used somehow?

cassandra/cython_marshal.pyx
```diff
@@ -0,0 +1,221 @@
...
+
+
+cdef varint_unpack(term):
+    """Unpack a variable-sized integer"""
+    if PY3:
+        return varint_unpack_py3(term)
+    else:
+        return varint_unpack_py2(term)
+
+# TODO: Optimize these two functions
```
aholmberg: Do you have something specific in mind, or do you just recognize an opportunity here?

cassandra/numpyparser.pyx
```diff
@@ -0,0 +1,174 @@
...
+
+        rowcount = read_int(reader)
+        array_descs, arrays = make_arrays(desc, rowcount)
+        arrs = &array_descs[0]
+
+        _parse_rows(reader, desc, arrs, rowcount)
+
+        arrays = [make_native_byteorder(arr) for arr in arrays]
+        result = dict(zip(desc.colnames, arrays))
+        # return pd.DataFrame(result)
```
aholmberg: leftover from earlier experimentation?

cassandra/typecodes.py
```diff
@@ -0,0 +1,62 @@
+"""
+Module with constants for Cassandra type codes.
+
+These constants are useful for
+
+    a) mapping messages to cqltypes                 (cassandra/cqltypes.py)
+    b) optimizezd dispatching for (de)serialization (cassandra/encoding.py)
```
aholmberg: typo "optimized"

setup.py
```diff
@@ -262,11 +262,16 @@ def run_setup(extensions):
 if "--no-cython" not in sys.argv:
     try:
         from Cython.Build import cythonize
-        cython_candidates = ['cluster', 'concurrent', 'connection', 'cqltypes', 'marshal', 'metadata', 'pool', 'protocol', 'query', 'util']
-        compile_args = [] if is_windows else ['-Wno-unused-function']
-        extensions.extend(cythonize(
-            [Extension('cassandra.%s' % m, ['cassandra/%s.py' % m], extra_compile_args=compile_args) for m in cython_candidates],
-            exclude_failures=True))
+        # cython_candidates = ['cluster', 'concurrent', 'connection', 'cqltypes', 'metadata',
```
aholmberg: Are you suggesting that we should not cythonize the core files? My trials showed 3-10% improvement for write-intensive workloads (where SerDes does not dominate).

aholmberg:
Thanks very much for this contribution. It looks good overall. I put a few minor questions and comments in the diff.

Some other general stuff (I can handle on commit if you don't care to):
- Possibly rename modules to include "_" for two-word names.
- Prepend copyright header in all files
- Add API docs for ProtocolHandler classes

cassandra/protocol.py
```diff
@@ -1015,6 +994,65 @@ def decode_message(cls, protocol_version, user_type_map, stream_id, flags, opcod
...
+        2. objparser.LazyParser
+            this parser decodes result messages lazily by returning an iterator
+
+        3. numpyparser.NumPyParser
+            this parser decodes result messages into NumPy arrays
+
+    The default is to use objparser.ListParser
+    """
+    # TODO: It may be cleaner to turn ProtocolHandler and ResultMessage into
+    # TODO:     instances and use methods instead of class methods
```
aholmberg: We should probably decide on this before introducing the API. I can do the refactor if you think it's worth it.

cassandra/cython_marshal.pyx
markflorisson: The trouble here is that static function calls are faster as they don't have to build a argument tuple. I'll make them cdef functions to illustrate this. Thanks!

setup.py
markflorisson: Sorry leftover comment since it takes forever to compile :)

cassandra/protocol.py
markflorisson: It's probably nice for dynamic instantiations of ProtocolHandler such as the Cython one, as the alternative is to use a closure which feels a little out of place. Are you planning to expose ProtocolHandler publicly?

aholmberg: Well, it's exposed to the degree that it's specified in [this config parameter](https://github.com/markflorisson/python-driver/blob/9240c71c02c27a6f20c73870a92ccdafab238df7/cassandra/cluster.py#L1524). This pattern was chosen for simplicity there, and to match the connection_class parameter, which supports a similar pluggable approach using class types. 
I could be convinced either way.

markflorisson:
I added some API documentation (maybe we want some more in the official docs?) and did some cleanups, thanks for all the suggestions @aholmberg.

cassandra/protocol.py
markflorisson: I see. Either way works for me, do whatever feels best, as the closure
itself isn't a problem. I'll remove the leftover TODOs.

coldeasy:
Is this intended to work with python 2.7? Building locally fails for me on this version, but succeeds with python3.

aholmberg:
@coldeasy yes, it's intended to work with Python 2.7. That's the primary runtime I've been using to test. What are the symptoms on failure?

coldeasy:
Ah looks like I had an old version (0.19.2) of cython installed. Installing latest (0.23) fixed the issue. For reference, this was the error:
```bash
$ python2 setup.py build
Compiling cassandra/cython_marshal.pyx because it changed.
Compiling cassandra/cython_utils.pyx because it changed.
Compiling cassandra/deserializers.pyx because it changed.
Compiling cassandra/ioutils.pyx because it changed.
Compiling cassandra/numpy_parser.pyx because it changed.
Compiling cassandra/obj_parser.pyx because it changed.
Compiling cassandra/parsing.pyx because it changed.
Compiling cassandra/row_parser.pyx because it changed.
Cythonizing cassandra/cython_marshal.pyx

Error compiling Cython file:
------------------------------------------------------------
...
        return varint_unpack_py3(to_bytes(term))
    else:
        return varint_unpack_py2(to_bytes(term))

# TODO: Optimize these two functions
cdef varint_unpack_py3(bytes term):
    ^
------------------------------------------------------------

cassandra/cython_marshal.pyx:107:5: closures inside cdef functions not yet supported
```

aholmberg:
Thanks for following up.

On Tue, Aug 11, 2015 at 4:47 PM, Colin Deasy <notifications@github.com>
wrote:

> Ah looks like I had an old version (0.19.2) of cython installed.
> Installing latest (0.23) fixed the issue. For reference, this was the error:
>
> $ python2 setup.py build
> Compiling cassandra/cython_marshal.pyx because it changed.
> Compiling cassandra/cython_utils.pyx because it changed.
> Compiling cassandra/deserializers.pyx because it changed.
> Compiling cassandra/ioutils.pyx because it changed.
> Compiling cassandra/numpy_parser.pyx because it changed.
> Compiling cassandra/obj_parser.pyx because it changed.
> Compiling cassandra/parsing.pyx because it changed.
> Compiling cassandra/row_parser.pyx because it changed.
> Cythonizing cassandra/cython_marshal.pyx
>
> Error compiling Cython file:
> ------------------------------------------------------------
> ...
>         return varint_unpack_py3(to_bytes(term))
>     else:
>         return varint_unpack_py2(to_bytes(term))
> # TODO: Optimize these two functions
> cdef varint_unpack_py3(bytes term):
>     ^
> ------------------------------------------------------------
>
> cassandra/cython_marshal.pyx:107:5: closures inside cdef functions not yet supported
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/datastax/python-driver/pull/382#issuecomment-130088086>
> .
>



-- 
<http://cassandrasummit-datastax.com/?utm_campaign=summit15&utm_medium=summiticon&utm_source=emailsignature>

aholmberg:
+1 here. Will merge pending input from our Test Eng stakeholders.

I have a small tweak I would like to introduce to facilitate tox unit testing, but that does not need to be done on this PR.

Thanks again for the contribution!

PR #387, opened 2015-08-18
Futures from pypi isn't py3 compatible

sontek:
Python3 ships with concurrent.futures so it shouldn't be pulled in but with latest concurrent.futures release on pypi they broke py3 compatibility:

https://github.com/agronholm/pythonfutures/issues/41

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,9 @@ from distutils.errors import (CCompilerError, DistutilsPlatformError,
                               DistutilsExecError)
 from distutils.cmd import Command
 
+py_version = sys.version_info[:2]
+PY3 = py_version[0] == 3
+
 try:
     import subprocess
     has_subprocess = True
@@ -297,7 +300,10 @@ def run_setup(extensions):
     if try_cython:
         kw['setup_requires'] = ['Cython']
 
-    dependencies = ['futures', 'six >=1.6']
+    dependencies = ['six >=1.6']
+
+    if not PY3:
+        dependencies.append('futures')
 
     setup(
         name='cassandra-driver',


aholmberg:
Interesting. I guess it previously just worked and a recent change introduced some Python 2-specific syntax?

sontek:
Yeah they introduced the problem in their latest release but I don't think you would ever want to install the backport library on Python 3 even if t did work, you would always want to be using the stdlib

aholmberg:
Totally agree. Had not really considered it since it wasn't causing issues.
Thanks for bringing it up.

On Tue, Aug 18, 2015 at 10:43 AM, John Anderson <notifications@github.com>
wrote:

> Yeah they introduced the problem in their latest release but I don't think
> you would ever want to install the backport library on Python 3 even if t
> did work, you would always want to be using the stdlib
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/datastax/python-driver/pull/387#issuecomment-132254297>
> .
>



-- 
<http://cassandrasummit-datastax.com/?utm_campaign=summit15&utm_medium=summiticon&utm_source=emailsignature>

multani:
Hi,

is there anything blocking this pull request to be merged? The driver currently doesn't work on Python 3 due to this bug (and the one in futures, first). I'd love to see a new release with this update!

sontek:
@multani As a temporary fix you can pin the version of futures so it pulls in a version that is compatible with python3

multani:
Ok, so it looks like it has to be pinned to 2.2.0 to work.

aholmberg:
@multani does this actually block installation or cause runtime errors for you? In my observation it was emitting the syntax error during egg_info for futures. pip was swallowing the error and returning success, and the driver was operational because of the stdlib implementation. Therefore I was not considering this a blocker, and planning to pick it up in the next major release.
If there is more to it, I welcome your input.

multani:
@aholmberg This doesn't prevent installation, although there's a warning emitted, but it clearly prevents usage when running Python 3:

```
>>> import cassandra.cluster
  File "<console>", line 1, in <module>
  File "cassandra/cluster.py", line 23, in init cassandra.cluster (cassandra/cluster.c:74118)
  File "/home/jballet/work/edgelab/repositories/edgelab-cassandra-cleanup/eggs/futures-3.0.3-py3.4.egg/concurrent/futures/__init__.py", line 8, in <module>
    from concurrent.futures._base import (FIRST_COMPLETED,
  File "/home/jballet/work/edgelab/repositories/edgelab-cassandra-cleanup/eggs/futures-3.0.3-py3.4.egg/concurrent/futures/_base.py", line 355
    raise type(self._exception), self._exception, self._traceback
                               ^
SyntaxError: invalid syntax
```

aholmberg:
Odd. My environment shows that error during install, but does not prevent it from being used (the stdlib futures is used). It must be a difference in versions. What are your python+pip+setuptools versions?

sontek:
@aholmberg This is the exact error I was getting that triggered me to open up the PR, for example:

```
$ mktmpenv -p /usr/bin/python3.4
$ pip install cassandra-driver
$ pip freeze|grep futures
futures==3.0.3
```

You can see it installed futures in python3 and that futures isn't compatible with python3.  So any usage of futures within the cassandra driver will now be broken.

aholmberg:
I get it. I'm just trying to understand why it's left in a working condition in my environment, and not in others.

multani:
@sontek Do you still have the error even in a new virtualenv?
@aholmberg @sontek  Actually, I tried to reproduce the problem with a fresh virtualenv using ``python3 -m venv --without-pip`` + downloading the latest pip and I get the behavior of @aholmberg where ``concurrent.futures`` from the system is used instead.

I got the problem when trying to install the library in a Buildout environment, and clearly, Buildout put all the installed dependencies at the beginning of ``sys.path`` which means the buggy version of ``futures`` is going to be used instead of the system's one.

sontek:
@multani You are right, I can't reproduce the error with a fully released version (because of sys.path):

```
mktmpenv -p /usr/bin/python3.4
Running virtualenv with interpreter /usr/bin/python3.4
Using base prefix '/usr'
New python executable in tmp-8663030927bd3f94/bin/python3.4
Also creating executable in tmp-8663030927bd3f94/bin/python
Installing setuptools, pip...done.
This is a temporary environment. It will be deleted when you run 'deactivate'.
(tmp-8663030927bd3f94)~/venvs/tmp-8663030927bd3f94
[sontek@aladdin]$ pip install futures
You are using pip version 6.0.8, however version 7.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting futures
  Using cached futures-3.0.3.tar.gz
Installing collected packages: futures
  Running setup.py install for futures
Successfully installed futures-3.0.3
(tmp-8663030927bd3f94)~/venvs/tmp-8663030927bd3f94
[sontek@aladdin]$ python
Python 3.4.3 (default, Mar 26 2015, 22:03:40) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from concurrent.futures import ThreadPoolExecutor
```

but I've definitely seen the error, just don't know how to make it happen =)

aholmberg:
Good observations. Thanks for the input.

multani:
If needed, I can provide a Buildout script which install the required dependencies, if that helps.

maxmcorp:
i had the problem under python3-4 too. Solved it  by setting futures directly in my setup.py for my project.

requires = [
    'pyramid',
    'pyramid_chameleon',
    'pyramid_debugtoolbar',
    'waitress',
    'bleach',
    'boto3',
    'cassandra-driver==2.7.1',
    'futures==2.0.0', # the temporary fix
    ]

maxmcorp:
Running on python 3 I made this temporary fix to get the right futures on the sys.path. According to your comments. It might be useful for somebody else googling the problem:

import sys
sys.path = [p for p in sys.path if not p.endswith('futures-3.0.3-py3.4.egg')]

After that it runs without that error message.

PR #388, opened 2015-08-18
Fixing a few minor bugs in our tests for 2.7 release

GregBestland:
Our control connection test should not run against anything with protocol level 3, as UDT will not be supported.

Updated our Metadata tests to correction check for explicitly defined thresholds on compaction strategies.

### DIFF
--- a/tests/integration/standard/test_control_connection.py
+++ b/tests/integration/standard/test_control_connection.py
@@ -34,6 +34,10 @@ def setup_module():
 
 class ControlConnectionTests(unittest.TestCase):
     def setUp(self):
+        if PROTOCOL_VERSION < 3:
+            raise unittest.SkipTest(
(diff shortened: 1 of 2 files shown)


tests/integration/standard/test_control_connection.py
```diff
@@ -34,6 +34,10 @@ def setup_module():
 
 class ControlConnectionTests(unittest.TestCase):
     def setUp(self):
+        if PROTOCOL_VERSION < 3:
+            raise unittest.SkipTest(
+                "Native protocol 3,0+ is required for UDTs using %r"
+                % (PROTOCOL_VERSION,))
```
aholmberg: This skips all tests in the class for protocol version < 3. Did you intend to raise from a specific test?

aholmberg:
+1

PR #389, opened 2015-08-20
New: AbstractQuerySet.__Iter__ is no more greedy

alanjds:
On https://datastax.github.io/python-driver/query_paging.html#handling-paged-results is told how to iterate ove large querysets since 2.6.0 .

However, the cqlengine querysets are fetching all pages before yielding the first result. This leads to timeouts or memory inflation on clients.

This PR makes __ iter__ become a non-greedy generator, yielding until the actual page is exhausted, and then fetching the next one.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2965,7 +2965,9 @@ class ResponseFuture(object):
 
         .. versionadded:: 2.0.0
         """
+        log.debug('Pagination: Fetching next page')
         if not self._paging_state:
+            log.debug('Pagination: No more pages to fetch')
             raise QueryExhausted()
 
         self._make_query_plan()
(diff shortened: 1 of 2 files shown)


aholmberg:
Thanks a lot for the contribution! We will have a cqlengine-focused release next, so I'll be looking at this then.
I've added a reference to this from an existing JIRA for tracking: https://datastax-oss.atlassian.net/browse/PYTHON-323

alanjds:
Thanks for the information, @aholmberg.

However, I still think that this is not an "improvement", but a "bugfix". I really would expect __ iter __ to fetch on demand. Otherwise it leads to the pointed annoyances: Timeouts and memory inflation.

aboudreault:
Thank you @alanjds for your contribution. Your work has been included in #323. I also added some perf improvements in __iter__ and __getitem__ to avoid using count() when possible (that was making iteration slow with large dataset). Feel free to comments.

PR #392, opened 2015-09-07
The default protocol version seems to be 4, not 2

bbirand:
But I imagine paging is also enabled when for protocol version 4?


aholmberg:
Thanks for the correction!

PR #394, opened 2015-09-08
PYTHON-397 Adding requirements for gevent and eventlet in test.

GregBestland:
Adding gevent, and eventlet requirements to the test-requirements.txt. We have unit tests that require them to run, and I think it's fair to require them for validation purposes.


aholmberg:
+1

PR #397, opened 2015-09-10
PYTHON-396 - Support build with older Cython and setup tools

aholmberg:


setup.py
```diff
@@ -297,7 +318,7 @@ def run_setup(extensions):
     kw['ext_modules'] = [Extension('DUMMY', [])]  # dummy extension makes sure build_ext is called for install
 
     if try_cython:
-        kw['setup_requires'] = ['Cython >=0.21']
+        kw['setup_requires'] = ['Cython']
```
GregBestland: My only question would be, are there other code compatibility issues with older versions (pre .20) of cython we need to be aware of? Maybe we should still specify a minimum version of Cython?

GregBestland:
Just the one question, otherwise i'm +1 on this. Clever fix.

PR #408, opened 2015-09-30
Adding tests for python-419

GregBestland:
Added a few tests to reproduce and test for the issues in python-419

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -25,7 +25,7 @@ build:
         sudo apt-get install -y libev4 libev-dev
       fi
       pip install -r test-requirements.txt
-
+      pip install nose-ignore-docstring
 
       if [[ $CYTHON == 'CYTHON' ]]; then
         pip install cython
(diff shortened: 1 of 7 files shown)


build.yaml
```diff
@@ -25,7 +25,7 @@ build:
         sudo apt-get install -y libev4 libev-dev
       fi
       pip install -r test-requirements.txt
-
+      pip install nose-ignore-docstring
```
aholmberg: yay

tests/integration/__init__.py
```diff
@@ -332,3 +332,85 @@ def wait_for_down(self):
...
+
+
+class BasicKeyspaceUnitTestCase(unittest.TestCase):
+    """
+    This is basic unit test case that can be leveraged to scope a keyspace to a specific unit test.
+    creates a keyspace named after the testclass with a rf of 1
+    """
+    @classmethod
+    def setUpClass(self):
+        self.commonSetUp('1')
```
aholmberg: Minor nit, but it seems more natural to pass a number for rf and let the method take care of formatting.

tests/integration/__init__.py
```diff
@@ -332,3 +332,85 @@ def wait_for_down(self):
...
+    def tearDownClass(self):
+        execute_until_pass(self.session, "DROP KEYSPACE {0}".format(self.ksname))
+        self.cluster.shutdown()
+
+    @classmethod
+    def commonSetUp(self, rf, create_table=False, skip_if_cass_version_less_than=None):
+        self.ksname = self.__class__.__name__
+        self.cluster = Cluster(protocol_version=PROTOCOL_VERSION)
+        self.session = self.cluster.connect()
+        print("Creating keyspace {0} with rf of {1}".format(self.ksname, rf))
```
aholmberg: Can we log instead of print?

tests/integration/__init__.py
```diff
@@ -332,3 +332,85 @@ def wait_for_down(self):
...
+    This is basic unit test case that can be leveraged to scope a keyspace to a specific unit test.
+    creates a keyspace named after the testclass with a rf of 1
+    """
+    @classmethod
+    def setUpClass(self):
+        self.commonSetUp('1')
+
+    @classmethod
+    def tearDownClass(self):
+        execute_until_pass(self.session, "DROP KEYSPACE {0}".format(self.ksname))
```
aholmberg: If we assume we have ksname, I think we should make it a property of this base class. May also consider providing a fname property based on function name for tables/types/etc. These can be lifted from standard.test_metadata classes.

tests/integration/cqlengine/test_ttl.py
```diff
@@ -19,12 +25,13 @@
...
+from tests.integration import CASSANDRA_VERSION
 
 
 class TestTTLModel(Model):
-    id      = columns.UUID(primary_key=True, default=lambda:uuid4())
-    count   = columns.Integer()
-    text    = columns.Text(required=False)
+    id = columns.UUID(primary_key=True, default=lambda: uuid4())
+    count = columns.Integer()
+    text = columns.Text(required=False)
```
aholmberg: thank you! Do not like that kind of formatting (former). :)

tests/integration/cqlengine/test_ttl.py
```diff
@@ -151,6 +156,7 @@ def test_ttl_included_with_blind_update(self):
         self.assertIn("USING TTL", query)
 
 
+@unittest.skipIf(CASSANDRA_VERSION < '2.0', "default_time_to_Live was introduce in C* 2.0, currently running {0}".format(CASSANDRA_VERSION))
```
aholmberg: Did you come to an understanding of how this works between unittest/unitest2 and runtimes?

tests/integration/standard/test_metadata.py
```diff
@@ -2111,6 +2113,60 @@ def test_create_view_metadata(self):
...
+
+        self.session.execute(alter_scores)
+        self.assertEqual(len(self.cluster.metadata.keyspaces[self.ksname].views), 1)
+
+        score_column = self.cluster.metadata.keyspaces[self.ksname].tables['scores'].columns['score']
+        self.assertEquals(score_column.typestring, 'blob')
+
+        score_mv_column = self.cluster.metadata.keyspaces[self.ksname].views["monthlyhigh"].columns['score']
+        self.assertEquals(score_mv_column.typestring, 'blob')
+
```
aholmberg: We might want a test where the MV is ```select *``` and make sure adding a column to the base table makes it appear in the MV

aholmberg:
+1 pending comments on diff

tests/integration/__init__.py
GregBestland: That's an excellent idea.

PR #412, opened 2015-10-03
Fix timeout in geventreactor

yi719:


yi719:
I am working on a website with Cassandra. Once in a while I get an OperationTimedOut exception. 
This usually happens on a specific query where I select multiple rows from a specific table. 
That table has a column of type ascii that contains Json blob which can be up to 64k in length but is usually much smaller.

If I understand correctly: there seems to be a problem in handle_read with responses whose length is mutiple of the incoming buffer size(i.e. 4096bytes)

the is_timeout method does not work with gevent's socket timeout.

so maybe we need update the is_timeout  method

aholmberg:
Maybe they changed how a timeout is propagated. Can you specify your gevent version?

Also, is there still some issue with ``handle_read``? You mention it above, but the change set doesn't seem to address anything there.

yi719:
@aholmberg 
I use gevent==1.0.2.

After this change,  OperationTimedOut never show up.

When  responses whose length is mutiple of the incoming buffer size(i.e. 4096bytes),  soccket.timeout will raise. If we do not catch this error in such saturation, self.defunct will be called, then leave the read loop

aholmberg:
Thanks for the additional info. I'm still confused as to why the response size has something to do with socket.timeout. Is there another issue that needs resolution?

aholmberg:
Also, would you mind squashing this into a single commit?

yi719:
> I'm still confused as to why the response size has something to do with socket.timeout. Is there another issue that needs resolution?

In handle_read's while loop, we keep on reading from socket unless len(buf) < self.in_buffer_size, or when response's length is 4096bytes(Cassandra server do not response more data),  we will hit socket.timeout.

>Also, would you mind squashing this into a single commit?
Done

aholmberg:
Thanks for bearing with me. I think I have a complete understanding of the problem now. I've expounded in a follow-on ticket: https://datastax-oss.atlassian.net/browse/PYTHON-429

aholmberg:
Thanks for the contribution!

yi719:
Thanks for your support

PR #416, opened 2015-10-09
PYTHON-416 - default CL is now LOCAL_QUORUM

aholmberg:


cassandra/cluster.py
```diff
@@ -1474,13 +1474,17 @@ class Session(object):
     .. versionadded:: 2.0.0
     """
 
-    default_consistency_level = ConsistencyLevel.ONE
+    default_consistency_level = ConsistencyLevel.LOCAL_QUORUM
     """
```
GregBestland: There is a separate default consistency level in CQLEngine, connection.py do we need to update that as well?
located here.
https://github.com/datastax/python-driver/blob/master/cassandra/cqlengine/connection.py#L36

aholmberg: Yes! That, or remove two separate default CLs. Good catch.

PR #419, opened 2015-10-13
PYTHON-318 - return trace data on ResultSet

aholmberg:


cassandra/cluster.py
```diff
@@ -2930,7 +2909,7 @@ def _set_result(self, response):
                         self, **response.results)
                 else:
                     results = getattr(response, 'results', None)
-                    if results is not None and response.kind == RESULT_KIND_ROWS:
+                    if results is not None and response.kind ==RESULT_KIND_ROWS:
```
mpenick: What did that space ever do to you?

aholmberg: It was collateral damage in a refactor/revert that happened. My condolences.

PR #422, opened 2015-10-14
Fixing issue with ; delimeter on as_cql_query for keyspace. 

GregBestland:
Also changing
Changing all_as_cql to internal method to avoid confusion

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -656,7 +656,7 @@ class KeyspaceMetadata(object):
         ret = "CREATE KEYSPACE %s WITH replication = %s " % (
             protect_name(self.name),
             self.replication_strategy.export_for_schema())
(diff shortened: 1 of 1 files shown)


cassandra/metadata.py
```diff
@@ -656,7 +656,7 @@ def as_cql_query(self):
         ret = "CREATE KEYSPACE %s WITH replication = %s " % (
             protect_name(self.name),
             self.replication_strategy.export_for_schema())
-        return ret + (' AND durable_writes = %s;' % ("true" if self.durable_writes else "false"))
+        return ret + (' AND durable_writes = %s' % ("true" if self.durable_writes else "false"))
```
aholmberg: Now we're missing a semicolon after the keyspace cql in KeyspaceMetadata.export_as_string

aholmberg: (fixed in 405)

PR #423, opened 2015-10-15
PYTHON-178 - Make bind ignore extra keys in dict values

aholmberg:


cassandra/query.py
```diff
@@ -511,10 +512,8 @@ def bind(self, values):
...
                             (col.name))
 
-            value_len = len(values)
-
-            if unbound_values:
-                raise ValueError("Unexpected arguments provided to bind(): %s" % unbound_values.keys())
+        value_len = len(values)
+        col_meta_len = len(col_meta)
 
         if value_len > col_meta_len:
```
mpenick: I don't this this check is valid anymore because the content of  `values` is bounded by the `col_meta`.

aholmberg: I think it is valid if incoming ``values`` is not  a ``dict``

mpenick: ah, yeah

PR #426, opened 2015-10-19
416 test

GregBestland:
Making local quorum default for cqlengine. Adding test to ensure default consistency level from base driver and clqengine remain the same.


aholmberg:
@GregBestland let me know what you think ^

GregBestland:
This is a much cleaner solution then what I came with @aholmberg . I'm 100% +1. I should have thought of this.

aholmberg:
My fault for the oversight. Thanks for looking it over.

PR #427, opened 2015-10-19
Moving travis over to the new infrastructure

GregBestland:
This should get us off the legacy infrastructure and onto the new container infrastructure. Everything seems to work great with these changes.


aholmberg:
Cool! Thanks for looking into that.

PR #436, opened 2015-11-03
PYTHON-422 - update schema type representation in Cassandra 3.0+

aholmberg:


GregBestland:
I looked over this as thoroughly  as I can I couldn't find any problems.
I'm going to try to find some additional test cases to make sure we cover every possible metadata type in conjunction with function/aggregate/table/udt, but we may already have coverage.

+1 great job.

PR #438, opened 2015-11-03
PYTHON-378 - make NetworkTopologyStrategy consider rack placement

aholmberg:


cassandra/metadata.py
```diff
@@ -475,14 +475,16 @@ def make_token_replica_map(self, token_to_host_owner, ring):
         ring_len_range = range(ring_len)
```
mpenick: What the thinking behind having the variables:  `ring_len` and `ring_len_range`? They seem to be used once. Could the line there used on be simplified to?

```python
for i in range(len(ring)):
    pass
```

aholmberg: I guessed it was vestigial from an earlier implementation of the algorithm. I can change. I thought I did (as I also noticed it), but must have backed it out in my own rework.

cassandra/metadata.py
```diff
@@ -506,20 +508,33 @@ def make_token_replica_map(self, token_to_host_owner, ring):
                     index += 1
                 dc_to_current_index[dc] = index
 
-                # now add the next RF distinct token owners to the set of
-                # replicas for this DC
+                replicas_remaining = remaining[dc]
```
mpenick: `remaining = dc_rf_map.copy()` is unnecessary as it's not modified.

aholmberg: good eye. Thanks

PR #440, opened 2015-11-04
Fixes for test case issues.

GregBestland:
Adding a skip where needed, fixing issue with cross test pollution in cqlengine.

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -22,7 +22,7 @@ from threading import Event
 from subprocess import call
 from itertools import groupby
 
-from cassandra import OperationTimedOut, ReadTimeout, ReadFailure, WriteTimeout, WriteFailure
+from cassandra import OperationTimedOut, ReadTimeout, ReadFailure, WriteTimeout, WriteFailure, AlreadyExists
 from cassandra.cluster import Cluster
 from cassandra.protocol import ConfigurationException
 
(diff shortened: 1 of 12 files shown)


GregBestland:
Please hold merging until integration test suite has been run.

tests/integration/__init__.py
```diff
@@ -22,7 +22,7 @@
 from subprocess import call
 from itertools import groupby
 
-from cassandra import OperationTimedOut, ReadTimeout, ReadFailure, WriteTimeout, WriteFailure
+from cassandra import OperationTimedOut, ReadTimeout, ReadFailure, WriteTimeout, WriteFailure, AlreadyExists
 from cassandra.cluster import Cluster
 from cassandra.protocol import ConfigurationException
```
aholmberg: I don't think it will ever be ConfigurationException. This is only the type/name of the message.

tests/integration/__init__.py
```diff
@@ -262,7 +262,8 @@ def execute_until_pass(session, query):
     while tries < 100:
         try:
             return session.execute(query)
-        except ConfigurationException:
+        except (ConfigurationException, AlreadyExists):
+            log.warn("Recieved already exists from query {0}   not exiting".format(query))
```
aholmberg: "already exists" is not completely accurate. You would be handling "non existing" if it's ConfigurationException on a DROP

tests/integration/cqlengine/query/test_named.py
```diff
@@ -280,6 +281,7 @@ def setUp(self):
 
     def tearDown(self):
         models.DEFAULT_KEYSPACE = self.default_keyspace
+        setup_connection(models.DEFAULT_KEYSPACE)
```
aholmberg: I think calling this multiple times will leave dangling sessions (cqlengine.connection does not clean up by itself).

PR #441, opened 2015-11-04
Handle case where query trace returns nothing

carlyeks:
Currently it throws an exception because we are trying to get the 0th element, but if there are no results we should just say the trace isn't yet incomplete.

This can occur when the tracing stage is backed up and the response is sent before the session has been persisted.


aholmberg:
sorry about that. Thanks for the catch.

PR #443, opened 2015-11-12
Fixing issue where sessions can leak upon test failure.

GregBestland:
Also added a little more validation. Probably not a big deal, but I wanted to make sure new metadata was reflected, rather then just checking to see if it was refreshed.

### DIFF
--- a/tests/integration/standard/test_metadata.py
+++ b/tests/integration/standard/test_metadata.py
@@ -559,15 +559,16 @@ class SchemaMetadataTests(BasicSegregatedKeyspaceUnitTestCase):
         cluster2 = Cluster(protocol_version=PROTOCOL_VERSION, schema_event_refresh_window=-1)
         cluster2.connect()
 
-        self.assertNotIn("mv1", cluster2.metadata.keyspaces[self.keyspace_name].tables[self.function_table_name].views)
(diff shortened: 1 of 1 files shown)


tests/integration/standard/test_metadata.py
```diff
@@ -559,15 +559,18 @@ def test_refresh_metadata_for_mv(self):
...
-        cluster2.shutdown()
+        try:
+            self.assertNotIn("mv1", cluster2.metadata.keyspaces[self.keyspace_name].tables[self.function_table_name].views)
+            self.session.execute("CREATE MATERIALIZED VIEW {0}.mv1 AS SELECT b FROM {0}.{1} WHERE b IS NOT NULL PRIMARY KEY (a, b)"
+                                 .format(self.keyspace_name, self.function_table_name))
+            self.assertNotIn("mv1", cluster2.metadata.keyspaces[self.keyspace_name].tables[self.function_table_name].views)
+
+            cluster2.refresh_table_metadata(self.keyspace_name, "mv1")
+            self.assertIn("mv1", cluster2.metadata.keyspaces[self.keyspace_name].tables[self.function_table_name].views)
+        except:
```
aholmberg: no ```except``` necessary if you're just going to let it propagate. try/finally is good

PR #446, opened 2015-11-13
Adding test for python-334

GregBestland:


tests/integration/standard/test_query.py
```diff
@@ -386,6 +385,21 @@ def test_no_parameters_many_times(self):
             self.test_no_parameters()
             self.session.execute("TRUNCATE test3rf.test")
 
+    def test_unicode(self):
+        ddl = '''
+            CREATE TABLE test3rf.testtext (
```
aholmberg: should this be using any of the base keyspace or table properties since the test is now inheriting the common base?

tests/integration/standard/test_query.py
```diff
@@ -788,3 +802,38 @@ def test_mv_filtering(self):
         self.assertEquals(results[1].day, 25)
         self.assertEquals(results[1].score, 3200)
         self.assertEquals(results[1].user, "pcmanus")
+
+
+class UnicodeQueryTest(BasicSharedKeyspaceUnitTestCase):
+
+    def setUp(self):
+        ddl = '''
+            CREATE TABLE test3rf.testtext (
```
aholmberg: same question as above

aholmberg:
+1 with questions

tests/integration/standard/test_query.py
GregBestland: It most certainly should not. Good catch

GregBestland:
Comments addressed merging now.

PR #448, opened 2015-12-01
Add tornado reactor

coldeasy:
- Adds basic integration with tornado, creating a separate ioloop in a separate thread
- TODO add tests, cleanup

### DIFF
--- a/cassandra/connection.py
+++ b/cassandra/connection.py
@@ -343,6 +343,31 @@ class Connection(object):
             for args in self.sockopts:
                 self._socket.setsockopt(*args)
 
+    def _read_socket(self):
+        try:
+            while True:
+                buf = self._socket.recv(self.in_buffer_size)
+                self._iobuf.write(buf)
+                if len(buf) < self.in_buffer_size:
+                    break
+        except socket.error as err:
+            if self._ssl_impl and isinstance(err, self._ssl_impl.SSLError):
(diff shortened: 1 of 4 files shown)


aholmberg:
@coldeasy we always appreciate feature contributions, but we may not be able to accept a new IO reactor at this time. We have much to do, with limited resources, and new reactors multiply the surface area for test, maintenance and support.

The existing reactors (save Twisted) are maintained internally because each removes a barrier to entry for a broad class of applications (for example, gevent is required for applications with monkey patching active). I am not aware of anything in the Tornado framework that precludes using one of the existing core reactors inside the driver (e.g. asyncore or libev -- correct me if I'm wrong).

Is this reactor something you might consider maintaining in a separate contrib repository? We could reference that from the project docs if you're interested.

coldeasy:
@aholmberg Okay, I can understand that. I was hoping to eventually remove the need for threads entirely and just use tornado - and this was a first step.
I think your idea of maybe having a separate repo for this (or a fork depending on how significantly entwined the changes turn out to be) makes sense.

aholmberg:
@coldeasy thanks for your understanding on this. Sorry we're not able to take it on.
If you decide to stand up a contrib repo, definitely let us know on the mailing list. We can also collaborate if the Connection interface/contract needs to be further formalized to make this development a success.

PR #450, opened 2015-12-10
Pass through metrics to response future for prepared statements

coldeasy:
@aholmberg I'm not sure if there is a good reason to not pass metrics to `ResponseFuture` for prepared statements - if there is, I'll revert these changes and add a comment to that as it's a little confusing.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1705,7 +1705,13 @@ class Session(object):
         message. See :ref:`custom_payload`.
         """
         message = PrepareMessage(query=query)
-        future = ResponseFuture(self, message, query=None, timeout=self.default_timeout)
+        future = ResponseFuture(
+            self,
+            message,
+            query=None,
+            timeout=self.default_timeout
+            metrics=self._metrics
+        )
         try:
             future.send_request()
             query_id, column_metadata, pk_indexes = future.result()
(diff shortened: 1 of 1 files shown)


aholmberg:
It's a fair question. I didn't write that, but understanding is that it's excluded because it's not really part of the hot path of request execution. Most metrics observers are more concerned with the core request execution, not the preparatory work that normally happens in initialization. Put another way: metrics apply to csession.execute[_async] client requests, an not other ancillary requests.

coldeasy:
I guess that makes sense, though if ancillary requests did contribute to latency in some way it would be hard to pinpoint without metrics. I'd be in favour of exposing as much metrics as possible and letting the client determine what can be discarded.

aholmberg:
Sorry about the delay. This slipped off my radar somehow.

> I'd be in favour of exposing as much metrics as possible and letting the client determine what can be discarded.

The problem is I don't think there's a way to extract *some* data from the PMF stats.

Right now I would answer the prepare latency question by simply timing the Session.prepare call at the application level.

I'm just not sure we're solving a problem here. That said, These things are infrequent and putting them in the stats should be mostly inconsequential. I also don't feel too strongly against it so I will merge if you view it as useful.

coldeasy:
@aholmberg I'm going to close this for now - it's not an urgent need.

PR #453, opened 2015-12-16
PYTHON-454 Pass authenticator class name to Authenticator implementat…

aholmberg:
…ions

Replacing #452

### DIFF
--- a/cassandra/auth.py
+++ b/cassandra/auth.py
@@ -15,7 +15,6 @@ try:
 except ImportError:
     SASLClient = None
 
-
 class AuthProvider(object):
     """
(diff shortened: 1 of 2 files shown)


aholmberg:
@mike-tr-adamson I'm not sure I like changing a cluster-level auth attribute based on the response from a single connection. What would you think of me refactoring to hold the class per connection and pass to the Auth initial_response and evaluate_challenge? I think this will make it robust connecting to clusters with non-uniform settings.
(I'll do the work -- just raising the issue in case you had other thoughts)

aholmberg:
argh. That requires an API change.

PR #455, opened 2016-01-01
improve ExponentialReconnectionPolicy,now can custom max attempts time

vipjml:


cassandra/policies.py
```diff
@@ -528,7 +528,7 @@ class ExponentialReconnectionPolicy(ReconnectionPolicy):
     a set maximum delay.
     """
 
-    def __init__(self, base_delay, max_delay):
+    def __init__(self, base_delay, max_delay,max_attempts=64):
```
aholmberg: Please use PEP-8 spacing throughout.

cassandra/policies.py
```diff
@@ -528,7 +528,7 @@ class ExponentialReconnectionPolicy(ReconnectionPolicy):
...
         """
         `base_delay` and `max_delay` should be in floating point units of
         seconds.
```
aholmberg: Docstring explanation for max_attempts, similar to ```ConstanctReconnectionPolicy```

aholmberg:
Are you planning to add unit tests?

vipjml:
I add a unit test,fix spacing and docstring

cassandra/policies.py
```diff
@@ -528,10 +528,14 @@ class ExponentialReconnectionPolicy(ReconnectionPolicy):
     a set maximum delay.
     """
 
-    def __init__(self, base_delay, max_delay):
+    def __init__(self, base_delay, max_delay, max_attempts=None):
```
aholmberg: I think we need to keep the default 64 to match the previous behavior if the parameter is unspecified.

Does it ever make sense to try forever (None)? With exponential backoff, the retry interval gets very large, and might lead to situations where reconnects are not very responsive (with a retry scheduled, but *way* in the future).

As an aside, are you trying to solve a specific problem here?

aholmberg:
If you could, please rebase/squash so we don't have so many commits for one change. Alternatively, I can just apply a single patch in your name.

vipjml:
The reason why I want to change max_attempts=None is that, In some situation, the client use up all retries of all nodes, for example,cassandra down for hours,the client will never recover except you recreate a  cassandra.cluster.Cluster or restart the client program. A simple way to regenerate this: create a client with ExponentialReconnectionPolicy(1,2), close cassandra after client connected, then wait for a while, after the client use up all retries, start cassandra, but the client will never recover,keep throwing NoHostAvailable.

vipjml:
I change default max_attempts to 64, but I think default max_attempts-64 should be better for most users

aholmberg:
@vipjml I understand the point and I agree. All I'm trying to do is avoid changing default behavior in a minor version. I can add a TODO there to update on the next major rev.

vipjml:
that is well

vipjml:
this is well !

aholmberg:
Thanks for the contribution!

PR #457, opened 2016-01-07
PYTHON-456 - geometric types

aholmberg:


dse/cluster.py
```diff
@@ -0,0 +1,24 @@
+from cassandra.cluster import Cluster, Session
+import dse.cqltypes  # unsued here, imported to cause type registration
+from dse.util import Point, Circle, LineString, Polygon
+
+
+class Cluster(Cluster):
+
+    def _new_session(self):
```
mpenick: Does `_new_session()` need to be reimplemented in DSE cluster? The code looks the same.

dse/cqltypes.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class PointType(CassandraType):
+    typename = 'PointType'
+
+    _platform_point = point_le if _platform_is_le else point_be
+    _type = struct.pack('=BI', _endian_flag, WKBGeometryType.POINT)
+
+    @staticmethod
+    def serialize(val, protocol_version):
```
mpenick: Are you going to add Cython de/serialization for these types?

mpenick:
Open-ended question: Are users expecting the type `Point` for a (x y) pairs inside other types or a list of floats?

dse/cqltypes.py
aholmberg: That would be a follow-on ticket if I do.

aholmberg:
> Are users expecting the type Point for a (x y) pairs inside other types or a list of floats?

I don't think users have any expectations atm. I'm not expecting these built-n types to be used for much. Planning to add conversions to optional third-party libs later.

dse/cluster.py
aholmberg: Yes, because ```dse.cluster.Session``` is not ```cassandra.cluster.Session```. I could make this a bit clearer by calling it ```DSESession``` here.

mpenick:
+1 All sounds good. Thanks.

PR #459, opened 2016-01-08
PYTHON-450 - resolve alignment problems with cython marshalling

aholmberg:


mpenick:
At first I thought that the endianess check might have be compiled out, but it does not. It's possible to avoid checking endianness:  https://github.com/mpenick/python-driver/commit/85ac2f7e2ab3e533e652762092ef69d613bedb0e. There isn't any reason to check endianness and these loops should unroll too.

Really quick POC. I'm didn't actually run this yet.

mpenick:
Also, why not uses "memcpy" instead of implement a custom version?  Will something like this not work?

```python
from libc.string cimport memcpy
```

aholmberg:
> Also, why not uses "memcpy" instead of implementing a custom version?
The reasoning is in the docstring of the function.

mpenick:
Using 
```python
from libc.string cimport memcpy
memcpy(dst, src, sizeof(blah));
```
emits a regular `memcpy` call in the C code which is preferred because it's really a compiler intrinsic.

aholmberg:
memcpy: I was using that first, but then moved his memcopy from the numpy module (https://github.com/datastax/python-driver/commit/827f4604ee8c5a5aa25b0417e41776a096201e01), taking for granted that he tried it and it performed better. I did some actual benchmarks and memcpy is 25% faster, even for small types.
Will revert to memcpy. Thanks for challenging the assumption.

aholmberg:
>  There isn't any reason to check endianness

I think it's harmless. My first thought was that the condensed version has more arithmetic operations (and more cycles in the inner loop). I tried both and it turns out the difference is negligible since ~16 cycles << call overhead for each element.

I did notice that the alternative version is not returning correct values. All things equal, I'll probably just keep this version unless you care enough to look into it.

PR #466, opened 2016-01-28
Fix Python3 bug with Timers and heaps

jonmcevoy:
Add comparison operator to allow Timers with the same end time to be
pushed into heaps. It's poorly documented, but the warning can be found
here:
https://docs.python.org/2/library/heapq.html#priority-queue-implementation-notes
"In the future with Python 3, tuple comparison breaks for (priority,
task) pairs if the priorities are equal and the tasks do not have a
default comparison order."

### DIFF
--- a/cassandra/connection.py
+++ b/cassandra/connection.py
@@ -1020,6 +1020,9 @@ class Timer(object):
         if timeout < 0:
             self.callback()
 
+    def __lt__(self, other):
+        return self.end < other.end
+
     def cancel(self):
         self.canceled = True
 


aholmberg:
Good catch. Thanks for the fix.

aholmberg:
(ping for CLA)

jonmcevoy:
I've contacted cla@datastax.com with some follow-up requests. It shouldn't take long to clear up. The issue being this is work of Trion Worlds, Inc (my employer) so they need to review the agreement.

kohei-cbd:
I was about to submit a pull request for the same fix, then I discovered this pending pull request.

Glad to see this being addressed.

aboudreault:
@jonmcevoy Any news from the cla?

jonmcevoy:
We're all good now. Sorry for the delay.

aboudreault:
Thanks!

kohei-cbd:
Any chance we could get this backported to the 3.1 branch?

aholmberg:
Yes! We were just talking about that internally. I'll get it there today.

kohei-cbd:
@aholmberg Thanks!  Much appreciated.

PR #467, opened 2016-02-02
Fix: crash while working with UDTs

reeteshranjan:
In some cases the ``value`` attribute is ``None``, and without checking we are accessing a method assuming it's a valid object. This fix helps achieving the desired functionality/outcome for such cases without breaking the intended logic.


aholmberg:
@aboudreault would you mind taking a look at this as part of your work on cqlengine?

aboudreault:
@aholmberg Sure. Will do asap.

aboudreault:
Thanks @reeteshranjan. I can confirm the issue. The patch LGTM. We can merge this PR. I am going to create another PR with the test.

PR #468, opened 2016-02-03
PYTHON-354 - DateType.deserialize being called with one argument vs two

aboudreault:
* Remove the DateType.deserialized() use since the mapper should not deal with serialized data
* Add DateTime tests for a timestamps and invalid inputs

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -18,7 +18,7 @@ import logging
 import six
 
 from cassandra import util
-from cassandra.cqltypes import DateType, SimpleDateType
+from cassandra.cqltypes import SimpleDateType
 from cassandra.cqlengine import ValidationError
 from cassandra.cqlengine.functions import get_total_seconds
 
@@ -435,10 +435,8 @@ class DateTime(Column):
             return value
         elif isinstance(value, date):
             return datetime(*(value.timetuple()[:6]))
-        try:
-            return datetime.utcfromtimestamp(value)
-        except TypeError:
-            return datetime.utcfromtimestamp(DateType.deserialize(value))
+
+        return datetime.utcfromtimestamp(value)
 
     def to_database(self, value):
         value = super(DateTime, self).to_database(value)
--- a/tests/integration/cqlengine/columns/test_validation.py
+++ b/tests/integration/cqlengine/columns/test_validation.py
@@ -88,6 +88,17 @@ class TestDatetime(BaseCassEngTestCase):
         dts = self.DatetimeTest.objects.filter(test_id=1).values_list('created_at')
         assert dts[0][0] is None
 
+    def test_datetime_invalid(self):
+        dt_value= 'INVALID'
+        with self.assertRaises(TypeError):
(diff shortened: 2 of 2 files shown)


aboudreault:
I've noticed that c* returns timestamp as offset-naive datetime. So the TZ used is probably the one configuredon the node. I wonder if this could be a problem since iso8601.parse_date() always use UTC by default.

aboudreault:
ah, based on the fact that we use "datetime.utcfromtimestamp" in that code base, we make the same assumption about UTC so it shouldn't be a problem.

aholmberg:
Yes. Timestamps in the database (and client) are timezone-agnostic
intentionally. We have https://datastax-oss.atlassian.net/browse/PYTHON-394
to write about that somewhere.

cassandra/cqlengine/columns.py
```diff
@@ -16,6 +16,8 @@
 from datetime import date, datetime
 import logging
 import six
+import struct
+import iso8601
```
aholmberg: I know the OP requested this, but I'm luke warm on adding a new dep for date parsing. We usually leave things like this to the application layer. The main reason is that we want a lean dependency list because of the cqlsh integration (Cassandra has to carry along tarballs of all dependencies). This is not stated anywhere, so maybe should I should add some discussion to the contribution doc.

All of the above said, the main constraint does not apply to the mapper layer, so we could actually get away with it here. I just want to point it out to make sure it's warranted, and make sure we don't leak it into the core. We also have to consider then that we are saying string dates are ok for the mapping layer but not for core.

cassandra/cqlengine/columns.py
```diff
@@ -435,10 +437,20 @@ def to_python(self, value):
...
+                return iso8601.parse_date(value)
+            except iso8601.ParseError:
+                pass
+
         try:
             return datetime.utcfromtimestamp(value)
         except TypeError:
-            return datetime.utcfromtimestamp(DateType.deserialize(value))
+            try:
+                return DateType.deserialize(value, 0)
```
aholmberg: I can't figure out why this was added.
https://github.com/datastax/python-driver/commit/70d866018d3fc56e960d0f7746a5a6d33f8aadd8
I don't know why we would ever be working with serialized values in the mapping layer.
There is no test covering this line. I'm leaning toward removing it and letting a TypeError propagate in this case.
(lmkwyt)

cassandra/cqlengine/columns.py
aboudreault: ah. Now having in mind that C* has to bundle  all deps, I agree that adding a new dependency just for this might be unnecessary. I'm fine to let this part handled on the application level.

cassandra/cqlengine/columns.py
aboudreault: +1

aboudreault:
I am going to modify this PR with the following changes:

* No more support for direct iso8601 date strings due to dependency addition
* Remove the DateType.deserialized() use since the mapper should not deal with serialized data
* Modify tests added to only keep those that are valid

aholmberg:
+1 Thanks for being flexible. I may need to get some of these design approaches down somewhere.

PR #469, opened 2016-02-04
Add a test for updating UDT with nones

aboudreault:
This is the additional test for the issue mentioned in #467


aholmberg:
tests #467

PR #472, opened 2016-02-08
PYTHON-452 - Fix Cython deserializer date overflow with large timestamp

aboudreault:
* Fix date overflow with large timestamp
* Fix microseconds precision for large timestamp. Due to the "double" data type, there was a difference between the cython ext and the pure python implementation. Now, the code uses int64 and avoid floating values to get the same result than the python impl.

### DIFF
--- a/tests/unit/cython/utils_testhelper.pyx
+++ b/tests/unit/cython/utils_testhelper.pyx
@@ -1,0 +1,23 @@
+# Copyright 2013-2016 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import datetime
+
+from cassandra.cython_utils cimport datetime_from_timestamp
+
+
+def test_datetime_from_timestamp(assert_equal):
+    assert_equal(datetime_from_timestamp(1454781157.123456), datetime.datetime(2016, 2, 6, 17, 52, 37, 123456))
+    # PYTHON-452
+    assert_equal(datetime_from_timestamp(2177403010.123456), datetime.datetime(2038, 12, 31, 10, 10, 10, 123456))
--- a/cassandra/__init__.py
+++ b/cassandra/__init__.py
@@ -23,7 +23,7 @@ class NullHandler(logging.Handler):
 logging.getLogger('cassandra').addHandler(NullHandler())
 
 
-__version_info__ = (3, 1, '0a2')
+__version_info__ = (3, 1, '0a2', 'post0')
 __version__ = '.'.join(map(str, __version_info__))
 
 
--- a/cassandra/__init__.py
+++ b/cassandra/__init__.py
@@ -23,7 +23,7 @@ class NullHandler(logging.Handler):
 logging.getLogger('cassandra').addHandler(NullHandler())
 
 
-__version_info__ = (3, 1, '0a1', 'post0')
+__version_info__ = (3, 1, '0a2')
 __version__ = '.'.join(map(str, __version_info__))
 
 
--- a/cassandra/cython_utils.pyx
+++ b/cassandra/cython_utils.pyx
@@ -35,10 +35,15 @@ from cassandra.util import is_little_endian
 
 import_datetime()
 
+DEF DAY_IN_SECONDS = 86400
+
 DATETIME_EPOC = datetime.datetime(1970, 1, 1)
 
 
 cdef datetime_from_timestamp(double timestamp):
-    cdef int seconds = <int> timestamp
-    cdef int microseconds = (<int64_t> (timestamp * 1000000)) % 1000000
-    return DATETIME_EPOC + timedelta_new(0, seconds, microseconds)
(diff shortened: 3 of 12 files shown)


aboudreault:
If we still want to include this in 3.1, I will replace this PR since the merge-into is master.

aholmberg:
Yes, let's target 3.1. Will merge back to master after we get this in.

cassandra/deserializers.pyx
```diff
@@ -127,8 +127,8 @@ cdef class DesCounterColumnType(DesLongType):
 
 cdef class DesDateType(Deserializer):
     cdef deserialize(self, Buffer *buf, int protocol_version):
-        cdef double timestamp = unpack_num[int64_t](buf) / 1000.0
```
aholmberg: I can see how switching to integer math could change precision, but what was the impetus here? It looks like the pure python version [starts with floating point math](https://github.com/datastax/python-driver/blob/3.0.0/cassandra/cqltypes.py#L558). Can you help me understand why this had to change?

cassandra/deserializers.pyx
```diff
@@ -127,8 +127,8 @@ cdef class DesCounterColumnType(DesLongType):
...
-        return datetime_from_timestamp(timestamp)
+        cdef int64_t timestamp_in_us = unpack_num[int64_t](buf) * 1000
```
aholmberg: Why are we scaling to us? Would it work to keep in ms and scale the us component inside the conversion (allows for higher max value here).

tests/unit/cython/utils_testhelper.pyx
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+import datetime
+
+from cassandra.cython_utils cimport datetime_from_timestamp_in_us
+
+
+def test_datetime_from_timestamp_in_us(assert_equal):
+    assert_equal(datetime_from_timestamp_in_us(1454781157123456), datetime.datetime(2016, 2, 6, 17, 52, 37, 123456))
+    assert_equal(datetime_from_timestamp_in_us(2177403010123456), datetime.datetime(2038, 12, 31, 10, 10, 10, 123456))
```
aholmberg: Would it be possible to add a test that compares the results of deserializers.DesDateType.deserialize and cqltypes.DateType.deserialize over a range of inputs?
I think you've covered a bit of that by adding tests for both time utilities, but for rigor it might be good to explicitly cover the entire conversion from wire format.

cassandra/deserializers.pyx
aboudreault: There is a difference between the Python float data type and the C double.  We have more precision with the Python implementation (for large numbers). 

Python
```
In [215]: 2177403010.123456 * 1000000
Out[215]: 2177403010123456.0
```

C
```
cdef double test = 2177403010.123456
print test * 1000000
>>> 2.17740301012e+15 (217740301012000)
```

That's why I've decided to use int64 over a floating value for our extension. If you want, I can dig more in the Python float versus C double to give you the exact explanation and details about how many bits are for the precision etc.

cassandra/deserializers.pyx
aboudreault: It would work. It also crossed my mind if we shouldn't handle this MS (Cassandra precision) rather than US (Python precision). Since the Python datetime has US support (datetime(year, month, day, hour, minute, second, u)s), I simply decided to go that way.

I really don't have a strong opinion on that. If you prefer I do that change, just confirm here.

tests/unit/cython/utils_testhelper.pyx
aboudreault: Sure, good point.

cassandra/deserializers.pyx
aholmberg: I'm a bit surprised by this since they should both be IEEE-754 doubles. Are you sure it's not just a print formatting issue? I don't want to make a science project out of it... just trying to understand, and make sure we're doing the same thing in both cython and python sides.

cassandra/deserializers.pyx
aholmberg: well, I normally try to avoid scaling overflows if I can. I haven't done the math in this case to see if any overflowing ``us`` count would even be representable in ``datetime``.
Regardless, since we're using signed types I think we should change it to avoid surprising 'wrapped' dates for large counts.

aboudreault: I'll do that change :+1:

cassandra/deserializers.pyx
aboudreault: Your concern make sense Adam.  The initial issue with this is basically a double rounding issue. 

So for a date specified by the user: datetime.datetime(2038, 12, 31, 10, 10, 10, 15021). Our Cython extension returns after deserialization.

Let me dig a little bit more and understand why the python impl doesn't have this issue.

aboudreault: You are right Adam. Both use a double internally and I can reproduce the rounding issue in pure python as well. The reason why we don't have the same behavior with the datetime_from_timestamp function using the python implementation is that we don't cast anything back to integer: the CPython code does it internally with the double passed for the seconds. The CPython code does one additional step than our code, a round() call for the leftover microseconds. I am going to apply this solution. There is one more thing that I need to investigate, which could potentially make the cython and python differ: when the number is exactly halfway between two integers.

https://github.com/python/cpython/blob/master/Modules/_datetimemodule.c#L2154

aholmberg: Nice analysis. Thanks for enlightening us.
Do you think we should try to align the cython version with CPython at this point, or just fix the overflow issue for this ticket, and have a follow-on issue to align (possibly by going to integer math on the Python side)?

aboudreault: +1 to fix the overflow issue and close this ticket. I will create another ticket for the cpython/python alignment.

aboudreault:
Closing in favor of #476

PR #473, opened 2016-02-09
PYTHON-395 - cqle: guard against formatting an invalid keyspace in a query

aholmberg:
Resolves AttributeError when trying to query without a keyspace name,
instead raising a CQLEngineException.

### DIFF
--- a/tests/integration/cqlengine/model/test_model.py
+++ b/tests/integration/cqlengine/model/test_model.py
@@ -16,7 +16,9 @@ try:
 except ImportError:
     import unittest  # noqa
 
-from cassandra.cqlengine import columns
+from mock import patch
+
+from cassandra.cqlengine import columns, CQLEngineException
 from cassandra.cqlengine.management import sync_table, drop_table, create_keyspace_simple, drop_keyspace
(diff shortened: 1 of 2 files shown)


tests/integration/cqlengine/model/test_model.py
```diff
@@ -105,6 +107,25 @@ class table(Model):
 
         drop_keyspace('keyspace')
 
+    def test_column_family(self):
+        class TestModel(Model):
+            k = columns.Integer(primary_key=True)
+
+        from cassandra.cqlengine import models
```
aboudreault: just curious, what is the reason for imports inside the test?

aboudreault:
+1

tests/integration/cqlengine/model/test_model.py
aholmberg: Good question. I think I put it here after looking that ``DEFAULT_KEYSPACE`` access in ``cqlengine.connection``. There the module import is imported inside the functions to avoid circular references with ``cqlengine.model``. 
Here in the test that is not an issue. I'll move it out.

PR #474, opened 2016-02-09
PYTHON-476 - remove cqlengine opmap race

aholmberg:


aboudreault:
+1 on this! Basically, with the metaclass addition, cls.opmap is now a different dict by instance rather than by class, right?

aholmberg:
We now have the opmap aggregated in the base class as the module is imported, rather than lazy-building it lazily as get_operator is called by clients.

aboudreault:
I see. So still in the base class but that way we avoid any possible race condition. Thanks.

PR #476, opened 2016-02-09
PYTHON-452 - Fix Cython deserializer date overflow with large timestamp

aboudreault:
* Fix date overflow with large timestamp
* Add tests for DateType Cython deserializer

### DIFF
--- a/tests/integration/cqlengine/columns/test_validation.py
+++ b/tests/integration/cqlengine/columns/test_validation.py
@@ -70,39 +70,39 @@ class TestDatetime(BaseCassEngTestCase):
                 return None
 
         now = datetime(1982, 1, 1, tzinfo=TZ())
-        dt = self.DatetimeTest.objects.create(test_id=0, created_at=now)
-        dt2 = self.DatetimeTest.objects(test_id=0).first()
+        dt = self.DatetimeTest.objects.create(test_id=1, created_at=now)
+        dt2 = self.DatetimeTest.objects(test_id=1).first()
         assert dt2.created_at.timetuple()[:6] == (now + timedelta(hours=1)).timetuple()[:6]
 
     def test_datetime_date_support(self):
         today = date.today()
-        self.DatetimeTest.objects.create(test_id=0, created_at=today)
-        dt2 = self.DatetimeTest.objects(test_id=0).first()
+        self.DatetimeTest.objects.create(test_id=2, created_at=today)
+        dt2 = self.DatetimeTest.objects(test_id=2).first()
         assert dt2.created_at.isoformat() == datetime(today.year, today.month, today.day).isoformat()
 
     def test_datetime_none(self):
-        dt = self.DatetimeTest.objects.create(test_id=1, created_at=None)
-        dt2 = self.DatetimeTest.objects(test_id=1).first()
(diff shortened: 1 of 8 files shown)


tests/unit/cython/types_testhelper.pyx
```diff
@@ -0,0 +1,65 @@
...
+    cdef Deserializer des = find_deserializer(DateType)
+
+    def deserialize(timestamp):
+        """Serialize a datetime and deserialize it using the cython deserializer"""
+
+        cdef BytesIOReader reader
+        cdef Buffer buf
+
+        dt = datetime.datetime.utcfromtimestamp(timestamp)
+        reader = BytesIOReader(b'\x00\x00\x00\x08' + DateType.serialize(dt, 0))
```
aboudreault: I'm not sure if this is the proper way, but get_buf() requires to have a first int32 value to be non-negative zero.

tests/integration/cqlengine/columns/test_validation.py
```diff
@@ -88,6 +88,23 @@ def test_datetime_none(self):
...
         assert dts[0][0] is None
 
+    def test_datetime_invalid(self):
+        dt_value= 'INVALID'
+        with self.assertRaises(TypeError):
+            self.DatetimeTest.objects.create(test_id=2, created_at=dt_value)
+
+    def test_datetime_timestamp(self):
+        dt_value = 1454520554
+        self.DatetimeTest.objects.create(test_id=2, created_at=dt_value)
```
aholmberg: I don't think this makes a huge difference, but I think this ``test_id`` should be different among tests.

tests/unit/cython/types_testhelper.pyx
```diff
@@ -0,0 +1,65 @@
...
+
+    # beyond 32b
+    expected = 2 ** 33
+    assert_equal(deserialize(expected), datetime.datetime(2242, 3, 16, 12, 56, 32))
+
+    # less than epoc (PYTHON-119)
+    expected = -770172256
+    assert_equal(deserialize(expected), datetime.datetime(1945, 8, 5, 23, 15, 44))
+
+    # work around rounding difference among Python versions (PYTHON-230)
```
aholmberg: Reminder to create a new ticket for going to integer math in date deserializers.

tests/unit/cython/types_testhelper.pyx
aholmberg: This way works. You are essentially prepending the value length for the protocol. You could also create a buffer and ``write_value`` if you felt like it:
https://github.com/datastax/python-driver/blob/3.1.0a1/cassandra/protocol.py#L1214-L1215

aholmberg:
A few minor comments on the diff. Take or leave.
+1

tests/integration/cqlengine/columns/test_validation.py
aboudreault: Yeah. It doesn't make a difference since the row is probably just upserted. But I'll do the change as a good habit. :+1:

tests/unit/cython/types_testhelper.pyx
aboudreault: I prefer that way. Thanks for pointing.

aboudreault:
Addressed comments. Will merge after CI.

PR #477, opened 2016-02-11
Changing set_workload call to match latest changes in ccm

GregBestland:


GregBestland:
I'm merging this to work with the updated version of ccm.

PR #478, opened 2016-02-11
PYTHON-306 - Model Tuple type in cqlengine

aholmberg:


aboudreault:
LGTM! Only tests are missing, you or Greg probably intend to add them.

aholmberg:
Yes. I did some manual testing here. Leaving test impl for Greg

PR #479, opened 2016-02-12
PYTHON-310: Add lazy_connect_lock to avoid race condition with multiple threads

aboudreault:
* Avoid lazy_connect_lock race condition


aholmberg:
+1 on synchronization. I'm waffling on deprecating lazy_connect. Discussing further...

aboudreault:
Remove the deprecation, will be discussed in another ticket

PR #480, opened 2016-02-17
PYTHON-258: Add CONTAINS operator for collections

aboudreault:
I tried to keep this simple and followed what I'm seen in our current code. Perhaps, I didn't forget something else affected by a new operator. If so, let me know.

I indent to add the documentation tomorrow morning.

### DIFF
--- a/tests/integration/cqlengine/query/test_queryset.py
+++ b/tests/integration/cqlengine/query/test_queryset.py
@@ -528,16 +528,16 @@ class TestQuerySetValidation(BaseQuerySetUsage):
         Tests that queries on an indexed field will work without any primary key relations specified
         """
         q = IndexedTestModel.objects(test_result=25)
-        assert q.count() == 4
+        self.assertEqual(q.count(), 4)
 
         q = IndexedTestModel.objects.filter(test_list__contains=42)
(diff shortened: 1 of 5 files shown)


tests/integration/cqlengine/query/test_queryset.py
```diff
@@ -516,6 +530,15 @@ def test_indexed_field_can_be_queried(self):
         q = IndexedTestModel.objects(test_result=25)
         assert q.count() == 4
 
+        q = IndexedTestModel.objects.filter(test_list__contains=42)
```
aholmberg: I know the existing codebase uses bare asserts. We didn't revise all tests when we integrated, but I would prefer not to propagate, and usually try to [fix as I touch a given test](https://github.com/datastax/python-driver/commit/a09c7c41ec9df725234279ed9b0350380b3915c2).

aboudreault: Thanks for letting me know. I wasn't sure if there was a preference there.

aholmberg:
+1 with minor (optional) comment on new tests

aboudreault:
comments have been addressed.

PR #481, opened 2016-02-19
[PYTHON-491] Integration tests for named parameters in queries

kishkaru:
Only including tests for simple statements (and as simple statements in batches), as the driver doesn't support named params in prepared statements.


aholmberg:
These tests are good, but overlap with already existing tests. I'd rather not have more to maintain and run since we're already covered. Thanks for taking a look.

PR #484, opened 2016-02-19
PYTHON-266: Add DISTINCT support

aboudreault:


cassandra/cqlengine/statements.py
```diff
@@ -550,6 +550,7 @@ class SelectStatement(BaseCQLStatement):
     def __init__(self,
                  table,
                  fields=None,
+                 distinct_fields=[],
```
aholmberg: Curious about why this was added in the middle of the args. I think this is probably okay as an internal API, but I usually only append keyword arguments where possible.

aholmberg: Would it be okay to default to None? Looks safe here, but mutable defaults are usually discouraged.

aboudreault: no objection to move. There is not reason really.

aboudreault: +1 for the None.

aholmberg:
+1

aboudreault:
Comments have been addressed.

PR #485, opened 2016-02-22
Fix hang when using eventlet and ssl

sigmunau:
The eventlet read loop runs select before each recv call. This will
fail because recv() on an ssl socket might read more data than
requested from the underlying socket in order to have enough context
for decryption. These extra bytes would be returned by the next recv
on the ssl socket, but are not known to underlying socket so select
will hang.

It is my understanding that eventlet implements blocking IO by using
poll under the hood and yielding, so the select should not be
necessary at all

### DIFF
--- a/cassandra/io/eventletreactor.py
+++ b/cassandra/io/eventletreactor.py
@@ -18,9 +18,9 @@
 
 from errno import EALREADY, EINPROGRESS, EWOULDBLOCK, EINVAL
 import eventlet
-from eventlet.green import select, socket
+from eventlet.green import socket
+import ssl
 from eventlet.queue import Queue
-from functools import partial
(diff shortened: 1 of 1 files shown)


aholmberg:
Thanks. Will take a look at this for 3.1.
https://datastax-oss.atlassian.net/browse/PYTHON-495

sigmunau:
Slightly revised the fix to solve issues with timeouts when socket is idle for some time

aholmberg:
@sigmunau what level of testing have you done with this? I'm finding the tests not working quite right when I run eventlet_nosetests. Are you just testing manually?

aholmberg:
Thanks for the contribution!

PR #343, opened 2015-06-03
Expose conflicting data in case of LWTException

dokai:
Adds a custom attribute, `existing`, to the LWTException instance which
allows the application layer to inspect the values read from Cassandra
which conflicted with the attempted conditional write.

The documentation already referred to the `existing` attribute but it
was not implemented.

PR #488, opened 2016-02-23
PYTHON-336 - Expose LWT results in cqlengine LWTException

aholmberg:
Superset of #343 and targeting 3.1 dev branch.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -70,8 +70,8 @@ from cassandra.pool import (Host, _ReconnectionHandler, _HostReconnectionHandler
                             HostConnectionPool, HostConnection,
                             NoConnectionsAvailable)
 from cassandra.query import (SimpleStatement, PreparedStatement, BoundStatement,
-                             BatchStatement, bind_params, QueryTrace, Statement,
-                             named_tuple_factory, dict_factory, FETCH_SIZE_UNSET)
+                             BatchStatement, bind_params, QueryTrace,
+                             named_tuple_factory, dict_factory, tuple_factory, FETCH_SIZE_UNSET)
 
 
 def _is_eventlet_monkey_patched():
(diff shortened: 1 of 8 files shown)


cassandra/cluster.py
```diff
@@ -3409,3 +3409,24 @@ def get_all_query_traces(self, max_wait_sec_per=None):
...
+    def was_applied(self):
+        """
+        For LWT results, returns whether the transaction was applied.
+
+        Result is indeterminate if called on a result that was not an LWT request.
+
+        Only valid when one of tne of the internal row factories is in use.
+        """
+        if self.response_future.row_factory not in (named_tuple_factory, dict_factory, tuple_factory):
+            raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factsory,))
```
aboudreault: just wondering, should we prefer/force the use of new string formatting (format())?

cassandra/cluster.py
```diff
@@ -3409,3 +3409,24 @@ def get_all_query_traces(self, max_wait_sec_per=None):
...
+        if len(self.current_rows) != 1:
+            raise RuntimeError("LWT result should have exactly one row. This has %d." % (len(self.current_rows)))
+
+        row = self.current_rows[0]
+        if isinstance(row, tuple):
+            return row[0]
+        else:
+            return row['[applied]']
```
aboudreault: Just to be sure, the column name really contains brackets in its name?

aboudreault: looks like so, just seen the old code below

aboudreault:
LGTM!

cassandra/cluster.py
aholmberg: Yes. I have a bit of documentation about it [here] (https://github.com/datastax/python-driver/blob/master/docs/lwt.rst#specialized-results). I'm not aware of any documentation about it in the server.

cassandra/cluster.py
aholmberg: core driver uses all '%', so this is consistent within that layer. cqlengine was using ``format()`` when we brought it in, but it had to be updated because it used unnamed placeholders. I haven't had time or cause to align the two so I've just been keeping it consistent within each package.

PR #490, opened 2016-02-24
PYTHON-258  tests error and negative scenarios

GregBestland:
Modified existing tests to make sure 0 results work correctly, and exceptions are thrown on non-indexed collections.
PYTHON-258

### DIFF
--- a/tests/integration/cqlengine/query/test_queryset.py
+++ b/tests/integration/cqlengine/query/test_queryset.py
@@ -32,6 +32,7 @@ from cassandra.cqlengine.management import sync_table, drop_table
 from cassandra.cqlengine.models import Model
 from cassandra.cqlengine import columns
 from cassandra.cqlengine import query
+from cassandra.cqlengine.query import QueryException
 from datetime import timedelta
 from datetime import tzinfo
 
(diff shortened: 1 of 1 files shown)


tests/integration/cqlengine/query/test_queryset.py
```diff
@@ -266,6 +271,10 @@ def setUpClass(cls):
                                         test_result=45, test_list=[1, 2, 3], test_set=set([1, 2, 3]),
                                         test_map={'1': 1, '2': 2, '3': 42})
 
+        IndexedTestModel.objects.create(test_id=14, attempt_id=4, description='list14', expected_result=75,
```
aholmberg: It's not really making a difference the way the tests are written, but I noticed ``test_id`` is not incremented like the previous entries.

aholmberg:
+1

tests/integration/cqlengine/query/test_queryset.py
aboudreault: We will mostly forgot that many times. I would that whenever we need an id, we use a test class function. ie. test_next_id()

GregBestland:
Fixed Id issue.

PR #491, opened 2016-02-24
PYTHON-337: case sensitive support for column family name

aboudreault:
Only docs is missing

### DIFF
--- a/cassandra/cqlengine/management.py
+++ b/cassandra/cqlengine/management.py
@@ -194,7 +194,7 @@ def sync_table(model):
         if table.indexes.get(index_name):
             continue
 
-        qs = ['CREATE INDEX {0}'.format(index_name)]
+        qs = ['CREATE INDEX {0}'.format(metadata.protect_name(index_name))]
         qs += ['ON {0}'.format(cf_name)]
         qs += ['("{0}")'.format(column.db_field_name)]
         qs = ' '.join(qs)
--- a/cassandra/cqlengine/models.py
+++ b/cassandra/cqlengine/models.py
@@ -15,7 +15,7 @@
 import logging
 import re
 import six
-import warnings
+from warnings import warn
 
 from cassandra.cqlengine import CQLEngineException, ValidationError
 from cassandra.cqlengine import columns
@@ -308,6 +308,8 @@ class BaseModel(object):
 
     __table_name__ = None
 
+    __table_name_case_sensitive__ = False
+
     __keyspace__ = None
 
     __discriminator_value__ = None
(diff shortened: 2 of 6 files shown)


aboudreault:
Added doc.

cassandra/cqlengine/management.py
```diff
@@ -194,7 +194,7 @@ def sync_table(model):
         if table.indexes.get(index_name):
             continue
 
-        qs = ['CREATE INDEX {0}'.format(index_name)]
+        qs = ['CREATE INDEX "{0}"'.format(index_name)]
```
aholmberg: Might want to use ``'CREATE INDEX {0}'.format(protect_name(index_name)`` for the extra escaping it provides.

cassandra/cqlengine/models.py
```diff
@@ -492,7 +494,10 @@ def column_family_name(cls, include_keyspace=True):
     def _raw_column_family_name(cls):
         if not cls._table_name:
             if cls.__table_name__:
-                cls._table_name = cls.__table_name__.lower()
+                if cls.__table_name_case_sensitive__:
+                    cls._table_name = cls.__table_name__
+                else:
+                    cls._table_name = cls.__table_name__.lower()
             else:
```
aholmberg: We should probably handle this for class-derived names too.

aholmberg: Just brainstorming -- what would you think of trying to move away from transformed table names in the future?
Instead of transforming names from camelcase to '\_' delimited, we would always use the names directly. We could transition as follows:
1.) now, introduce a model attribute "transform table name" (or whatever you want to call it), defaulted ``True`` to match current behavior.
2.) Warn if transformed name != literal name. This will encourage people to change their model names to match the actual table name in the database
3.) Next major rev, change default "transform table name" to ``False``, and remove transform logic -- tables will always be named exactly ``cls.__table_name__`` or ``cls.__name__``.

lmkwyt

aboudreault: IMO, I like the camelcase to '_' and lowercase transformation. This is exactly what I want to see by default. Django has the same behavior. Personnaly, I don't want to see camelcase in my db schema. It might be only me.

```
class Automobile(Model) -> automobile
class AutomobileGarage(Model) -> automobile_garage
```  

So, the issue here I think is only that specifying __table_name__ manually doesn't respect the case sensitivity, when it should.

Let's discuss this more tomorrow.

aholmberg: Agreed on explicit table name. I was just going down the path of making model names explicit for simplicity. In your example above, I would encourage users to just name the model ``automobile_garage`` in the first place, instead of having the mapper transform the name.

In any case, I don't feel too strongly about it. We're going to discuss this today internally. I will also ping OP on the ticket to make sure it addresses their issue.

PR #492, opened 2016-02-25
PYTHON-266 : Adding checks for empty, and invalid distinct queries

GregBestland:
Also cleaning up the test suite to not use indexed collections with C*< 2.1


aboudreault:
Thanks Greg. +1

PR #493, opened 2016-02-25
PYTHON-346 - cqle: allow db_field for fields in UDTs

aholmberg:
Previously UDTs with different db_fields would not map properly, resulting in None for all fields
inserted or queried.


aboudreault:
LGTM!

PR #494, opened 2016-02-26
Corrected incorrect reference to self._columns in UserType

mzbyszynski:
This is a very small fix for [PYTHON-502](https://datastax-oss.atlassian.net/browse/PYTHON-502). 

Please let me know if the unit test I added to confirm the fix is not in the correct place.


aholmberg:
thanks!

PR #495, opened 2016-02-26
PYTHON-351 - fix various queries when using db_type on partition key column

aholmberg:
also, a bonus fix for delete queries on models with composite partition keys


aboudreault:
LGTM!

PR #497, opened 2016-02-26
PYTHON-432: Add IF EXISTS support to UPDATE and DELETE statements of CQLEngine

aboudreault:


aboudreault:
Added some documentation.

docs/api/cassandra/cqlengine/models.rst
```diff
@@ -82,7 +82,7 @@ Model
 
     .. automethod:: create
 
-    .. method:: if_not_exists()
+    .. method:: if_not_exists(boolean)
```
aholmberg: I don't think this is right. It should not take an argument.

aboudreault: This issue is related to the fact that these settings use Descriptor.

docs/api/cassandra/cqlengine/models.rst
```diff
@@ -100,6 +100,24 @@ Model
...
+        Check the existence of an object before an update or delete. The existence of an
+        object is determined by its primary key(s). And please note using this flag
+        would incur performance cost.
+
+        If the update or delete isn't applied, a :class:`~cassandra.cqlengine.query.LWTException` is raised.
+
+        .. code-block:: python
+
+            try:
+                TestIfExistsModel.if_exists(id=id).update(count=9, text='111111111111')
```
aholmberg: I don't think if_exists accepts kwargs (?)

aboudreault: this is wrong though, will fix

docs/api/cassandra/cqlengine/models.rst
```diff
@@ -100,6 +100,24 @@ Model
 
         This method is supported on Cassandra 2.0 or later.
 
+    .. method:: if_exists(boolean)
```
aholmberg: Does not take boolean

docs/api/cassandra/cqlengine/models.rst
aholmberg: I see. Sorry for the confusion. I think we should try to get away from that inconsistency. Maybe we can default that parameter ``True`` in the setter and take it away later.

aboudreault:
Addressed comments.

PR #498, opened 2016-02-26
Docs for PYTHON-346, some nit test changes.

GregBestland:


tests/integration/cqlengine/model/test_udts.py
```diff
@@ -447,9 +447,21 @@ class User(UserType):
...
+        @since 3.1.0
+        @jira_ticket PYTHON-346
+        @expected_result The actual cassandra column will use the db_field specified.
+
+        @test_category data_types:udt
+        """
         class db_field_different(UserType):
             age = columns.Integer(db_field='a')
             name = columns.Text(db_field='n')
+            o_id = columns.Integer(db_field='id')
```
aholmberg: Help me out -- what does this add to these tests?

GregBestland: I can remove it if needed at the time I wasn't sure exactly how UDT, and normal fields were being map. The thought here was was that if the field processing was recursive that there might be an opportunity for issues duplicate id's between UDT's and field id's. Depending on how that Id mapping was implemented, much to dismay it worked fine, but I thought it was worth a go.

GregBestland:
Issues taken care rebased master merging.

PR #499, opened 2016-02-29
PYTHON-445 - fix attribute error, and warn when cqlengine batches executed multiple times

aholmberg:


cassandra/cqlengine/query.py
```diff
@@ -197,6 +198,13 @@ def add_callback(self, fn, *args, **kwargs):
         self._callbacks.append((fn, args, kwargs))
 
     def execute(self):
+        if self._executed and self.warn_multiple_exec:
+            msg = "Batch executed multiple times."
+            if self._context_entered:
+                msg += " If using the batch as a context manager, there is no need to call execute directly."
+            warn(msg)
+        self._executed = True
+
```
aboudreault: :+1:

aboudreault:
yep. make sense to me!

PR #502, opened 2016-02-29
cqle: show None values as deleted if the value is explicitly set

aholmberg:
Fixes an issue where a model instance created with explicit None fields
would not persist as such on save()
PYTHON-475

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -37,7 +37,7 @@ class BaseValueManager(object):
 
     @property
     def deleted(self):
-        return self.value is None and self.previous_value is not None
+        return self.value is None and (self.explicit or self.previous_value is not None)
 
     @property
     def changed(self):
@@ -270,8 +270,6 @@ class Blob(Column):
         val = super(Bytes, self).to_database(value)
         return bytearray(val)
 
-    def to_python(self, value):
-        return value
 
 Bytes = Blob
 


aboudreault:
LGTM. Might be good to add a test for this too. //cc Greg

aholmberg:
yes. The ticket description has an example of the problem for the test.

xiaost:
this modification is evil. 
many tombstone cells are created when saving new rows with None fields.
it should be optional or disable the feature when using `create` func

@aholmberg @aboudreault

aholmberg:
This should only affect places where you create something with an explicit None. If you don't want the value overwritten, it should not be set. Are you seeing otherwise?

shuoli84:
Most of time we treat None as do nothing. A function doesn't return == it returns None. Kinda tricky, I guess most developers wouldn't expect a None => A tombstone.

aholmberg:
> Most of time we treat None as do nothing. A function doesn't return == it returns None.

I am aware of the semantics of None in Python. `None` by itself is one thing, but if you explicitly name that attribute in a `kwargs` dict where it could otherwise be **absent**, that is another thing. 

If you explicitly set a value to `None`, it will be deleted in the database. In other words (and now switching to database parlance), `NULL` is distinct from `UNSET`. Attribute `UNSET` implies "do nothing". Setting an explicit `NULL` causes a cell to be deleted (or, written `NULL`). This is the direct way to model a difference between leaving a value untouched, and explicitly setting it to NULL.

Please see https://datastax-oss.atlassian.net/browse/PYTHON-475 for some additional background.

aholmberg:
I created https://datastax-oss.atlassian.net/browse/PYTHON-623 to expound on this in the documentation.

shuoli84:
Still prefer something explicit, e.g an unset operator. field__unset=True, or field=Unset (Some const stands for unset).

PR #503, opened 2016-03-01
PYTHON-373: Add tests for CQLEngine and range deletion

aboudreault:


GregBestland:
+1 looks good to me I'm merging this.

PR #508, opened 2016-03-03
PYTHON-323: Add ability to set query's fetch_size and limit using ORM

aboudreault:
todo: more docs


aholmberg:
+1 changes. We can bring docs in on master

PR #509, opened 2016-03-03
Adding integration test for PYTHON-486

GregBestland:
Fixing some bad test documentation, and adding integration test for PYTHON-486.

### DIFF
--- a/tests/integration/cqlengine/connections/test_connection.py
+++ b/tests/integration/cqlengine/connections/test_connection.py
@@ -1,0 +1,91 @@
+# Copyright 2013-2016 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest  # noqa
+
+
+from cassandra.cqlengine.models import Model
+from cassandra.cqlengine import columns, connection
+from cassandra.cqlengine.management import sync_table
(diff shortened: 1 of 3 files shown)


tests/integration/cqlengine/connections/test_connection.py
```diff
@@ -0,0 +1,88 @@
...
+
+        @test_category object_mapper
+        """
+
+        connection.set_session(self.session1)
+        sync_table(TestConnectModel)
+        TCM1 = TestConnectModel.create(id=1, keyspace=self.keyspace1)
+        connection.set_session(self.session2)
+        sync_table(TestConnectModel)
+        TCM2 = TestConnectModel.create(id=1, keyspace=self.keyspace2)
```
aholmberg: If we're going to use count below, I think we would need to change the id to make sure we're not just overwriting in the same ks/table.
Alternatively, use if-not-exists when creating, or manually inspect cluster metadata to make sure the sync_table targets the expected keyspace.

GregBestland: I could be missing something here, but I think we could catch such a instance in our asserts below. That was the purpose of the separately named keyspace values in the table. If such an over-write occurred the asserts below fail. As the key space value would not match what was expected. You would end up fetching the same keyspace value from the model twice which wouldn't pass muster.

aholmberg: ok. I guess I was just confused.

tests/integration/cqlengine/connections/test_connection.py
```diff
@@ -0,0 +1,88 @@
...
+        cls.setup_cluster.shutdown()
+
+    def setUp(self):
+        self.c = Cluster(protocol_version=PROTOCOL_VERSION)
+        self.session1 = self.c.connect(keyspace=self.keyspace1)
+        self.session1.row_factory = dict_factory
+        self.session2 = self.c.connect(keyspace=self.keyspace2)
+        self.session2.row_factory = dict_factory
+
+    def test_connection_session_switch(self):
```
aholmberg: I think we might leak a few sessions here: session1, session2, and the one that is originally setup in cqlengine.connection. Should we shut those down?

GregBestland: Agreed.

GregBestland:
Issues should now be addressed.

PR #510, opened 2016-03-04
Added default_serial_consistency_level

pistolero:
To make easier developing application working with several DCs we need to have `Cluster.default_serial_consistency_level` in addition to `Cluster.default_consistency_level`.


aholmberg:
Thanks!

PR #511, opened 2016-03-04
Fix EventletConnection without ssl on recent versions of eventlet

sigmunau:
It seems recent versions of eventlet (tested on 0.18.4) has changed
which exception is raised on timeout during recv(). This was not
handled properly by PR #485 which caused a regression.


aboudreault:
@aholmberg I let you double check this PR since you've already played with that stuff in 485. Thanks

aholmberg:
+1 This aligns with a [similar fix](https://github.com/datastax/python-driver/commit/f80c1ae714ce9de111e4b5d321dfad78635d1245) in geventreactor. Thanks for fixing it here.

PR #512, opened 2016-03-04
Simple test for Python-503, slight doc tweak

GregBestland:
In addition to this simple test I also ran a regression with DesBytesTypeByteArray set as the default deserializer.It all passed. Given the usecase this should be sufficient.

### DIFF
--- a/tests/integration/standard/test_types.py
+++ b/tests/integration/standard/test_types.py
@@ -21,6 +21,7 @@ from datetime import datetime
 import math
 import six
 
+import cassandra
 from cassandra import InvalidRequest
 from cassandra.cluster import Cluster
 from cassandra.concurrent import execute_concurrent_with_args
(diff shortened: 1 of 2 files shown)


tests/integration/standard/test_types.py
```diff
@@ -98,6 +99,36 @@ def test_can_insert_blob_type_as_bytearray(self):
...
+
+        @since 3.1
+        @jira_ticket PYTHON-503
+        @expected_result byte array should be deserialized appropriately.
+
+        @test_category queries:custom_payload
+        """
+        original = None
+        try:
+            if hasattr(cassandra.deserializers, 'DesBytesTypeByteArray'):
```
aholmberg: I don't think this check is needed

GregBestland:
Unneeded check removed merging into master

PR #516, opened 2016-03-08
UUID default value

tigrus:
Currently example from python driver homepage not works. This is fix for this case: 
http://stackoverflow.com/questions/35870127/cassandra-default-column-value


tigrus:
I had bad understanding of default values. Sorry.

PR #517, opened 2016-03-09
BugFix: It is required to setup() cqlengine before executing queries.

fatelei:
# Description
In the concurrent environment, if the global `lazy_connect_args` is accessed by different threads or coroutine, it will cause exception: It is required to setup() cqlengine before executing queries.

# Fix
set `lazy_connect_args` to be `threading.local` object.

### DIFF
--- a/cassandra/cqlengine/connection.py
+++ b/cassandra/cqlengine/connection.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 from collections import namedtuple, defaultdict
+from threading import local
 import logging
 import six
-import threading
 
 from cassandra.cluster import Cluster, _NOT_SET, NoHostAvailable, UserTypeDoesNotExist
 from cassandra.query import SimpleStatement, Statement, dict_factory
@@ -32,8 +32,8 @@ Host = namedtuple('Host', ['name', 'port'])
 
 cluster = None
 session = None
-lazy_connect_args = None
-lazy_connect_lock = threading.RLock()
+lazy_connect_args = local()
+lazy_connect_args.options = None
 
 
 # Because type models may be registered before a connection is present,
@@ -121,7 +121,7 @@ def setup(
         kwargs['consistency'] = consistency
         kwargs['lazy_connect'] = False
         kwargs['retry_connect'] = retry_connect
-        lazy_connect_args = (hosts, kwargs)
+        lazy_connect_args.options = (hosts, kwargs)
         return
 
     cluster = Cluster(hosts, **kwargs)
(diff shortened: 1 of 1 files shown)


aboudreault:
Thank you @fatelei for the PR. This seems to be a duplicated of the following ticket: https://datastax-oss.atlassian.net/browse/PYTHON-310 , which has also a solution.

What is the exception/failure you are experiencing in your environment? Do you have a test case so we could easily reproduce the issue?

fatelei:
There is a chance same coroutine access the RLock，as the official document says:
> threading.RLock()
A factory function that returns a new reentrant lock object. A reentrant lock must be released by the thread that acquired it. Once a thread has acquired a reentrant lock, the same thread may acquire it again without blocking; the thread must release it once for each time it has acquired it.

And this is test server code.

```
# -*- coding: utf8 -*-

import os
import gevent
from threading import RLock
from flask import Flask

app = Flask(__name__)
rlock = RLock()
i = 0

def test():
    gevent.sleep(10)

@app.route("/", methods=["GET"])
def main():
    global i
    with rlock:
        if not i:
            i = 1
            print "here %d" % i
            test()
    print os.getpid()
    return ""

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=5000, debug=True)
```

This is my client code.

```
# -*- coding: utf8 -*-

import gevent.monkey
gevent.monkey.patch_socket()

import gevent
import urllib2


def fetch(pid):
    response = urllib2.urlopen('http://192.168.99.101:5000')
    result = response.read()

    print('Process %s:%s' % (pid, result))


def asynchronous():
    threads = []
    for i in range(1, 10):
        threads.append(gevent.spawn(fetch, i))
    gevent.joinall(threads)


asynchronous()
```

The server output:

image

You can see `here 1` is outputed twice, though issue does not always occur. @aboudreault

fatelei:
I use the latest cassandra-driver，but the exception is still here.

image

aboudreault:
Thanks for the test case @fatelei. I will take a look at this asap.

aboudreault:
@fatelei , When you say that you use the latest cassandra-driver, do you mean you are using the latest release (3.0.0) or you are using the latest master in development? This issue is effectively present in 3.0.0 but the fix will be available in the 3.1.0 release.

Also, the reason why you see  "here 1" twice is because each worker that gunicorn starts has its own application. All workers has their own flask app, RLock and thereby their own connection to the cluster.

fatelei:
@aboudreault I use version 3.1.0a2 in my production environment.

fatelei:
I will show more about the test cases.

aboudreault:
Can you give a try to 3.1.0?

fatelei:
ok

fatelei:
Here，I use `3.1.0` the update operation creates a new record instead of updating exists record.

aboudreault:
Hello @fatelei. Did you get a chance to look into this?

fatelei:
Trying

fatelei:
Hi, @aboudreault I have tried the version 3.1.0 in my production environment, the bug has been fixed, and I will close this pull request. Thanks for your advice.

PR #518, opened 2016-03-11
Implement a BlackList policy for the Python driver.

tferrandiz:
The policy allows to blacklist a list of DCs or a list of hosts
to prevent the driver from connecting to them.
It is a chainable policy similar to the Token-aware policy.

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -309,6 +309,47 @@ class DCAwareRoundRobinPolicy(LoadBalancingPolicy):
     def on_remove(self, host):
         self.on_down(host)
 
+class BlackListPolicy(LoadBalancingPolicy):
+    """
+    A :class:`.LoadBalancingPolicy` wrapper that adds the possibility to blacklist
+    a list of datacenters and/or a list of hosts to a child policy.
+    """
+
+    _child_policy = None
+    _blacklisted_dcs = None
+    _blacklisted_hosts = None
+
+    def __init__(self, child_policy, blacklisted_dcs=[], blacklisted_hosts=[]):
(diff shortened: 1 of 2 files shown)


aholmberg:
Thanks for the contribution. We'll be taking a look at this with the next core release.

mambocab:
Hi @tferrandiz -- we're having some internal discussion about whether and how to integrate a BlackList LBP. What's the use case that prompted you to write this?

tferrandiz:
Hi,
The use case is the following:
We were using a Cassandra ring spread into 3 data centers. 2 of them were used for live traffic and the 3rd to run off-line analytics jobs. So the purpose of the blacklist is to allow the admins to blacklist the analytics DC to make sure it does not directly receive traffic from the application.
Using a blacklist instead of a whitelist allows the admins to add nodes in the 2 main DC or even add a new DC without having to add them manually in the whitelist.

Hope this helps.
Thomas

mambocab:
We've decided not to support a Blacklist LBP. We've found that it can look useful to some users, but is often actually harmful. Thanks for the PR, though, especially for the tests.

mambocab:
For a little more information -- we are discussing implementing a policy that filters hosts based on a user-provided predicate function, like [the Java driver's `HostFilterPolicy`](http://docs.datastax.com/en/drivers/java/3.2/com/datastax/driver/core/policies/HostFilterPolicy.htmlhttp://docs.datastax.com/en/drivers/java/3.2/com/datastax/driver/core/policies/HostFilterPolicy.html). Such a policy makes it easier for users to implement their own blacklist policy.

PR #520, opened 2016-03-17
PYTHON-249 - Fixing conditional deletes in cqlengine

aholmberg:


aboudreault:
LGTM! +1  for the naming refactoring, make sense.

PR #522, opened 2016-03-22
PYTHON-260 - cqle: warn when column or field types are different from the model

aholmberg:


aboudreault:
:+1:

aboudreault:
I will let Greg decides if we add a test for this or not.

PR #526, opened 2016-03-23
PYTHON-273 - option to make models with datetime comparable

aholmberg:


aboudreault:
Assuming that someone will add a test for this, LGTM!

PR #527, opened 2016-03-24
PYTHON-520 - don't select values for constrained columns

aholmberg:


aboudreault:
+1. I like how this is implemented.

PR #528, opened 2016-03-25
PYTHON-260 added test for additional logging

GregBestland:
Maybe a bit overboard here, but I added some tests for both new logging messages.


aholmberg:
+1
Just for reference, there is also [this technique](https://github.com/datastax/python-driver/blob/96d38eadfeb97f0bc3fd66ac378f66f578eed3ab/tests/integration/cqlengine/test_batch_query.py#L220-L226) for verifying warning messages.

PR #529, opened 2016-03-27
Correctly encode UDT types

emadwill:
User-defined types are not currently encoded correctly.
If there's a UDT type "Point" with fields "x" and "y", it is currently encoded as:
Point(x=1.23, y=12.2)
However, in CQL, it should be encoded as:
{x: 1.23, y=12.2}

### DIFF
--- a/cassandra/encoder.py
+++ b/cassandra/encoder.py
@@ -138,8 +138,11 @@ class Encoder(object):
     def cql_encode_object(self, val):
         """
         Default encoder for all objects that do not have a specific encoder function
-        registered. This function simply calls :meth:`str()` on the object.
+        registered. If this object has _asdict() defined (e.g. namedtuple), it will be called first.
+        Otherwise, this function simply calls :meth:`str()` on the object.
         """
+        if hasattr(val, '_asdict'):
+            return self.cql_encode_map_collection(val._asdict())
         return str(val)
 
     def cql_encode_float(self, val):


aholmberg:
I'm not sure we will be able to accept this change in behavior. Can you describe what you're trying to do a bit more? Are you [registering the user type](http://datastax.github.io/python-driver/user_defined_types.html#registering-a-class-to-map-to-a-udt)?

emadwill:
Motivation: I'm trying to encode the results of a query into CQL format. More specifically, a simple program similar to mysqldump that prints "INSERT INTO ... VALUES (...)" for every row in a query.

Normal types (OrderedDict, list, datetime, int, etc.) are encoded correctly, but UDTs are encoded in the standard python namedtuple way not the standard CQL format. e.g. "Point(x=1.23, y=12.2)" instead of "{x: 1.23, y=12.2}".

emadwill:
I did, I managed to encode the UDTs in the CQL format after I defined an __str__ method in the class that I register with the UDT and did the encoding manually.

Shouldn't the default behavior for the Encoder.cql_encode_all_types() function to output the CQL format for a UDT though?

aholmberg:
> Shouldn't the default behavior for the Encoder.cql_encode_all_types() function to output the CQL format for a UDT though?
That's exactly what registering does -- it creates an encoder function that writes objects of the registered type in CQL UDT format.
http://datastax.github.io/python-driver/user_defined_types.html#registering-a-class-to-map-to-a-udt
https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L1932-L1938
Can you provide a gist that shows what you are trying to do?

emadwill:
Here's what I get when I register the type and try to encode it:
https://gist.github.com/emadwill/6a8ea5cc804e4c22753b8c289c16cd65

aholmberg:
Thanks. Will continue discussion of this code in the gist.

PR #531, opened 2016-03-30
PYTHON-530 - fix db_field for queryset update

aholmberg:
also adds some refactoring to simplify and cut down maintenance or issues like this


aboudreault:
+1. Nice refactoring!

PR #534, opened 2016-03-31
PYTHON-532 - validate PK composition

aholmberg:


cassandra/cqlengine/management.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections import namedtuple
+from itertools import chain
```
aboudreault: doesn't seem to be used

aboudreault:
LGTM!

cassandra/cqlengine/management.py
aholmberg: oops. That's from an earlier iteration. Thanks!

PR #536, opened 2016-04-03
Fix for #Python-533. Secondary index on set field

tigrus:
Sync table fails on complex fields with index=True. It tries to create duplicate index. 

This happens because for complex field - index target look like "values(<column_name>)" instead of "<column_name>".  

Jira Issue: https://datastax-oss.atlassian.net/browse/PYTHON-533?jql=text%20~%20%22secondary%22


aholmberg:
Thanks for the contribution!

PR #539, opened 2016-04-04
PYTHON-528 - cqle: support more comparisons in LWT conditional

aholmberg:
also fixes an issue where iff did not use the column's db_field

PYTHON-528


aboudreault:
+1

PR #541, opened 2016-04-06
PYTHON-444 - richer error for ResponseFuture timeout callback

aholmberg:
PYTHON-444


aboudreault:
yep, I like this change.

PR #544, opened 2016-04-08
PYTHON-327 - Cluster config to disable schema and token processing

aholmberg:


aboudreault:
+1 , even if we haven't reproduced the user issue, I like that the driver is more configurable on that side.

PR #549, opened 2016-04-11
PYTHNON-525. Adding decorator for counting underlying query execution

GregBestland:
Also Updating some tests to add expected query counts.

There are more places we can annotate, such as table syncs and UDTS, but this is a decent enough start, and would have caught the error that spawned this issue.

### DIFF
--- a/tests/integration/cqlengine/__init__.py
+++ b/tests/integration/cqlengine/__init__.py
@@ -90,9 +90,13 @@ def execute_count(expected):
             count.get_counter()
             # DeMonkey Patch our code
             cassandra.cqlengine.connection.execute = original_function
+            # Check to see if we have a pre-existing test case to work from.
+            if len(args) is 0:
+                test_case = unittest.TestCase("__init__")
+            else:
+                test_case = args[0]
             # Check to see if the count is what you expect
-            tc = unittest.TestCase("__init__")
-            tc.assertEquals(count.get_counter(), expected, "Expected number of cassandra.cqlengine.connection.execute calls doesn't match actual number invoked Expected: {0}, Invoked {1}".format(count.get_counter(), expected))
+            test_case.assertEqual(count.get_counter(), expected, msg="Expected number of cassandra.cqlengine.connection.execute calls doesn't match actual number invoked Expected: {0}, Invoked {1}".format(count.get_counter(), expected))
             return to_return
         # Name of the wrapped function must match the original or unittest will error out.
         wrapped_function.__name__ = fn.__name__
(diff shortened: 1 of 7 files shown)


tests/integration/cqlengine/__init__.py
```diff
@@ -44,3 +52,57 @@ def setup_connection(keyspace_name):
...
+            cassandra.cqlengine.connection.execute = count.wrapped_execute
+            # Invoked the underlying unit test
+            to_return = fn(*args, **kwargs)
+            # Get the count from our monkey patched counter
+            count.get_counter()
+            # DeMonkey Patch our code
+            cassandra.cqlengine.connection.execute = original_function
+            # Check to see if the count is what you expect
+            tc = unittest.TestCase("__init__")
+            tc.assertEquals(count.get_counter(), expected, "Expected number of cassandra.cqlengine.connection.execute calls doesn't match actual number invoked Expected: {0}, Invoked {1}".format(count.get_counter(), expected))
```
aholmberg: assertEqual instead of assertEquals (pending deprecation)

tests/integration/cqlengine/__init__.py
```diff
@@ -44,3 +52,57 @@ def setup_connection(keyspace_name):
...
+            # Monkey patch in our StatementCounter wrapper
+            cassandra.cqlengine.connection.execute = count.wrapped_execute
+            # Invoked the underlying unit test
+            to_return = fn(*args, **kwargs)
+            # Get the count from our monkey patched counter
+            count.get_counter()
+            # DeMonkey Patch our code
+            cassandra.cqlengine.connection.execute = original_function
+            # Check to see if the count is what you expect
+            tc = unittest.TestCase("__init__")
```
aholmberg: This creates a new test instance that doesn't seem to fail the one we're wrapping. Could we pass in a reference to the existing test and use that to assert?

tests/integration/cqlengine/__init__.py
aholmberg: When I run locally I get loads of these errors, but the test doesn't fail.
`  tc.assertEquals(count.get_counter(), expected, "Expected number of cassandra.cqlengine.connection.execute calls doesn't match actual number invoked Expected: {0}, Invoked {1}".format(count.get_counter(), expected))
ok`

tests/integration/cqlengine/__init__.py
GregBestland: This logic was originally outside of the context of the wrapper function, so the Test Case wasn't there. Now it should be a simple matter to pull the Test case off the args list and use that. Good catch thanks.

PR #550, opened 2016-04-12
PYTHON-286 - Change default conviction behavior for heartbeat timeouts

aholmberg:


tests/unit/test_connection.py
```diff
@@ -382,8 +384,8 @@ def send_msg(msg, req_id, msg_callback):
         connection.send_msg.assert_has_calls([call(ANY, request_id, ANY)] * get_holders.call_count)
         connection.defunct.assert_has_calls([call(ANY)] * get_holders.call_count)
         exc = connection.defunct.call_args_list[0][0][0]
-        self.assertIsInstance(exc, Exception)
-        self.assertEqual(exc.args, Exception('Connection heartbeat failure').args)
+        self.assertIsInstance(exc, ConnectionException)
+        self.assertRegex(exc.args[0], r'^Received unexpected response to OptionsMessage.*')
```
aboudreault: right, assertRegex is not available in py2.7

aboudreault:
So, with the default max_connections_per_host set to 2, a single failure (1 conn) will keep the host in a reconnecting mode. But if we get 2 quick failures with both connections, it will be set as down as before. That's it?

tests/unit/test_connection.py
aholmberg: argh! I was thrown off by having unittest2 installed in my environment. Revisedin 6684e1a

aholmberg:
> But if we get 2 quick failures with both connections, it will be set as down as before. That's it?

Not quite. Basically we're saying heartbeat timeouts alone aren't sufficient to mark a host *down*. We'll close the timed-out connection, but leave the connection pool in-place and keep trying to replace the connection unless we get a down event some other way.
Before, a timeout would immediately "convict" a node, cause the pool to be closed, and a reconnect schedule started.

PR #558, opened 2016-04-15
543

GregBestland:


tests/integration/__init__.py
```diff
@@ -86,6 +86,15 @@ def _tuple_version(version_string):
 
 USE_CASS_EXTERNAL = bool(os.getenv('USE_CASS_EXTERNAL', False))
 
+# If set to to true this will force the Cython tests to run regardless of whether they are installed
+cython_env = os.getenv('VERIFY_CYTHON', "False")
```
aholmberg: can we make this variable have boolean semantics like the others in the project

aholmberg:
It looks like you rebased some changes from 542. Those should not be needed since they are already in master. Should go away when this branch gets rebased again.

tests/integration/__init__.py
GregBestland: This was done intentionally to make things a little easier to integrate on the Jenkins side. It's easy to map the existing cython flag right in, It's a little more complicated to try to add or remove ithe based on the environmental config in the job builder. If possible I'd like to keep it the way it is, unless you view that as intolerable.

PR #560, opened 2016-04-19
PYTHON-296 Add tests for hostname validation

GregBestland:


tests/integration/long/test_ssl.py
```diff
@@ -193,6 +242,62 @@ def test_can_connect_with_ssl_client_auth(self):
...
+        basic operations, with match_hostname specified
+        @jira_ticket PYTHON-296
+        @since 3.3
+
+        @expected_result The client can connect via SSL and preform some basic operations
+
+        @test_category connection:ssl
+        """
+
+        # Need to get absolute paths for certs/key
```
aholmberg: These tests seem almost identical aside from additional ssl options. Is it worth refactoring?

tests/integration/long/test_ssl.py
```diff
@@ -193,6 +242,62 @@ def test_can_connect_with_ssl_client_auth(self):
...
+        abs_driver_certfile = os.path.abspath(DRIVER_CERTFILE)
+
+        tries = 0
+        while True:
+            if tries > 5:
+                raise RuntimeError("Failed to connect to SSL cluster after 5 attempts")
+            try:
+                cluster = Cluster(protocol_version=PROTOCOL_VERSION, ssl_options={'ca_certs': abs_path_ca_cert_path,
+                                                                                  'ssl_version': ssl.PROTOCOL_TLSv1,
+                                                                                  'keyfile': abs_driver_keyfile,
```
aholmberg: I'm not sure this tests any more than the one above. Because the server is not doing mutual authentication, I think the only thing in play here is the ca cert (same as above).

aholmberg:
I think we should have a test that verifies the driver rejecting a cert that does not match. I think we can easily do this by connecting with contact point 127.0.0.2 or 3. The cert in this test is for 127.0.0.1, so those other contact points should fail.

GregBestland:
Good points adam I'll see what I can do about the refactor and the negative test.

GregBestland:
I think these changes should address all the issues raised. Thanks for the feedback

PR #562, opened 2016-04-20
PYTHON-548 - prevent none in token host mapping

aholmberg:


sethrosenblum:
Looks right to me, thanks for the quick turnaround!

aboudreault:
+1

PR #568, opened 2016-04-27
shows that the numpy parser shouldn't reject empty strings

sontek:


sontek:
The test isn't great because I didn't know how you like to structure things but it does fail on master and passes with this change.  

This is to show the fix for https://datastax-oss.atlassian.net/browse/PYTHON-550

sontek:
#569 is a better implementation I believe

PR #569, opened 2016-04-27
Support empty string values in NumpyProtocolHandler

markflorisson:
This PR addresses a bug in NumpyProtocolHandler to properly handle empty strings:

    https://datastax-oss.atlassian.net/browse/PYTHON-550

It doesn't support NULL string values however, which requires more invasive changes (e.g. making use of NumPy's missing values support).

### DIFF
--- a/cassandra/deserializers.pyx
+++ b/cassandra/deserializers.pyx
@@ -44,6 +44,8 @@ cdef class Deserializer:
 
 cdef class DesBytesType(Deserializer):
     cdef deserialize(self, Buffer *buf, int protocol_version):
+        if buf.size == 0:
+            return ""
         return to_bytes(buf)
 
 # this is to facilitate cqlsh integration, which requires bytearrays for BytesType
@@ -51,6 +53,8 @@ cdef class DesBytesType(Deserializer):
 # deserializers.DesBytesType = deserializers.DesBytesTypeByteArray
 cdef class DesBytesTypeByteArray(Deserializer):
     cdef deserialize(self, Buffer *buf, int protocol_version):
+        if buf.size == 0:
+            return bytearray()
         return bytearray(buf.ptr[:buf.size])
 
 # TODO: Use libmpdec: http://www.bytereef.org/mpdecimal/index.html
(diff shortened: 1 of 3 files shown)


sontek:
travis isn't running the integration tests but I've verified manually that this does in fact show (and fix) the problem.

aholmberg:
Thanks for this @markflorisson. It's unclear to me why the conditionals are added in ``cassandra.deserializers``.  Zero seems to work naturally with the slicing. Is this an optimization?

markflorisson:
Hi @aholmberg, I haven't tried it without the conditionals, but the 'ptr' field in the buffer will be NULL, which may be a problem.

aholmberg:
Thanks. I did try it, and it was working. I guess it's because no memory is accessed creating bytes with size zero.

aholmberg:
Thanks for the contribution! Any recommendations on handling null would be welcome. Presently starting here: http://docs.scipy.org/doc/numpy-1.10.1/neps/missing-data.html#definition-of-missing-data

aholmberg:
It looks like this changes the behavior when receiving null values (used to translate to {{None}} for objects, now raises). I will look into it, but I may have to roll this back for this release.

aholmberg:
fbe53a5

markflorisson:
I'm think NumpyProtocolHandler was already rejecting NULL values. I've pushed a new commit that uses None for NULL object values, but rejects NULL values for native/numeric data types.

Supporting NULL for numeric data types is harder, since missing values are not exposed in the python buffer interface. Some other mechanism is needed to efficiently load data with missing values, perhaps based on this:

    http://docs.scipy.org/doc/numpy-1.10.1/neps/missing-data.html#mask-binary-format

It should be possible to access and manuallly update the mask array from Cython.

markflorisson:
Ah you've also pushed an update, should I still do a new PR with the changes?

aholmberg:
My change just restored the ability to read {{None}} for objects. Do you have more than that?

markflorisson:
It's very similar, if you want to you can port the changes/cherry-pick the commit as well, or just keep your commit:

    https://github.com/markflorisson/python-driver/commit/3968bfaa1ec124600ee9ff1c9d8f3bcde811a637

aholmberg:
Yours is better. Thanks.

PR #572, opened 2016-05-04
PYTHON-220 - validate request-per-connection thresholds

aholmberg:


cassandra/cluster.py
```diff
@@ -765,6 +765,10 @@ def set_min_requests_per_connection(self, host_distance, min_requests):
             raise UnsupportedOperation(
                 "Cluster.set_min_requests_per_connection() only has an effect "
                 "when using protocol_version 1 or 2.")
+        if min_requests < 0 or min_requests > 126 or \
+           min_requests >= self._max_requests_per_connection[host_distance]:
+            raise ValueError("min_requests must be in the range [0, 127) and less than the max_requests for this host_distance (%d)" %
```
aboudreault: small typo in the range representation --> [0, 127)

cassandra/cluster.py
```diff
@@ -782,6 +786,10 @@ def set_max_requests_per_connection(self, host_distance, max_requests):
             raise UnsupportedOperation(
                 "Cluster.set_max_requests_per_connection() only has an effect "
                 "when using protocol_version 1 or 2.")
+        if max_requests < 1 or max_requests > 127 or \
+           max_requests <= self._min_requests_per_connection[host_distance]:
+            raise ValueError("max_requests must be in the range (0, 127] and greater than the min_requests for this host_distance (%d)" %
```
aboudreault: same comment than above

aboudreault:
minor comments, +1

cassandra/cluster.py
aholmberg: That's interval notation: https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints
If you think it's too confusing I could change it. Realistically these parameters should never be near the endpoints though.

aholmberg: (merging for now, but will resolve with a change after we finish discussing)

PR #574, opened 2016-05-04
PYTHON-434 - py2.6 compatibility for benchmark scripts

aholmberg:


cassandra/io/asyncorereactor.py
```diff
@@ -102,7 +102,7 @@ def handle_read(self):
     def notify_loop(self):
         if not self._notified:
             self._notified = True
-            os.write(self.write_fd, 'x')
+            os.write(self.write_fd, b'x')
```
aboudreault: I suppose this is more a py3 fix rather than  a 2.6 one, right? If not, why is this required?

aboudreault:
+1

cassandra/io/asyncorereactor.py
aholmberg: You are correct. I noticed the issue as I was testing these changes across all versions.

PR #579, opened 2016-05-09
PYTHON-219 - Address confusion around constructing SimpleStatements

aholmberg:


aboudreault:
LGTM. I am wondering if we should switch retry_policy and consistency_level parameters order in the next major release...

PR #581, opened 2016-05-11
Updates to queryset docs

WilliamQLiu:
* Cannot filter by primary_key only (in example, created 'id' and set as a 'partition_key')
* Updated import from 'cqlengine' to 'columns' and 'functions'


aholmberg:
Thanks for the contribution!

PR #584, opened 2016-05-12
PYTHON-559 - normalize BytesToken values to binary type

aholmberg:
fixes UnicodeDecodeError when one value has non-ascii values

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -1519,10 +1519,8 @@ class BytesToken(Token):
 
     def __init__(self, token_string):
         """ `token_string` should be string representing the token. """
-        if not isinstance(token_string, six.string_types):
-            raise TypeError(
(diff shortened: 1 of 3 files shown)


cassandra/metadata.py
```diff
@@ -1449,6 +1453,13 @@ def hash_fn(cls, key):
     def from_key(cls, key):
         return cls(cls.hash_fn(key))
 
+    @classmethod
+    def from_string(cls, token_string):
+        raise NotImplementedError()
+        """ `token_string` should be the string representation from the server. """
+        # The hash partitioners just store the deciman value
+        return cls(int(token_string))
+
```
aboudreault: raise before return?

aboudreault:
+1

cassandra/metadata.py
aholmberg: Good catch. Leftover lines after a refactor. 7f7f798

PR #585, opened 2016-05-16
Don't use a kwarg to pass a callback to weakref.ref

wulczer:
Turns out that until recently, weakref.ref() silently ignored its kwargs, but
a recent Python patch causes it to throw an exception instead. This makes cqlsh
error out immediately when using it with the latest Python version.

In order to correctly register the weakref callback, pass it as a regular
argument instead of using a kwarg.

For more information, see this Python issue and related Python 2.7 commit:

    https://bugs.python.org/issue17765
    https://hg.python.org/cpython/rev/60de9c6188cc

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2260,7 +2260,7 @@ class ControlConnection(object):
         # _clear_watcher will be called when this ControlConnection is about to be finalized
         # _watch_callback will get the actual callback from the Connection and relay it to
         # this object (after a dereferencing a weakref)
-        self_weakref = weakref.ref(self, callback=partial(_clear_watcher, weakref.proxy(connection)))
+        self_weakref = weakref.ref(self, partial(_clear_watcher, weakref.proxy(connection)))
         try:
             connection.register_watchers({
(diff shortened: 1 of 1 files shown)


aholmberg:
Thanks for the heads up. Just to make sure I understand the impact -- the runtime change is not in any released versions, correct?

Will merge when CLA is ticked.

wulczer:
Hey, yes, the latest released Python 2.7 doesn't contain the change yet. I only saw this because Debian uses the lastest Python master.

I'm working on getting the CLA signed.

wulczer:
Just a heads-up: I'm jumping through hoops to get the CLA signed :/ Please be patient.

aholmberg:
Thanks. Unfortunate that this kind of stuff is required.

aholmberg:
@wulczer any prospects for getting this signed? Otherwise I can just commit a change under my own authorship with thanks to you.

wulczer:
@aholmberg I say please go ahead and commit under your authorship. The wheels of bureaucracy turn slowly but let's get bugs fixed anyway :)

aholmberg:
Commit mentioned above. Thanks again.

PR #586, opened 2016-05-16
Make LBP tests more robust, and fix regression related to PYTHON-531

aholmberg:


GregBestland:
This runs much better. There is a small decommission issue to work out here, but I'm going to merge this in, and work that separately.

PR #587, opened 2016-05-19
PYTHON-565

aboudreault:
Resolve hosts in WhiteListRoundRobinPolicy


GregBestland:
You can wait for Adam if you want but +1 from me on this, and on the test as well.

aboudreault:
Let's merge this. Adam has already seen the patch.

aholmberg:
I was +1 in a different channel, just waiting for that test. lgtm. Thanks for the quick turnaround.

PR #588, opened 2016-05-25
Fixed tests for some legacy thrift tables

stef1927:
As discussed on CASSANDRA-11055, this is one more fix for the 2.7.2-cassandra-2.1 branch.

### DIFF
--- a/tests/integration/standard/test_metadata.py
+++ b/tests/integration/standard/test_metadata.py
@@ -975,14 +975,12 @@ Approximate structure, for reference:
 
 CREATE TABLE legacy.composite_comp_with_col (
     key blob,
-    t timeuuid,
-    b blob,
-    s text,
+    column1 'org.apache.cassandra.db.marshal.DynamicCompositeType(org.apache.cassandra.db.marshal.BytesType, org.apache.cassandra.db.marshal.UTF8Type, org.apache.cassandra.db.marshal.TimeUUIDType)',
     "b@6869746d65776974686d75736963" blob,
     "b@6d616d6d616a616d6d61" blob,
-    PRIMARY KEY (key, t, b, s)
+    PRIMARY KEY (key, column1)
 ) WITH COMPACT STORAGE
-    AND CLUSTERING ORDER BY (t ASC, b ASC, s ASC)
+    AND CLUSTERING ORDER BY (column1 ASC)
     AND caching = '{"keys":"ALL", "rows_per_partition":"NONE"}'
     AND comment = 'Stores file meta data'
     AND compaction = {'min_threshold': '4', 'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32'}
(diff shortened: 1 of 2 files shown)


tests/integration/standard/test_metadata.py
```diff
@@ -1103,12 +1103,13 @@ def test_legacy_tables(self):
 
 CREATE TABLE legacy.composite_comp_no_col (
     key blob,
-    column1 'org.apache.cassandra.db.marshal.DynamicCompositeType(org.apache.cassandra.db.marshal.TimeUUIDType, org.apache.cassandra.db.marshal.BytesType, org.apache.cassandra.db.marshal.UTF8Type)',
```
aholmberg: I don't want to beat this to death since this structure is not expressible in CQL in this version, but aren't we losing something by not showing DCT is involved?
If this seems fine to the C* stakeholders, I'm fine with it too.

stef1927: I don't know about the C* stakeholders since I am not familiar with how Thrift was or is still used, we'd have to ask @iamaleksey, but I do know that a dynamic composite type is implemented in CQL with clustering columns, read the "composites" section of [this blog](http://www.datastax.com/dev/blog/thrift-to-cql3).

aholmberg: I am aware of the analog between "static" composites and clustering. However, I don't think DynamicCompositeType has a native CQL equivalent. I thought the only way to get the DCT behavior was to use the internal class/alias literal.

iamaleksey:
DynamicCompositeType is *not* implemented with clustering columns. That said, the current representation is incorrect, and is missing all the aliases.

iamaleksey:
Should be:

column1 'org.apache.cassandra.db.marshal.DynamicCompositeType(t => org.apache.cassandra.db.marshal.TimeUUIDType, b => org.apache.cassandra.db.marshal.BytesType, s => org.apache.cassandra.db.marshal.UTF8Type)'

PRIMARY KEY (key, column1)

s, b, and t must *not* be in the list of columns, because they aren't.

"b@6869746d65776974686d75736963" and "b@6d616d6d616a616d6d61" blob should be both static

and a value blob column is missing

stef1927:
It looks like my understanding of DynamicCompositeType was incorrect, let me try to restore the existing behaviour at a minimum, and see if we can add the aliases, if possible.

stef1927:
As discussed on CASSANDRA-11055, this is what we obtain with the latest commit:

```
CREATE TABLE legacy.composite_comp_with_col (
    key blob,
    column1 'org.apache.cassandra.db.marshal.DynamicCompositeType(org.apache.cassandra.db.marshal.BytesType, org.apache.cassandra.db.marshal.UTF8Type, org.apache.cassandra.db.marshal.TimeUUIDType)',
    "b@6869746d65776974686d75736963" blob,
    "b@6d616d6d616a616d6d61" blob,
    PRIMARY KEY (key, column1)
) WITH COMPACT STORAGE
    AND CLUSTERING ORDER BY (column1 ASC)
[...]

CREATE TABLE legacy.composite_comp_no_col (
    key blob,
    column1 'org.apache.cassandra.db.marshal.DynamicCompositeType(org.apache.cassandra.db.marshal.BytesType, org.apache.cassandra.db.marshal.UTF8Type, org.apache.cassandra.db.marshal.TimeUUIDType)',
    value blob,
    PRIMARY KEY (key, column1)
) WITH COMPACT STORAGE
    AND CLUSTERING ORDER BY (column1 ASC)
[...]
```
It is not 100% correct yet but I am not sure it's worth investing more work on this.

aholmberg:
[PYTHON-579](https://datastax-oss.atlassian.net/browse/PYTHON-579) for DCT follow-up.

PR #589, opened 2016-05-26
Don't modify the Metadatas' hosts structure while iterating over it

multani:
On Python 3.x, dict.values() returns a view object and not a list of the
values.
Updating the content of the original dict while iterating over the
values may raise a RuntimeError exception, which is the case in the
Cluster._refresh_node_list_and_token_map() method to remove discovered
hosts which are not part of the specified cluster contact points.

The bug doesn't happen using Python 2.x as dict.values() returns a new
list containing the values of the dictionary, without any other
reference to original dictionary object.

This fix forces the original Python 2.x behavior to be the same in
Python 3.x as well.

PYTHON-572

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -324,7 +324,7 @@ class Metadata(object):
         Returns a list of all known :class:`.Host` instances in the cluster.
         """
         with self._hosts_lock:
-            return self._hosts.values()
+            return list(self._hosts.values())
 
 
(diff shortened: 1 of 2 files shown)


multani:
The test I added is very basic, but sufficient to show the problem and the fix.

aholmberg:
Thanks for pointing it out. We'll get it in for the next release.

PR #592, opened 2016-06-02
Avoid LWTExceptions when updating columns that are part of the condition

eli-green:
Currently it is possible to issue an update with conditions that also makes those conditions impossible to satisfy for the subsequent DELETE statement for NULL columns. This patch removes any updated fields from the conditional statement before issuing the DELETE statement.


aholmberg:
Thanks! We'll take a look at it for the next release.
https://datastax-oss.atlassian.net/browse/PYTHON-580

PR #594, opened 2016-06-03
PYTHON-569 - optional execution profiles in core driver

aholmberg:


docs/execution_profiles.rst
```diff
@@ -0,0 +1,159 @@
+Execution Profiles (experimental)
+=================================
+
+Execution profiles are an experimental API aimed at making it easier to execute requests in different ways within
+a single connected ``Session``. Execution profiles are being introduced the exploding number of configuration options,
+especially as the database platform evolves more complex workloads.
+
+The Execution Profile API is being introduced now, in an experimental capacity in order to take advantage of it in
+existing projects, and to guage interest and feedback in the community. For now, the legacy configuration remains
```
jorgebay: ~~guage~~ gauge interest

docs/execution_profiles.rst
```diff
@@ -0,0 +1,159 @@
+Execution Profiles (experimental)
+=================================
+
+Execution profiles are an experimental API aimed at making it easier to execute requests in different ways within
+a single connected ``Session``. Execution profiles are being introduced the exploding number of configuration options,
```
jorgebay: > are being introduced the exploding

could be: "...are being introduced to deal with the exploding..."

cassandra/cluster.py
```diff
@@ -669,6 +812,24 @@ def __init__(self,
         if connection_class is not None:
             self.connection_class = connection_class
 
+        self.profile_manager = ProfileManager()
+        self.profile_manager.profiles[EXEC_PROFILE_DEFAULT] = ExecutionProfile(self.load_balancing_policy,
+                                                                               self.default_retry_policy,
+                                                                               Session._default_consistency_level,
+                                                                               Session._default_serial_consistency_level,
+                                                                               Session._default_timeout,
+                                                                               Session._row_factory)
```
aboudreault: In the next major release, are these Session attributes going to be removed? or the default exec profile will always be constructed that way?

aholmberg: If we commit to exec profiles, I would expect the session attributes to go away. For now I was just cross-referencing to show the original source, and that they default to the same values.

cassandra/cluster.py
```diff
@@ -789,6 +950,34 @@ def __init__(self, street, zipcode):
...
+        this method will wait up to five seconds for the pool creation to complete, so the profile can be used immediately
+        upon return. This behavior can be controlled using ``pool_wait_timeout`` (see
+        `concurrent.futures.wait <https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait>`_
+        for timeout semantics).
+        """
+        if not isinstance(profile, ExecutionProfile):
+            raise TypeError("profile must be an instance of ExecutionProfile")
+        if self._config_mode == _ConfigMode.LEGACY:
+            raise ValueError("Cannot add execution profiles when legacy parameters are set explicitly. TODO: link to doc")
+        self.profile_manager.profiles[name] = profile
```
aboudreault: Since the function name is "add_*", I would raise an exception if the user is trying to override an existing profile

aholmberg: good point. I could also change it to "set_*". Do you think there's any case for dynamically changing a named profile?

aboudreault: I don't think this will change dynamically, but restricting a change by only providing a "add_*" might not be ideal. +1 to change it to "set_*" and documenting that it will replace silently the profile if it already exists.

aboudreault:
+1 Awesome work \o/

cassandra/cluster.py
```diff
@@ -789,6 +951,36 @@ def __init__(self, street, zipcode):
...
+        Adding a new profile updates the connection pools according to the specified ``load_balancing_policy``. By default,
+        this method will wait up to five seconds for the pool creation to complete, so the profile can be used immediately
+        upon return. This behavior can be controlled using ``pool_wait_timeout`` (see
+        `concurrent.futures.wait <https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait>`_
+        for timeout semantics).
+        """
+        if not isinstance(profile, ExecutionProfile):
+            raise TypeError("profile must be an instance of ExecutionProfile")
+        if self._config_mode == _ConfigMode.LEGACY:
+            raise ValueError("Cannot add execution profiles when legacy parameters are set explicitly. TODO: link to doc")
```
stamhankar999: You have a stray TODO in the error message. Maybe make that into a comment above instead.

aholmberg: or just put the link in :)
Thanks

cassandra/cluster.py
```diff
@@ -1772,8 +1989,31 @@ def _create_response_future(self, query, parameters, trace, custom_payload, time
...
+                timeout = self.default_timeout
+
+            cl = query.consistency_level if query.consistency_level is not None else self.default_consistency_level
+            serial_cl = query.serial_consistency_level if query.serial_consistency_level is not None else self.default_serial_consistency_level
+
+            retry_policy = query.retry_policy or self.cluster.default_retry_policy
+            row_factory = self.row_factory
+            load_balancing_policy = self.cluster.load_balancing_policy
+        else:
+            execution_profile = self._get_execution_profile(execution_profile)
```
stamhankar999: This implies that the execution_profile could be a string (e.g. the name of a profile) or an ExecutionProfile instance. If the latter, isn't it possible that the ExecutionProfile is not configured with available hosts, not registered in the profile-manager, etc. Wouldn't it be safer to require the arg to be a registered profile name only?

aholmberg: It's useful for one-off executions, intended for after we remove legacy execute parameters. See http://datastax.github.io/python-driver/api/cassandra/cluster.html#cassandra.cluster.Session.execution_profile_clone_update, as mentioned in the docstring.

PR #596, opened 2016-06-09
PYTHON-583: Add API to get the host metadata associated with the control connection node

aboudreault:
PYTHON-583

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1364,7 +1364,7 @@ class Cluster(object):
         """
         Returns the control connection host metadata.
         """
-        connection = getattr(self.control_connection, '_connection')
+        connection = self.control_connection._connection
         host = connection.host if connection else None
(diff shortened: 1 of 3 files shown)


cassandra/cluster.py
```diff
@@ -1360,6 +1360,14 @@ def _target_type_from_refresh_args(keyspace, table, usertype, function, aggregat
             return SchemaTargetType.KEYSPACE
         return None
 
+    def get_control_connection_host(self):
+        """
+        Returns the control connection host metadata.
+        """
+        connection = getattr(self.control_connection, '_connection')
```
aholmberg: I'm just curious why this is getattr instead of direct access.

aholmberg:
+1 (with 1 minor question)

cassandra/cluster.py
aboudreault: Fixed

PR #597, opened 2016-06-14
PYTHON-577: Fix libev/asyncore re-initialization in case of a fork

aboudreault:


aholmberg:
+1

aboudreault:
Modified the PR to only include the _dispatcher_map change

PR #599, opened 2016-06-18
cassandra_init() added **kwargs

LutzSteinborn:
Added **kwargs because without you are getting an exception about unknown kwargs.

### DIFF
--- a/docs/cqlengine/third_party.rst
+++ b/docs/cqlengine/third_party.rst
@@ -17,7 +17,7 @@ Here's how, in substance, CQLengine can be plugged to `Celery
     from cqlengine.connection import (
         cluster as cql_cluster, session as cql_session)
 
-    def cassandra_init():
+    def cassandra_init(**kwargs):
         """ Initialize a clean Cassandra connection. """
         if cql_cluster is not None:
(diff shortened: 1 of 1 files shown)


aholmberg:
I don't see any indication that these handlers require arguments.
http://docs.celeryproject.org/en/latest/userguide/signals.html#worker-process-init
http://uwsgi-docs.readthedocs.io/en/latest/PythonDecorators.html#uwsgidecorators.postfork
Can you describe a little more about what you're seeing? Do you have some example code?

LutzSteinborn:
Please look at my question at stackoverflow:
http://stackoverflow.com/questions/37774330/cant-bring-cassandra-and-celery-together/37819775#37819775

The answer is from the creator of Celery. So may be he knows much better then me whats going on.

aholmberg:
Thanks for the follow-up. From the link http://docs.celeryproject.org/en/latest/userguide/signals.html#basics

> The best practice for signal handlers is to accept arbitrary keyword arguments (i.e. **kwargs). That way new celery versions can add additional arguments without breaking user code.

Thanks for the report!

PR #601, opened 2016-06-28
Fix unsupported protocol version error message

thobbs:
I noticed this while working on [CASSANDRA-11464](https://issues.apache.org/jira/browse/CASSANDRA-11464).


aholmberg:
Thanks!

PR #602, opened 2016-06-29
Comparison operators for Columns

richardasaurus:
A small change but important for a 3rd party lib which uses these.

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -168,6 +168,36 @@ class Column(object):
         self.position = Column.instance_counter
         Column.instance_counter += 1
 
+    def __ne__(self, other):
+        if isinstance(other, Column):
+            return self.position != other.position
+        return NotImplemented
+
+    def __eq__(self, other):
+        if isinstance(other, Column):
+            return self.position == other.position
+        return NotImplemented
+
+    def __lt__(self, other):
+        if isinstance(other, Column):
+            return self.position < other.position
+        return NotImplemented
+
+    def __le__(self, other):
+        if isinstance(other, Column):
+            return self.position <= other.position
+        return NotImplemented
+
+    def __gt__(self, other):
+        if isinstance(other, Column):
+            return self.position > other.position
+        return NotImplemented
+
+    def __ge__(self, other):
+        if isinstance(other, Column):
+            return self.position >= other.position
+        return NotImplemented
+
     def validate(self, value):
         """
         Returns a cleaned and validated value. Raises a ValidationError


cassandra/cqlengine/columns.py
```diff
@@ -72,6 +73,7 @@ def get_property(self):
             return property(_get, _set)
 
 
+@total_ordering
```
aholmberg: We still support Python 2.6. This was added in 2.7. We may need to think of alternatives.

richardasaurus: I could add this backport https://pypi.python.org/pypi/total-ordering

Or just add all the ordering methods to the class. Whichever you prefer

aholmberg: If you're willing, in this case I think we would prefer writing the comparison methods ([trying to minimize dependencies](https://github.com/datastax/python-driver/blob/master/CONTRIBUTING.rst#design-and-implementation-guidelines)).

richardasaurus:
@aholmberg made that change

aholmberg:
Thanks. We'll get it in the next release.

aholmberg:
https://datastax-oss.atlassian.net/browse/PYTHON-595

cassandra/cqlengine/columns.py
```diff
@@ -168,6 +168,36 @@ def __init__(self,
         self.position = Column.instance_counter
         Column.instance_counter += 1
 
+    def __ne__(self, other):
+        if isinstance(other, Column):
+            return self.instance_counter != other.instance_counter
```
aholmberg: These comparisons all use ``Column.instance_counter``, which is the same for all instances. Do you want to use ``self.position`` instead?

cassandra/cqlengine/columns.py
```diff
@@ -168,6 +168,36 @@ def __init__(self,
...
+            return self.instance_counter == other.instance_counter
+        return NotImplemented
+
+    def __lt__(self, other):
+        if isinstance(other, Column):
+            return self.instance_counter < other.instance_counter
+        return NotImplemented
+
+    def __le__(self, other):
+        if isinstance(other, Column):
```
aholmberg: Any thoughts about implementing these others in terms of ``__lt__`` and ``__eq__`` to avoid writing this type check so many times?

cassandra/cqlengine/columns.py
richardasaurus: yes that's right will change it

cassandra/cqlengine/columns.py
richardasaurus: you mean like?
```python
     def __le__(self, other):
        return (self.instance_counter < other.instance_counter or 
                self.instance_counter == other.instance_counter)
```

aholmberg: That is what I meant, but now that you write it out, it doesn't seem to buy much. I guess I'm ambivalent on that.

aholmberg:
Thanks for the update. Will write a few tests and merge tomorrow.

richardasaurus:
Great thanks

PR #606, opened 2016-07-04
Fix for CASSANDRA-11850

stef1927:
As discussed in the ticket, cqlsh won't start with Python 2.7.12.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2144,7 +2144,7 @@ class ControlConnection(object):
         # _clear_watcher will be called when this ControlConnection is about to be finalized
         # _watch_callback will get the actual callback from the Connection and relay it to
         # this object (after a dereferencing a weakref)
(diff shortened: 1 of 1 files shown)


khaledh:
I'm assuming this fix targets 2.7.2, but it seems the 2.7.2 package doesn't have it. How do we get this fix?

aholmberg:
This was added to 2.7.2 just for integration into Cassandra 2.1.  To get it in a release, please see 3.4+
https://github.com/datastax/python-driver/commit/1afa1f4b2f29fafa99b7f5e551b717278df5515f

khaledh:
Thanks. We're currently evaluating upgrading to the latest driver version.

aholmberg:
Just to make sure you're aware: http://datastax.github.io/python-driver/upgrading.html#upgrading-to-3-0

PR #609, opened 2016-07-06
 PYTHON-71 - use prepared metadata and skip row result meta

aholmberg:
in protocol v2+

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2045,11 +2045,11 @@ class Session(object):
                 query_string, cl, serial_cl,
                 fetch_size, timestamp=timestamp)
         elif isinstance(query, BoundStatement):
-            message = ExecuteMessage(
-                query.prepared_statement.query_id, query.values, cl,
-                serial_cl, fetch_size,
-                timestamp=timestamp)
             prepared_statement = query.prepared_statement
+            message = ExecuteMessage(
+                prepared_statement.query_id, query.values, cl,
+                serial_cl, fetch_size,
+                timestamp=timestamp, skip_meta=bool(prepared_statement.result_metadata))
         elif isinstance(query, BatchStatement):
             if self._protocol_version < 2:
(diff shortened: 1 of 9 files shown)


cassandra/query.py
```diff
@@ -362,36 +361,34 @@ class PreparedStatement(object):
     may affect performance (as the operation requires a network roundtrip).
     """
 
-    column_metadata = None
+    column_metadata = None  #TODO: make this bind_metadata in next major
```
aboudreault: added in PYTHON-504

cassandra/query.py
```diff
@@ -362,36 +361,34 @@ class PreparedStatement(object):
...
+    consistency_level = None
+    custom_payload = None
+    fetch_size = FETCH_SIZE_UNSET
+    keyspace = None  # change to prepared_keyspace in major release
```
aboudreault: added in PYTHON-504

cassandra/cluster.py
```diff
@@ -2124,14 +2124,14 @@ def prepare(self, query, custom_payload=None):
         future = ResponseFuture(self, message, query=None, timeout=self.default_timeout)
         try:
             future.send_request()
-            query_id, column_metadata, pk_indexes = future.result()
+            query_id, bind_metadata, pk_indexes, result_metadata = future.result()
```
aboudreault: In the future, if we extend that result part again, we might prefer to create a structure for the future result attributes.

aboudreault:
LGTM!

cassandra/cluster.py
aholmberg: good idea. There are several places where this would be useful.

aholmberg: @aboudreault I created https://datastax-oss.atlassian.net/browse/PYTHON-601 to refactor at some point. We might be able to work it into this release if there is time.

PR #613, opened 2016-07-07
PYTHON-580 - cqlengine conditional handling for update delete phase

aholmberg:
builds on #592, fixing broken non-conditional queries


aboudreault:
+1

PR #615, opened 2016-07-08
Fix problem in detecting schema mismatch due to peer.is_up = None

stef1927:
Whilst working on CASSANDRA-11950, I was having problems eliminating `UnknownColumnFamilyException` warnings from the logs. The cause is two-fold:

* `peer.is_up` can be None rather than True
* the distance can be `HostDistance.IGNORED` for remote nodes, unless we use a DC aware load balancing policy with `used_hosts_per_remote_dc` set to the maximum number of remote nodes.

I'm not sure we should exclude nodes with distance ignored because even if the client doesn't contact them directly, they can still have mutations propagated to them, for example at EACH_QUORUM, at a minimum could we make this configurable?

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2945,7 +2945,10 @@ class ControlConnection(object):
                 continue
             addr = self._rpc_from_peer_row(row)
             peer = self._cluster.metadata.get_host(addr)
-            if peer and peer.is_up and pm.distance(peer) != HostDistance.IGNORED:
+            # Are we sure we need to consider the distance? Even if we do not contact
+            # a remote host directly, mutations are still propagated to it, can we at
+            # least have a parameter to ignore it?
+            if peer and peer.is_up is not False and pm.distance(peer) != HostDistance.IGNORED:
                 versions[schema_ver].add(addr)
 
         if len(versions) == 1:


aholmberg:
Unfortunately this is an imperfect mechanism. I originally introduced that logic to address [PYTHON-531](https://datastax-oss.atlassian.net/browse/PYTHON-531), where it appeared that schema migrations were failing when another node was down.

> distance ignored because even if the client doesn't contact them directly, they can still have mutations propagated to them

This is a good point. Since 531 I've [added] (https://github.com/datastax/python-driver/commit/01a38259e4a1864db79e7c1e5e2e469a3372bf31) [several](https://github.com/datastax/python-driver/commit/8addbad42c428f0086b13b09495b4d2b1950243f) [changes] (https://github.com/datastax/python-driver/commit/dcd47a3a943036d8e65de52b826fd2ca8b4dae85) improving host state management. With those ``is_up is not False`` should have us covered, and allow us to remove that distance conditional.

aholmberg:
I incorporated your change and removed host distance in a new PR ^. lmk if this suits your needs.
Thanks for the input on this.

stef1927:
PR https://github.com/datastax/python-driver/pull/617 should cover this perfectly, closing this PR. Thank you for fixing this so quickly!

PR #617, opened 2016-07-08
Don't omit ignored nodes from schema agreement, if they're up

aholmberg:
Follow-on to #615


stef1927:
Thanks for this, it should work perfectly!

PR #619, opened 2016-07-14
Removed an inconsistent sleep use in a test

aboudreault:
@GregBestland  to review


GregBestland:
+1 this is the right way to do this.

PR #621, opened 2016-07-15
Correct typo in partition_key check

BenBrostoff:
I think the attribute referenced should be <code>partition_key</code>. I noticed this issue when I received an <code>AttributeError</code> upon trying to delete a row.


aboudreault:
Thank you Ben!

PR #622, opened 2016-07-18
Extend no-op model update tests to primary keys

kdeldycke:
This PR extends the initial no-op model update unit-tests to primary keys, as I expect the update operation to be skipped and ignored entirely if the value of the primary key hasn't changed a bit, mirroring the current behaviour on arbitrary columns.

I also added a new unit test section covering the same update operations trough the use of value assignation.

### DIFF
--- a/tests/integration/cqlengine/model/test_updates.py
+++ b/tests/integration/cqlengine/model/test_updates.py
@@ -79,8 +79,8 @@ class ModelUpdateTests(BaseCassEngTestCase):
         self.assertEqual(m2.count, m1.count)
         self.assertEqual(m2.text, m0.text)
 
-    def test_noop_model_update(self):
-        """ tests that calling update on a model with no changes will do nothing. """
+    def test_noop_model_direct_update(self):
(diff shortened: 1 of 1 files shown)


kdeldycke:
For context, this covers an old issue I opened in 2014: https://github.com/cqlengine/cqlengine/pull/212

The root cause has been addressed since then in the the core of the driver. Hence the current PR, covering for regression.

GregBestland:
I had to slightly modify the noop validation tests to account for client side validation around primary keys. Let me know if that doesn't track with your intentions.

kdeldycke:
Thanks @GregBestland for the quick merge! :)

The failing of the tests you adjusted was my main point all along. It's just that I [realised 2 days ago](https://github.com/datastax/python-driver/pull/623#issuecomment-233426580) integration tests were not launched by Travis or a default `tox` call. So I had the false impression my issue was resolved.

Anyway, thanks for fixing it, at least we now have proper coverage of the current behaviour. I'll propose a change of the way this is validated in another PR to keep the discussion compartmented. Let's consider this PR done in the mean time.

PR #624, opened 2016-07-18
PYTHON-538: cqlengine deprecate Model.__default_ttl__

aboudreault:
__default_ttl__ will be deprecated in 4.0. We now relies on the cassandra default_time_to_live feature.

### DIFF
--- a/docs/cqlengine/queryset.rst
+++ b/docs/cqlengine/queryset.rst
@@ -343,6 +343,42 @@ None means no timeout.
     Setting the timeout on the model is meaningless and will raise an AssertionError.
 
 
+.. _ttl-change:
+
+Default TTL and Per Query TTL
+=============================
+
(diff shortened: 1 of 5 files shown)


docs/cqlengine/queryset.rst
```diff
@@ -343,6 +343,42 @@ None means no timeout.
...
+        class User(Model):
+            __options__ = {'default_time_to_live': 20}
+
+            user_id = columns.UUID(primary_key=True)
+            ...
+
+        User.objects.create(user_id=1)  # Default TTL 20 will be set automatically on the server
+
+        User.objects.ttl(10).create(user_id=1)  # TTL 10
+        User.objects.create(user_id=1).ttl(10).update(age=21)  # TTL 10
```
aholmberg: create and update on the same line are a little strange. Would it make sense to use the instance from the create above?

aboudreault: +1

aholmberg:
+1
You made this very straightforward. Nice work.

PR #623, opened 2016-07-18
PYTHON-609: Align Text and Ascii columns behaviour

kdeldycke:
This PR:
* Add full unit-tests of CQLengine's `Ascii` columns.
* Extend `Text` columns unit-tests for minimal and maximal length.
* Thoroughly cover edge-cases on empty / blank / `None` value strings handling, especially in relation to the `required` column parameter.
* Align `Ascii` column behaviour to `Text`, by adding `min_length` and `max_length` parameter support.

PR #628, opened 2016-07-20
Python 603

GregBestland:
Contains additional tests from pr-623, and tries to standardize the way we do validation for ASCII and Text types in clqengine.


kdeldycke:
Oh no. I just pushed my work on https://github.com/datastax/python-driver/pull/623 :(

GregBestland:
closed in favor of
https://github.com/datastax/python-driver/pull/623

PR #630, opened 2016-07-20
PYTHON-600 - make connection timeout infinite by default

aholmberg:
removes the need for eventlet and gevent checking for timeouts


aboudreault:
+1

PR #632, opened 2016-07-22
Raise error on primary key update only if its value changed

kdeldycke:
This PR follows up on #622, and only raise an error on primary key update attempt if their values effectively changed.

### DIFF
--- a/cassandra/cqlengine/models.py
+++ b/cassandra/cqlengine/models.py
@@ -705,18 +705,23 @@ class BaseModel(object):
         It is possible to do a blind update, that is, to update a field without having first selected the object out of the database.
         See :ref:`Blind Updates <blind_updates>`
         """
-        for k, v in values.items():
-            col = self._columns.get(k)
+        for column_id, v in values.items():
+            col = self._columns.get(column_id)
 
             # check for nonexistant columns
             if col is None:
-                raise ValidationError("{0}.{1} has no column named: {2}".format(self.__module__, self.__class__.__name__, k))
+                raise ValidationError(
+                    "{0}.{1} has no column named: {2}".format(
+                        self.__module__, self.__class__.__name__, column_id))
+
+            setattr(self, column_id, v)
 
             # check for primary key update attempts
-            if col.is_primary_key:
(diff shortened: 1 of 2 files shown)


kdeldycke:
Any news on that one? :)

aholmberg:
Sorry for the delay. We've had some changes on this team, and this slipped through the cracks. Will get it in the next minor release.

cassandra/cqlengine/models.py
```diff
@@ -705,18 +705,23 @@ def update(self, **values):
...
             if col is None:
-                raise ValidationError("{0}.{1} has no column named: {2}".format(self.__module__, self.__class__.__name__, k))
+                raise ValidationError(
+                    "{0}.{1} has no column named: {2}".format(
+                        self.__module__, self.__class__.__name__, column_id))
 
-            # check for primary key update attempts
-            if col.is_primary_key:
-                raise ValidationError("Cannot apply update to primary key '{0}' for {1}.{2}".format(k, self.__module__, self.__class__.__name__))
+            setattr(self, column_id, v)
```
kdeldycke: No, the idea is to update the column, whatever its kind (primary column or not), with its new value. Then let the column reports itself its mutation, based on the value alone, through the `get_changed_columns()` method.

`get_changed_columns()` is smart enough to figure out if the column really changed, thanks to its state tracking.

mambocab: Sorry, I wasn't clear.

Imagine a model `m` where `m.pk == 0`. If a user calls `m.update(pk=1)`, where `pk` is the primary key:

- on `master`, `update` will throw a `ValidationError`, after which `m.pk == 0`.
- on this branch, `update` will throw a `ValidationError`, but `m.pk == 1`.

Am I right? and if so, did you intend to change `update`'s behavior this way? If not, could you please fix this, and add a test for that regression? Thanks!

kdeldycke: Thanks @mambocab for the feedback! I may have overlooked something here. I'll double check that soon.

aboudreault:
Closing in favor of #732 with a slightly different version. Thank Kevin for this.

kdeldycke:
No worries. And sorry for the delay.

PR #634, opened 2016-07-22
PYTHON-561: Changed per-cluster metrics path and added ability to define an explicit stats name

aboudreault:


GregBestland:
+1 LGTM I'm going to merge this in.

PR #636, opened 2016-07-27
PYTHON-598: Add ContextQuery to allow switching the model keyspace easily

aboudreault:
* Add ContextQuery
* Add multiple keyspaces support to sync_table
* Add multiple keyspaces support to drop_table

missing doc...

### DIFF
--- a/cassandra/cqlengine/management.py
+++ b/cassandra/cqlengine/management.py
@@ -121,7 +121,9 @@ def _get_index_name_by_column(table, column_name):
 
 def sync_table(model, keyspaces=None):
     """
-    Inspects the model and creates / updates the corresponding table and columns for all keyspaces.
+    Inspects the model and creates / updates the corresponding table and columns.
+
+    If `keyspaces` is specified, the table will be synched for all specified keyspaces. Note that the `Model.__keyspace__` is ignored in that case.
 
     Any User Defined Types used in the table are implicitly synchronized.
 
@@ -447,7 +449,9 @@ def _update_options(model):
 
 def drop_table(model, keyspaces=None):
     """
-    Drops the table indicated by the model, if it exists, for all keyspaces.
+    Drops the table indicated by the model, if it exists.
+
+    If `keyspaces` is specified, the table will be dropped for all specified keyspaces. Note that the `Model.__keyspace__` is ignored in that case.
 
     **This function should be used with caution, especially in production environments.
     Take care to execute schema modifications in a single context (i.e. not concurrently with other clients).**
--- a/cassandra/cqlengine/management.py
+++ b/cassandra/cqlengine/management.py
@@ -21,7 +21,7 @@ import warnings
 
 from cassandra import metadata
 from cassandra.cqlengine import CQLEngineException
-from cassandra.cqlengine import columns
+from cassandra.cqlengine import columns, query
 from cassandra.cqlengine.connection import execute, get_cluster
 from cassandra.cqlengine.models import Model
 from cassandra.cqlengine.named import NamedTable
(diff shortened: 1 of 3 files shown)


cassandra/cqlengine/management.py
```diff
@@ -119,9 +119,9 @@ def _get_index_name_by_column(table, column_name):
             return index_metadata.name
 
 
-def sync_table(model):
+def sync_table(model, keyspaces=None):
     """
-    Inspects the model and creates / updates the corresponding table and columns.
+    Inspects the model and creates / updates the corresponding table and columns for all keyspaces.
```
aholmberg: "For default, or any specified keyspaces" (?)

aboudreault: For all keyspaces specified. umm... should we automatically include the default Class.__keyspace__ in the keyspaces list? it is not at the moment.

cassandra/cqlengine/management.py
```diff
@@ -431,15 +445,27 @@ def _update_options(model):
     return False
 
 
-def drop_table(model):
+def drop_table(model, keyspaces=None):
     """
-    Drops the table indicated by the model, if it exists.
+    Drops the table indicated by the model, if it exists, for all keyspaces.
```
aholmberg: same comment tweak question as above

cassandra/cqlengine/management.py
```diff
@@ -135,6 +135,20 @@ def sync_table(model):
 
     *There are plans to guard schema-modifying functions with an environment-driven conditional.*
     """
+
+    if keyspaces:
+        if not isinstance(keyspaces, (list, tuple)):
+            raise ValueError('keyspaces must be a list or a tuple.')
+
+        for keyspace in keyspaces:
```
aholmberg: Do you think it's worth also including `Model.__keyspace__`, if set? Otherwise we might want to explain that the `keyspaces` parameter overrides model attribute and default keyspace.

cassandra/cqlengine/management.py
```diff
@@ -431,15 +445,27 @@ def _update_options(model):
...
     Take care to execute schema modifications in a single context (i.e. not concurrently with other clients).**
 
     *There are plans to guard schema-modifying functions with an environment-driven conditional.*
     """
+
+    if keyspaces:
+        if not isinstance(keyspaces, (list, tuple)):
+            raise ValueError('keyspaces must be a list or a tuple.')
+
+        for keyspace in keyspaces:
```
aholmberg: same question about including `Model.__keyspace__`

cassandra/cqlengine/management.py
aboudreault: I wont. I will add a comment in the doc to be clear.

aboudreault:
pushed a small docs improvement

cassandra/cqlengine/management.py
```diff
@@ -121,7 +121,9 @@ def _get_index_name_by_column(table, column_name):
 
 def sync_table(model, keyspaces=None):
     """
-    Inspects the model and creates / updates the corresponding table and columns for all keyspaces.
+    Inspects the model and creates / updates the corresponding table and columns.
+
+    If `keyspaces` is specified, the table will be synched for all keyspaces. Note that the `Model.__keyspace__` is ignored in that case.
```
aholmberg: I suggest "all *specified* keyspaces". The way its worded now makes it sound like it will put the table in every keyspace on the cluster. :)

aholmberg:
+1 with nit in doc wording

cassandra/cqlengine/management.py
aboudreault: I agree! I will fix that! Thanks!

PR #637, opened 2016-07-28
[CQLengine] Set default value of columns on instantiation

kdeldycke:
I just stumbled upon an inconsistency regarding the way `get_changed_columns()` reports columns whose values have been updated.

`get_changed_columns()` in itself works fine. The heavy lifting has been implemented a year ago through #373 and #348, and is upstream since the early days of v3.0.0.

But if the behaviour of [`get_changed_columns()` is thoroughly covered by unit-tests](https://github.com/datastax/python-driver/commit/17f2bc13c2fda731e9de8217790730b151f962ec), there's an edge case when an instance hasn't been persisted yet. This issue only concerns unset columns with a default value.

This issue is demonstrated in the unit-tests attached to this PR. The fix consists in setting the default value of a column right away at instantiation if no value has been provided by the user.

On my machine, no regression have been detected by both the `unit` and `integration` test suites.

### DIFF
--- a/cassandra/cqlengine/models.py
+++ b/cassandra/cqlengine/models.py
@@ -361,19 +361,16 @@ class BaseModel(object):
 
         self._values = {}
         for name, column in self._columns.items():
-            value = values.get(name)
+            # Set default values on instantiation. Thanks to this, we don't have
+            # to wait anylonger for a call to validate() to have CQLengine set
+            # default columns values.
+            column_default = column.get_default() if column.has_default else None
+            value = values.get(name, column_default)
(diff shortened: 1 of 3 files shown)


aholmberg:
Thanks. I'll take a look today.

aholmberg:
Thanks for another contribution!

kdeldycke:
Thanks @aholmberg for the review and quick merge! Really happy to contribute! :)

kdeldycke:
As a side note, this PR relates to a very old one: https://github.com/cqlengine/cqlengine/pull/189

PR #640, opened 2016-08-04
PYTHON-614 - allow beta proto version flag

aholmberg:


thobbs:
FYI, this doesn't handle the beta flag in `_ProtocolHandler.decode_message`, so you'll get warning logs like these:

~~~
cassandra.protocol: WARNING: Unknown protocol flags set: 10. May cause problems.
~~~

GregBestland:
+1 LGTM I'm going to merge this now.

PR #643, opened 2016-08-08
PYTHON-618 - add hash function to cqlengine.columns.Column

aholmberg:
restore hashability to Column in Python 3.x


GregBestland:
+1 LGTM I'm merging this

PR #644, opened 2016-08-08
PYTHON-379 - avoid quadratic token processing for RF > nodes

aholmberg:


GregBestland:
This looks good to me. +1 Merging.

PR #646, opened 2016-08-11
PYTHON-284 - request size attribute, request init listener

aholmberg:


GregBestland:
+1 LGTM, good design, easy to follow, and optional as to not affect performance unless you want it.

I'm merging this into master.

PR #653, opened 2016-08-18
Fix to keep type queries from breaking metadata construction in C* <2.1

GregBestland:
This should keep the new metadata type query from blowing up in C* 2.0

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -1598,9 +1598,16 @@ class _SchemaParser(object):
 
     def _query_build_rows(self, query_string, build_func):
         query = QueryMessage(query=query_string, consistency_level=ConsistencyLevel.ONE)
-        response = self.connection.wait_for_response(query, self.timeout)
-        result = dict_factory(*response.results)
-        return [build_func(row) for row in result]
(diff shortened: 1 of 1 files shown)


cassandra/metadata.py
```diff
@@ -1598,9 +1598,13 @@ def _query_build_row(self, query_string, build_func):
...
-        response = self.connection.wait_for_response(query, self.timeout)
-        result = dict_factory(*response.results)
-        return [build_func(row) for row in result]
+        responses = self.connection.wait_for_responses((query), timeout=self.timeout, fail_on_error=False)
+        (success, response) = responses[0]
+        if success:
+            result = dict_factory(*response.results)
+            return [build_func(row) for row in result]
+        else:
+            return []
```
aholmberg: One minor question: curious if you thought about handling the exception by type [like we do on the main query path](https://github.com/datastax/python-driver/blob/3.6.0/cassandra/metadata.py#L2023-L2027) (this can prevent it swallowing different kinds of unexpected errors).

GregBestland:
I added an instance check as prescribed.

PR #655, opened 2016-08-20
Adds license information to the setup function

anweshadas:
As suggested in https://packaging.python.org/distributing/#license

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -403,6 +403,7 @@ def run_setup(extensions):
         description='Python driver for Cassandra',
         long_description=long_description,
         url='http://github.com/datastax/python-driver',
+        license='ASL 2.0',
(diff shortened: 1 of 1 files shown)


aholmberg:
Can you help me understand what you are trying to accomplish with this? Are there canonical specifications for the license string? Does it buy you anything when the [license is already defined](https://github.com/datastax/python-driver/blob/3.6.0/setup.py#L416) in the [Trove classifiers](https://pypi.python.org/pypi?%3Aaction=list_classifiers)?
According to [this](https://docs.python.org/2/distutils/setupscript.html#additional-meta-data) (note 6), the license attribute is used when *not* defining it in the classifiers.

aholmberg:
closing after some consideration and no response from user

PR #656, opened 2016-08-26
CLQEngine multiple sessions/keyspaces support

aboudreault:


cassandra/cqlengine/named.py
```diff
@@ -86,6 +87,13 @@ class NamedTable(object):
 
     _partition_key_index = None
 
+    __connection__ = None
+    _connection = None
+
+    using = UsingDescriptor()
+
+    _get_connection = BaseModel._get_connection
```
aboudreault: this has to be verified... not sure it's safe

cassandra/cqlengine/connection.py
```diff
@@ -28,67 +28,200 @@
 
 NOT_SET = _NOT_SET  # required for passing timeout to Session.execute
 
-Host = namedtuple('Host', ['name', 'port'])
-
-cluster = None
```
aholmberg: I like the object oriented approach, but wondering if we should set these module attributes when the default connection is made, so we don't lose attributes that were not explicitly "private".

cassandra/cqlengine/connection.py
```diff
@@ -28,67 +28,200 @@
...
-session = None
-lazy_connect_args = None
-lazy_connect_lock = threading.RLock()
-
+# connections registry
+DEFAULT_CONNECTION = '_default_'
```
aholmberg: Any reason we shouldn't use an `object` default key?

cassandra/cqlengine/connection.py
```diff
@@ -28,67 +28,200 @@
...
+_connections = {}
 
 # Because type models may be registered before a connection is present,
 # and because sessions may be replaced, we must register UDTs here, in order
 # to have them registered when a new session is established.
 udt_by_keyspace = defaultdict(dict)
 
 
+def format_log_context(msg, connection=None, keyspace=None):
```
aholmberg: Should this be `connection_name` so I don't get confused about passing connection objects in?

cassandra/cqlengine/management.py
```diff
@@ -119,7 +161,7 @@ def _get_index_name_by_column(table, column_name):
             return index_metadata.name
 
 
-def sync_table(model, keyspaces=None):
+def sync_table(model, keyspaces=None, connections=None):
```
aholmberg: connections in docstring

cassandra/cqlengine/management.py
```diff
@@ -441,13 +488,13 @@ def _update_options(model):
...
         query = "ALTER TABLE {0} WITH {1}".format(model.column_family_name(), options)
-        execute(query)
+        execute(query, connection=connection)
         return True
 
     return False
 
 
-def drop_table(model, keyspaces=None):
+def drop_table(model, keyspaces=None, connections=None):
```
aholmberg: connections in docstring

cassandra/cqlengine/management.py
```diff
@@ -37,7 +38,24 @@
 schema_columnfamilies = NamedTable('system', 'schema_columnfamilies')
 
 
-def create_keyspace_simple(name, replication_factor, durable_writes=True):
+def get_context(keyspaces, connections):
```
aholmberg: module private?

cassandra/cqlengine/named.py
aholmberg: Still looking into this? Seems like it should work...

cassandra/cqlengine/connection.py
aboudreault: I would suggest do it for 3.x, but remove in 4.0. So cluster will always point to the DEFAULT cluster object.

cassandra/cqlengine/connection.py
aboudreault: no reason. good idea

cassandra/cqlengine/connection.py
aboudreault: based on our discussion, is it OK to keep this as is, since the Connection objects are kindly private in cqlengine.connection ?

cassandra/cqlengine/management.py
aboudreault: +1

cassandra/cqlengine/named.py
aboudreault: alright. I wasn't sure if the hybrid_method could be binded to the other class or something.

aboudreault:
Addressed comments and added tests

cassandra/cqlengine/connection.py
```diff
@@ -43,7 +43,9 @@
 
 def format_log_context(msg, connection=None, keyspace=None):
     """Format log message to add keyspace and connection context"""
-    connection_info = connection if connection else DEFAULT_CONNECTION
+    connection_info = connection
+    if not connection_info:
```
aholmberg: Could this be simplified to `connection_info = connection or 'DEFAULT_CONNECTION'`?

aboudreault: well, I need a string representation of the default connection, for the log... since I changed the DEFAULT_CONNECTION key to an object

aboudreault: oh, yeah. totally.

aholmberg:
+1

PR #659, opened 2016-08-31
PYTHON-218 added test for speculative execution policies

GregBestland:
Looking to make sure I didn't miss anything major here with regards to code paths.

### DIFF
--- a/tests/integration/standard/test_policies.py
+++ b/tests/integration/standard/test_policies.py
@@ -1,0 +1,99 @@
+# Copyright 2013-2016 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


tests/integration/standard/test_policies.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # This LBP should keep host list to 3
+        result = self.session.execute(statement, execution_profile='spec_ep_rr')
+        self.assertEqual(3, len(result.response_future.attempted_hosts))
+        # Spec_execution policy should limit retries to 1
+        result = self.session.execute(statement, execution_profile='spec_ep_rr_lim')
+
+        self.assertEqual(2, len(result.response_future.attempted_hosts))
+        # Spec_execution policy should not be used if  the query is not idempotent
+        result = self.session.execute(statement_non_idem)
```
aholmberg: I think this is using default profile, which does not have spec exec. Do we want to add a test for non-idempotent+spec exec policy?

GregBestland: That was my intent here actually good catch.

PR #666, opened 2016-09-19
Add NetworkAwareRoundRobinPolicy load balancer

jolynch:
_What_
This is my first attempt at a Rack aware (and DC aware) load balancing strategy, which is currently called "NetworkAwareRoundRobinPolicy" (a nod to the NetworkTopologyStrategy) which routes requests to local rack nodes first and later to remote nodes. I've included some additional tests and DCAware tests all pass using NetworkAware as the underlying lb.

If you guys like this idea I can roll it out to our deployment  (few hundred C\* nodes, few thousand clients) and ensure that it gives the expected benefits.

_Why_
In cloud providers like AWS, racks are often availability zones (and RF = # of racks), and routing to the wrong availability zone can cost a few milliseconds of latency on read requests. This is particularly important when comparing point query performance of LOCAL_ONE Cassandra queries against say something like AZ local MySQL or memcache.  The goal of this change is to maintain backwards compatibility with DCAware (which is probably better in general), but allow performance sensitive clients to always pick rack local coordinators.

_Concerns_
I'm slightly concerned that this could cause hotspotting on particular racks, and as such I don't think it should be the default strategy. Also the benefit for quorum operations would be not that great since the coordinator will still have to reach cross rack, but at least it saves a single cross AZ trip. If you have further concerns, or think this should be separate, or not in the main repo at all, I dig it, just wanted to contribute this and see where it went.

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from itertools import islice, cycle, groupby, repeat
+from itertools import islice, cycle, groupby, repeat, chain
 import logging
 from random import randint
 from threading import Lock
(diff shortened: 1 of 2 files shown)


jolynch:
Cut PYTHON-639 and linked it :-)

mambocab:
Thanks for the PR and making a ticket. I've scheduled review for version 3.9.

mambocab:
Closing -- see [PYTHON-639](https://datastax-oss.atlassian.net/browse/PYTHON-639) for a discussion of our decision-making process there.

PR #669, opened 2016-10-12
PYTHON-631 consistent error message on deserialization failure

aboudreault:


cassandra/protocol.py
```diff
@@ -648,7 +648,7 @@ def recv_results_rows(cls, f, protocol_version, user_type_map, result_metadata):
                 except Exception as e:
                     raise DriverException('Failed decoding result column "%s" of type %s: %s' % (colnames[i],
                                                                                                  coltypes[i].cql_parameterized_type(),
-                                                                                                 e.message))
+                                                                                                 str(e)))
```
aboudreault: Exception.message is deprecated in Python3.

cassandra/row_parser.pyx
```diff
@@ -33,7 +34,16 @@ def make_recv_results_rows(ColumnParser colparser):
...
+        try:
+            parsed_rows = colparser.parse_rows(reader, desc)
+        except Exception as e:
+            # Use explicitly the TupleRowParser to display better error messages for column decoding failures
+            rowparser = TupleRowParser()
+            reader.buf_ptr = reader.buf
+            reader.pos = 0
+            rowcount = read_int(reader)
+            for i in range(rowcount):
+                rowparser.unpack_row(reader, desc)
```
aboudreault: The real fix is in the TupleRowParser class. The only reason this try/except block exists is in case some users are using custom parser ... written in Cython. Do they have any reason to do that? I'm not entirely sure it is worth to handle this case here and would be in favor to remove that part. What do you think?

aholmberg: +1 I would be okay removing, as you suggest. I'm not aware of anyone using custom parsers outside the ones we provide.

aboudreault:
Sure. will do.

PR #670, opened 2016-10-20
Fixup attribute name on cluster reconnect

farragar:
We were having some trouble reconnecting to the cluster after all nodes went down. During investigation, we found what looks like a big typo here.


aboudreault:
Thanks!

PR #671, opened 2016-10-26
Fix for broken Cython dependency. (PYTHON-656)

colinmcintosh:


aboudreault:
Thank you for the patch. I pushed an alternative fix here https://github.com/datastax/python-driver/commit/4f00f5fada77ec2b33bdda6af01c6d4895555046 , to only cap the upper range of the dependency.

PR #673, opened 2016-10-27
Quote index name in cql string where necessary

beobal:
See CASSANDRA-12847

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -1336,14 +1336,14 @@ class IndexMetadata(object):
         index_target = options.pop("target")
         if self.kind != "CUSTOM":
             return "CREATE INDEX %s ON %s.%s (%s)" % (
-                self.name,  # Cassandra doesn't like quoted index names for some reason
+                protect_name(self.name),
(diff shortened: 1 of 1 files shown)


aholmberg:
I'm not sure we can do this blindly since it may not work with older versions, which we still support.

beobal:
> Can we assume that any index identifier that would need quotes could not have existed in earlier versions of Cassandra?

I think so, yes, because prior to CASSANDRA-8365 quotes would've been rejected if supplied, and if a mixed/upper case name was used, it would've been automatically downcased like any other unquoted identifier. So if the identifier requires quoting, we must be talking to C* >= 2.1.3

aholmberg:
Sounds good. I can't figure out why it was ever needed, since `protect_name` is a "maybe" based on identifier case.
Thanks for the PR

stef1927:
This should fix [PYTHON-616](https://datastax-oss.atlassian.net/browse/PYTHON-616), right?

aholmberg:
@stef1927 yes. Thanks for mentioning that link.

PR #674, opened 2016-10-28
Expand QUERY/EXECUTE/BATCH flags from [byte] to [int]

stef1927:
This is the pull request for [CASSANDRA-12838](https://issues.apache.org/jira/browse/CASSANDRA-12838).

Please don't merge it into cassandra-test until CASSNDRA-12838 is committed, in fact is there a way to coordinate this so that we avoid having dtests fail?


aholmberg:
Thanks for the PR. Awaiting your signal before pushing to cassandra-test.

PR #676, opened 2016-11-08
escape literal asterisks in docstring

mambocab:
Addresses [PYTHON-645](https://datastax-oss.atlassian.net/browse/PYTHON-645).

I've tested locally with `python setup.py doc`, which no longer prints these warnings:

```/home/mambocab/cstar_src/python-driver/cassandra/cqlengine/query.py:docstring of cassandra.cqlengine.query.ContextQuery:4: WARNING: Inline emphasis start-string without end-string.
/home/mambocab/cstar_src/python-driver/cassandra/cqlengine/query.py:docstring of cassandra.cqlengine.query.ContextQuery:5: WARNING: Inline strong start-string without end-string.
```

I also viewed the docs locally with `python -m SimpleServer` and they look like I expect:

driver-docs

### DIFF
--- a/cassandra/cqlengine/query.py
+++ b/cassandra/cqlengine/query.py
@@ -274,8 +274,8 @@ class ContextQuery(object):
     A Context manager to allow a Model to switch context easily. Presently, the context only
     specifies a keyspace for model IO.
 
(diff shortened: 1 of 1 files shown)


aboudreault:
Thanks Jim!

PR #677, opened 2016-11-15
PYTHON-649

mambocab:
I don't necessarily expect this to get merged as-is; this is more of a review request. I want to check:

- that others like the `for user_type, conn in product(set(udts), connection._connections.values())` approach
- that the tests in `tests/integration/cqlengine/model/test_udts.py` cover the cases we care about, in terms of ordering UDT definitions and connection creations. I think they have, but I'd like a review and to have another look myself.

### DIFF
--- a/cassandra/cqlengine/connection.py
+++ b/cassandra/cqlengine/connection.py
@@ -222,9 +222,6 @@ def set_session(s):
     This may be relaxed in the future
     """
 
-    if DEFAULT_CONNECTION not in _connections:
-        raise CQLEngineException('Cannot set session on default connection; no default connection set.')
-
     conn = get_connection()
 
     if conn.session:
--- a/cassandra/cqlengine/connection.py
+++ b/cassandra/cqlengine/connection.py
@@ -142,6 +142,31 @@ class Connection(object):
 def register_connection(name, hosts=None, consistency=None, lazy_connect=False,
                         retry_connect=False, cluster_options=None, default=False,
                         session=None):
+    """
+    Add a connection to the connection registry. ``hosts`` and ``session`` are
+    mutually exclusive, and ``consistency``, ``lazy_connect``,
+    ``retry_connect``, and ``cluster_options`` only work with ``hosts``. Using
+    ``hosts`` will create a new :class:`cassandra.cluster.Cluster` and
+    :class:`cassandra.cluster.Session`.
+
+    :param list hosts: list of hosts, (``contact_points`` for :class:`cassandra.cluster.Cluster`).
+    :param int consistency: The default :class:`~.ConsistencyLevel` for the
+        registered connection's new session. Default is the same as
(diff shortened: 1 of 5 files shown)


cassandra/cqlengine/models.py
```diff
@@ -1021,8 +1022,8 @@ def _transform_column(col_name, col_obj):
         for col in column_dict.values():
             columns.resolve_udts(col, udts)
 
-        for user_type in set(udts):
-            user_type.register_for_keyspace(klass._get_keyspace())
+        for user_type, conn in product(set(udts), connection._connections.values()):
+            user_type.register_for_keyspace(klass._get_keyspace(), connection=conn)
```
aboudreault: This would probably work. However, It might be better to keep the 'connection' logic **only** in the management commands. I think It  will become confusing if we start adding some of that logic in different places. What do you think?

mambocab: I see what you mean -- let me have a look at what consolidating the logic would involve.

mambocab:
Pushed 5c649a9; rebasing for a cleaner history.

mambocab:
I've pushed a [new test for behavior that was broken with #656](692c3a8), and [a fix for that](c4f9989). I think there's definitely some code dedupe to be done here before merge.

cassandra/cqlengine/connection.py
```diff
@@ -222,7 +228,14 @@ def set_session(s):
...
 
-    conn = get_connection()
+    try:
+        conn = get_connection()
+    except CQLEngineException:
+        # no default connection set; initalize one
+        default_name = 'default'
+        conn = Connection.from_session(default_name, s)
+        _connections[default_name] = conn
+        set_default_connection[default_name]
```
aboudreault: This doesn't seem to work as expected. I'm getting this exception:
```
    connection.set_session(s)
  File "/home/aboudreault/dev/python-driver/cassandra/cqlengine/connection.py", line 238, in set_session
    set_default_connection[default_name]
TypeError: 'function' object has no attribute '__getitem__'
```
You probably want to do set_default_connection(...).

mambocab:
Fixed the `set_default_connection` call; not sure how I missed that.

Unit tests now pass locally. Do you agree that it's worth the time to refactor before we commit? I don't like the duplication between `register_connection` and `set_session` -- it's only a couple lines, but I think they're likely to change relatively soon.

aboudreault:
+1 

I agree that set_session could be removed in the next major release. register_connection should be able to register a new connection from a session object. I don't have any objection if we can add that support now but we need to keep set_session. You can add  a note about the the removal of it in https://datastax-oss.atlassian.net/browse/PYTHON-504

aboudreault:
Looks good. it might be good to add some documentation about that, with 2 examples, and also specifying what configuration arguments are incompatible with the session param.

aboudreault:
+1

PR #678, opened 2016-11-17
Avoid using of nonexistent prepared statement in ResponseFuture when handle succesful response to prepare a statement request

a-lst:
When we use a response future with a batch statement, the 'prepared_statement' attribute of the response future is not set, but we try to use it under some circumstances.


mambocab:
Thanks for the PR! I've filed [PYTHON-706](https://datastax-oss.atlassian.net/browse/PYTHON-706) for this and scheduled its review for 3.9.

mambocab:
Thanks for the patch, especially for including tests! Unfortunately, I don't see how this can happen outside of tests. You mention something about using batch statements -- do you have a script that can demonstrate this behavior while actually executing CQL? Thanks.

a-lst:
Hi. I think, that CQL script won't be able to help to demonstrate this behaviour. I'll try to describe it using the pseudo-code.

1. cluster = Cluster(contact_points=['192.168.0.2', '192.168.0.3'])
    session = cluster.connect()

After the 1-st step the 'session._pools' dictionary could contain at least 1 host.

2. prepared_stmt = session.prepare(...)

At the 2-nd step we use the 'session._pools' dictionary to prepare some query on all hosts ('cluster.prepare_on_all_hosts' is True by default). But the 'session._pools' dictionary still could contain not all of the hosts.

3. bound_stmt = prepared_stmt.bind(...)
    batch_stmt = BatchStatement(...)
    batch_stmt.add(bound_stmt)
    session.execute(batch_stmt)

At the 3-rd step the '_create_response_future' function of the 'session' instance is used to build instance of 'ResponseFuture'. The 'prepared_statement' attribute of 'ResponseFuture' instance is None for batch statements. The query could be sent to host on which it hasn't been prepared yet(recall the 2-nd step). In this case we get 'PreparedQueryNotFound' response and try to reprepare the statement. At last, we go to the '_execute_after_prepare' function of the 'session' instance and get an error.

The 'wait_for_all_pools' parameter of the 'cluster.connect' function solves the problem, but it's not documented and it's usage is not so obvious.

bjmb:
This test is failing on master now but seems to work fine here.

tests/integration/standard/test_query.py
```diff
@@ -469,6 +466,38 @@ def test_prepare_on_all_hosts(self):
...
+            # is not prepared
+            batch_statement = BatchStatement(consistency_level=ConsistencyLevel.ONE)
+            batch_statement.add(insert_statement, (1, 2))
+            session.execute(batch_statement)
+            select_results = session.execute("SELECT * FROM test3rf.test WHERE k = 1")
+            first_row = select_results[0][:2]
+            self.assertEqual((1, 2), first_row)
+
+        finally:
+            clus.shutdown()
```
mambocab: You can avoid doing this in a `try/finally` by running `self.addCleanup(clus.shutdown)` after constructing `clus`.

tests/integration/standard/test_query.py
```diff
@@ -494,19 +494,72 @@ def test_prepare_batch_statement(self):
...
+        session = clus.connect(wait_for_all_pools=True)
+
+        session.execute("DROP TABLE IF EXISTS %s" % table)
+        session.execute("CREATE TABLE %s (k int PRIMARY KEY, a int, b int, d int)" % table)
+        insert_statement = session.prepare("INSERT INTO %s (k, b, d) VALUES  (?, ?, ?)" % table)
+
+        # Altering the table might trigger an update in the insert metadata
+        session.execute("ALTER TABLE %s ADD c int" % table)
+
+        # We query the three hosts
```
mambocab: Do we have any guarantee that these statements actually hit each of the 3 hosts? I don't think it changes the effectiveness of the test, but this comment might not be accurate.

bjmb: I reckon that the three hosts are queried, one for each batch statement, due to the ForcedHostSwitchPolicy, without this it wouldn't usually happen, I'll add a bit more of explanation

mambocab: Ohhh, got it, totally missed that. Excellent -- go ahead and change the comment and merge then. Thanks!

PR #681, opened 2016-12-13
Timestamp generator

mambocab:
Addresses PYTHON-676. The history isn't great, but I plan to squash before/on merge if there are no objections.

Unit and integration tests pass locally on py2.7. I have not tested this over actual timestamp discontinuities.

### DIFF
--- a/cassandra/timestamps.py
+++ b/cassandra/timestamps.py
@@ -1,0 +1,107 @@
+# Copyright 2013-2016 DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This module contains utilities for generating timestamps for client-side
+timestamp specification.
+"""
+
+import logging
+import time
+from threading import Lock
+
+log = logging.getLogger(__name__)
+
+class MonotonicTimestampGenerator(object):
+    """
+    An object that, when called, returns ``int(time.time() * 1e6)`` when
+    possible, but, if the value returned by ``time.time`` doesn't increase,
+    drifts into the future and logs warnings.
+    Exposed configuration attributes can be configured with arguments to
+    ``__init__`` or by changing attributes on an initialized object.
+
+    .. versionadded:: 3.8.0
+    """
+
+    warn_on_drift = True
+    """
+    If true, log warnings when timestamps drift into the future as allowed by
+    :attr:`warning_threshold` and :attr:`warning_interval`.
+    """
+
+    warning_threshold = 0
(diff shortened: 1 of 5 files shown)


aboudreault:
LGTM Jim. Can you give a try to our benchmark scripts to see how the lock in that generator affects the performance?

cassandra/timestamps.py
```diff
@@ -0,0 +1,107 @@
...
+        Makes ``MonotonicTimestampGenerator`` objects callable; defers
+        internally to _next_timestamp.
+        """
+        with self.lock:
+            return self._next_timestamp(now=int(time.time() * 1e6),
+                                        last=self.last)
+
+    def _maybe_warn(self, now):
+        # should be called from inside the self.lock.
+        diff = now - self._last_warn
```
bjmb: I think this line should be:
diff = self.last - now
as this value may be logged below

cassandra/timestamps.py
```diff
@@ -0,0 +1,107 @@
...
+        since_last_warn = now - self._last_warn
+
+        warn = (self.warn_on_drift and
+                (diff >= self.warning_threshold) and
```
bjmb: This comparison would be:
self.last + 1 - now >= self.warning_threshold
as the comparison should be done with the value that is going to be returned right?

cassandra/timestamps.py
mambocab: Yep, good catch!

cassandra/timestamps.py
mambocab: I think it should be `since_last_warn >= self.warning_threshold` -- we're trying to avoid flooding logs, so we care about how long it's been since the last warning was issued. Sound right?

bjmb: I think that is checked in the next line right?
`since_last_warn >= self.warning_interval`

This lines checks the returned timestamp drifts more than ``warning_threshold`` seconds into the future.

mambocab: Yep -- sorry, I should've spent more time reading for context.

mambocab: Sorry again for my confusion. To address your original concern: I think either behavior (comparing to `self.last - now` or to `self.last + 1 - now`) is reasonable. I had a look at at the Java implementation:

https://github.com/datastax/java-driver/blob/da72105c2fcef8f0865047815df146cfbb2354e6/driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java#L60

https://github.com/datastax/java-driver/blob/da72105c2fcef8f0865047815df146cfbb2354e6/driver-core/src/main/java/com/datastax/driver/core/LoggingMonotonicTimestampGenerator.java#L65

It looks like the behavior there is the same as that in the PR -- so `self.last - now`. Personally, I don't see a good reason to make the implementation work differently than the Java driver's. Do you object to it strongly enough?

bjmb: Yeah, `self.last - now` should be fine

bjmb:
I've changed some things about the seconds / microseconds inside the class `MonotonicTimestampGenerator`. Maybe a review would be worth it.

mambocab:
@bjmb I've added a small fix that tears down the patched `time.time`. I'm +1 to your changes, they look great. Unless you object, I'll squash my initial commits together, squash the last 3 together, and merge.

bjmb:
@mambocab great! Sure, go ahead with the squash and merge.

PR #682, opened 2016-12-19
PYTHON-669: Don't prepare queries on ignored hosts on_up

mambocab:
Adds a test for PYTHON-669 and fixes the behavior it tests for.


GregBestland:
+1LGTM

PR #684, opened 2016-12-28
setup.py: build libev extension on pypy

iksaif:
Enable libev extension on pypy as I don't see any good reason not to do it.
Tested on pypy 5.3.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -166,9 +166,9 @@ The optional C extensions are not supported on big-endian systems.
 
 pypy_unsupported_msg = \
 """
-=================================================================================
-Some optional C extensions are not supported in PyPy. Only murmur3 will be built.
-=================================================================================
+===================================================================================
+Some optional C extensions are not supported in PyPy. murmur3, libev will be built.
+===================================================================================
 """
 
 is_windows = os.name == 'nt'
@@ -186,7 +186,7 @@ elif not is_supported_arch:
 
 try_extensions = "--no-extensions" not in sys.argv and is_supported_platform and is_supported_arch and not os.environ.get('CASS_DRIVER_NO_EXTENSIONS')
 try_murmur3 = try_extensions and "--no-murmur3" not in sys.argv
-try_libev = try_extensions and "--no-libev" not in sys.argv and not is_pypy and not is_windows
+try_libev = try_extensions and "--no-libev" not in sys.argv and not is_windows
 try_cython = try_extensions and "--no-cython" not in sys.argv and not is_pypy and not os.environ.get('CASS_DRIVER_NO_CYTHON')
 try_cython &= 'egg_info' not in sys.argv  # bypass setup_requires for pip egg_info calls, which will never have --install-option"--no-cython" coming fomr pip
 


iksaif:
Looks like it works fine with my pypy but not with Travis's. I guess I'll need to find out with which version exactly it started working.

aholmberg:
> Looks like it works fine with my pypy but not with Travis's.

You're finding out first-hand why we don't build all extensions in all runtimes. It adds another dimension to our already-vast matrix of runtimes and reactor loop tests.

Have you benchmarked this in pypy? I seem to recall that other reactors JIT-compiled in pypy were doing as well or better than the libev extension. Does libev do better in your runtime?

See also #361

iksaif:
Would you have any ready-to-use benchmark tools that I could use ? Else I'll try to see what I can use on my side. I suspect that it will work well with Pypy > 5 but it's hard to guarantee it.

Worst case maybe it's fine to build it even if it crashes sometime as long as it's not the default ?

aholmberg:
> Would you have any ready-to-use benchmark tools that I could use?

Running [./benchmarks/future_full_pipeline.py](https://github.com/datastax/python-driver/blob/master/benchmarks/future_full_pipeline.py) is usually good for some quick comparisons. If you run locally, make sure to run a few times to warm up the server JVM.

> Worst case maybe it's fine to build it even if it crashes sometime as long as it's not the default ?

That's probably not advisable since we're not testing, and can't be assured there are no runtime issues.

iksaif:
```
$ python setup.py build
$ pip install -e .
$ python -V
Python 2.7.10 (5.4.1+dfsg-1, Sep 07 2016, 21:37:41)
[PyPy 5.4.1 with GCC 6.2.0 20160901]
$ python ./benchmarks/future_full_pipeline.py
2017-01-03 16:10:42,110 [INFO] root: Using 'cassandra' package from ['/home/iksaif/dev/python-driver/cassandra']
2017-01-03 16:10:47,598 [INFO] root: ==== AsyncoreConnection ====
2017-01-03 16:10:59,182 [INFO] root: Total time: 5.41s
2017-01-03 16:10:59,183 [INFO] root: Average throughput: 1848.85/sec
2017-01-03 16:10:59,184 [INFO] root: Using 'cassandra' package from ['/home/iksaif/dev/python-driver/cassandra']
2017-01-03 16:11:04,738 [INFO] root: ==== LibevConnection ====
2017-01-03 16:11:13,291 [INFO] root: Total time: 3.05s
2017-01-03 16:11:13,291 [INFO] root: Average throughput: 3273.44/sec
2017-01-03 16:11:13,291 [INFO] root: Using 'cassandra' package from ['/home/iksaif/dev/python-driver/cassandra']
2017-01-03 16:11:19,008 [INFO] root: ==== TwistedConnection ====
2017-01-03 16:11:27,291 [INFO] root: Total time: 2.25s
2017-01-03 16:11:27,291 [INFO] root: Average throughput: 4448.86/sec
```

Looks pretty good, I also didn't expect Twisted to be that good.
I changed tox.ini and setup.py to enable (and test) libev only with pypy5.

iksaif:
(note: the current travis failure is unrelated, support for python2.6 should probably be dropped)

aholmberg:
Yes, thanks. That's a known issue with a dependency. It will be addressed separately.

iksaif:
Since the libev extension seems to work with recent versions of pypy and is more performant, should we consider adding it ?

The twisted implementation is fine, but it's not as easily embedable (I tried and and several issues with my application that is already using twisted, need to spend some time to fix them)

mambocab:
We're not inherently opposed to this, but there are a couple pieces of work that need to be done before we'd add `pypy`+`libev` to the test matrix:

- We'd like to see a more rigorous benchmark demonstrating the benefits of `libev` over `asyncore`. For rigor here, we'd like the following:
  - The benchmark should be run multiple times against the same C\* cluster, throwing away the first run to allow the server JVM to warm up. 3 post-warmup runs should be fine, though more are welcome.
  - The runs should run more iterations (with `-n`) to make sure the JIT kicks in and to offset the cost of compilation.
- We'd like to know what versions of `pypy` work and don't work with `libev`.

If you could provide some benchmark results and let us know what versions of `pypy` we can test and support with `libev`, we can move forward with reviewing this. Thanks!

iksaif:
Ran the benchmark with the following paths

```diff
diff --git a/benchmarks/base.py b/benchmarks/base.py
index 10cce0c..71efcc2 100644
--- a/benchmarks/base.py
+++ b/benchmarks/base.py
@@ -90,6 +90,7 @@ def setup(options):
             session.execute("""
                 CREATE KEYSPACE %s
                 WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': '2' }
+                AND durable_writes = false
                 """ % options.keyspace)
 
             log.debug("Setting keyspace...")
@@ -126,6 +127,19 @@ def teardown(options):
 
 def benchmark(thread_class):
     options, args = parse_options()
+    if options.warmups:
+        log.info("Warmup:")
+    for i in xrange(options.warmups):
+        benchmark_once(thread_class, options, args)
+        log.info("")
+
+    if options.runs > 1:
+        log.info("Benchmark:")
+    for i in xrange(options.runs):
+        benchmark_once(thread_class, options, args)
+        log.info("")
+
+def benchmark_once(thread_class, options, args):
     for conn_class in options.supported_reactors:
         setup(options)
         log.info("==== %s ====" % (conn_class.__name__,))
@@ -238,7 +252,10 @@ def parse_options():
                       help='Specify the column type for the schema (supported: int, text, float, uuid, timestamp)')
     parser.add_option('--read', action='store_true', dest='read', default=False,
                       help='Read mode')
-
+    parser.add_option('-r', '--runs', type='int', default=1,
+                      help='Number of runs [default: %default]')
+    parser.add_option('-w', '--warmups', type='int', default=0,
+                      help='Number of warmup runs [default: %default]')
 
     options, args = parser.parse_args()
 ```

Here is the median trhoughput for each reactor (5 runs after 5 warmups):

reactor | throughput
 --- | --- 
asyncore | 18792.5
libev | 17158.5
twisted | 17616

So you're right, I don't think the reactor actually make much difference.

mambocab:
Cool, thanks for running that. Since we don't have evidence that `libev` provides a major performance benefit, I'm going to close this PR.

If you wind up finding evidence that this is worth maintaining, please let us know. Thanks again for the report and for your benchmarking effort.

PR #685, opened 2017-01-06
Timer race condition

Massi78:
Using the *execute_async()* method, it may happen either *_set_final_result()* or *_set_final_exception()* is called before the *_start_timer()* execution.
Since both *_set_final_result()*  and  *_set_final_exception()* are responsible for cancelling the timer contained in the *ResponseFuture* object, in the case above the started timer won't be cancelled until the timeout condition is triggered even if the result has been already returned

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -117,6 +117,7 @@ DEFAULT_MAX_CONNECTIONS_PER_REMOTE_HOST = 2
 
 
 _NOT_SET = object()
+_INVALID = object()
 
 
 class NoHostAvailable(Exception):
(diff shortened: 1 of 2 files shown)


aholmberg:
Thanks for the PR. This looks related to [PYTHON-644](https://datastax-oss.atlassian.net/browse/PYTHON-644), which we happen to be looking into presently. We are planning to get this in for the next release.

mambocab:
Thanks again for the PR. We took a slightly different approach in #690, so I'm going to close this. Please feel free to make another PR or file a JIRA if the changes there don't address your concerns.

PR #686, opened 2017-01-11
PYTHON-644

mambocab:
This change adds a sentinel to indicate when a `ResponseFuture`'s timer has been cancelled. This is similar to #685, but also uses a lock to eliminate race conditions. For instance, the lock prevents this series of events:

1. timer calls `_on_speculative_execute` on thread 1
2. `_set_final_result` cancels the timer on thread 2
3. `_on_speculative_execute` starts a new timer on thread 1

With these changes, I have not produced memory leaks with the script provided in [PYTHON-644](https://datastax-oss.atlassian.net/browse/PYTHON-644) when running it locally many times.

@aholmberg and I both suspect that there's a better solution than explicit locking, but I haven't figured it out yet.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -117,6 +117,7 @@ DEFAULT_MAX_CONNECTIONS_PER_REMOTE_HOST = 2
 
 
 _NOT_SET = object()
+_CANCELLED = object()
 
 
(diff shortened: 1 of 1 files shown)


mambocab:
Closing this PR for now, while I investigate a lockless solution that Adam suggested.

PR #689, opened 2017-01-16
PYTHON-655 Duration Type

aboudreault:


cassandra/marshal.py
```diff
@@ -84,3 +92,61 @@ def varint_pack(big):
...
+
+        n += 1
+        values.append(decode_zig_zag(val))
+
+    return tuple(values)
+
+
+def vints_pack(values):
+    revbytes = bytearray()
+    values = [int(v) for v in values[::-1]]
```
aholmberg: I know we're in a hurry, so I wouldn't make a science project out of it now, but I'm curious if you tested this reversing algorithm against another that builds the buffer in the right order?

cassandra/util.py
```diff
@@ -1193,3 +1193,48 @@ def _sanitize_identifiers(field_names):
...
+    days = 0
+    nanoseconds = 0
+
+    def __init__(self, months=0, days=0, nanoseconds=0):
+        self.months = months
+        self.days = days
+        self.nanoseconds = nanoseconds
+        self.validate()
+
+    def validate(self):
```
aboudreault: The  server doesn't validate that case. It will rather return a modified Duration with fixed values. Sandeep and I agreed adding this validation on the client could avoid some confusion.
Here is an example of the server behavior without this validation. ie. If you create a Duration(-10,4,2) and fetch it back, you'll get a Duration(-10,-4,-2). 

Let me know if you are fine with it or we can discuss more about that need.

cassandra/util.py
```diff
@@ -1193,3 +1193,48 @@ def _sanitize_identifiers(field_names):
...
+        return self.months > 0 or self.days > 0 or self.nanoseconds > 0
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self.months == other.months and self.days == other.days and self.nanoseconds == other.nanoseconds
+
+    def __repr__(self):
+        return "Duration({0}, {1}, {2})".format(self.months, self.days, self.nanoseconds)
+
+    def __str__(self):
+        return '{0}{1}mo{2}d{3}ns'.format(
```
aboudreault: Fixing the use of %. I just always .format() and forgot % is used in all the package code base.

Also fixed the float/decimal using %d. Only integers are accepted. This will do the same thing than the binary encoding, which cast  to integers.

cassandra/cqltypes.py
```diff
@@ -660,6 +660,24 @@ def serialize(val, protocol_version):
...
+
+    @staticmethod
+    def deserialize(byts, protocol_version):
+        months, days, nanoseconds = vints_unpack(byts)
+        return util.Duration(months, days, nanoseconds)
+
+    @staticmethod
+    def serialize(duration, protocol_version):
+        try:
+            duration.validate()
```
aboudreault: I think it is better to keep it since a user could modify the duration instance manually by accessing its attribute.

cassandra/marshal.py
aboudreault: I suppose you are talking in term of performance?

aholmberg:
+1

PR #690, opened 2017-01-17
PYTHON-644 - Timer logic

mambocab:
First commit reverts part of the fix for [PYTHON-367](https://datastax-oss.atlassian.net/browse/PYTHON-367) (#379) that put `_start_timer` calls in code that could be called from secondary threads. This caused a race between timer cancellation and that `_start_timer` call.

The second commit adds a `_cancel_timer` call before defuncting connections if the `ResponseFuture` isn't going to retry the request. Commit message explaining this reproduced below:

>This allows us to revert some of the fix to PYTHON-367 without the timer
heap growing out of control as described on that ticket. We cancel
before defuncting to ensure that the timer doesn't hang around while
running callbacks. Between this change and the delegation of callbacks
to a secondary thread (see commit 5192302, part of the original
PYTHON-367 fix), this commit should not cause a regression on
PYTHON-367.

>Note that we don't cancel timers before defuncting and retrying -- we're
not done with the logical request and still want that timer to fire.

>Now that timers are started on ResponseFuture initialization, we
actually use the value of timeout in the tests, so we no longer pass
that argument, allowing the test to use the default value.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3608,6 +3608,7 @@ class ResponseFuture(object):
                 # we got some other kind of response message
                 msg = "Got unexpected message: %r" % (response,)
                 exc = ConnectionException(msg, host)
+                self._cancel_timer()
                 self._connection.defunct(exc)
                 self._set_final_exception(exc)
         except Exception as exc:
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3277,6 +3277,7 @@ class ResponseFuture(object):
         self._errbacks = []
         self._spec_execution_plan = speculative_execution_plan or self._spec_execution_plan
         self.attempted_hosts = []
+        self._start_timer()
 
     def _start_timer(self):
         if self._timer is None:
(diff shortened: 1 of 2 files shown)


mambocab:
Travis tests on my fork all passed, except the expected Python 2.6 errors.

aboudreault:
LGTM Jim! I tested and confirm it fixes the issue. (Tested with libev and asyncore)

cassandra/cluster.py
```diff
@@ -3612,6 +3608,7 @@ def _set_result(self, host, connection, pool, response):
                 # we got some other kind of response message
                 msg = "Got unexpected message: %r" % (response,)
                 exc = ConnectionException(msg, host)
+                self._cancel_timer()
```
bjmb: I was wondering if this line is really necessary,  as the timer will be canceled inside inside
`self._set_final_exception(exc)`

mambocab: The goal is to cancel as quickly as possible -- without waiting for callbacks to execute in `defunct` below -- to avoid a pileup of timers when clients don't back off requests on a failing connection as described in [PYTHON-367](https://datastax-oss.atlassian.net/browse/PYTHON-367). Do you think this accomplishes that goal?

bjmb: Yeah, makes sense

PR #693, opened 2017-01-19
add shuffle_replicas config option

mambocab:
See [PYTHON-643](https://datastax-oss.atlassian.net/browse/PYTHON-643).

### DIFF
--- a/tests/integration/long/test_loadbalancingpolicies.py
+++ b/tests/integration/long/test_loadbalancingpolicies.py
@@ -533,25 +533,7 @@ class LoadBalancingPolicyTests(unittest.TestCase):
         keyspace = 'test_token_aware_with_rf_2'
         cluster, session = self._set_up_shuffle_test(keyspace, replication_factor=2)
 
-        LIMIT_TRIES = 20
-        previous_query_count_two, previous_query_count_three = None, None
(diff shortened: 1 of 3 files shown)


aboudreault:
LGTM. simple and clean.

tests/integration/long/test_loadbalancingpolicies.py
```diff
@@ -520,6 +520,89 @@ def test_token_aware_with_local_table(self):
 
         cluster.shutdown()
 
+    def test_token_aware_with_shuffle_rf2(self):
+        """
+        Test to validate the hosts are shuffled when the `shuffle_replicas` is truthy
```
mambocab: I see that this test asserts that the correct number of queries are made, but I don't see how it tests that they're shuffled. It's still a valuable test, but either this docstring is wrong or I don't understand the test.

I'm actually not sure how to test "replicas are shuffled". Maybe we should also add a regression test for PYTHON-643 -- test "if you make enough queries, some query will be issued to replicas in a different order than the first query was".

bjmb: You're right, is not that straightforward to check randomness is happening. I updated the tests following your suggestion.

PR #694, opened 2017-01-20
PYTHON-692, Allow hosts=None for register_connection

mambocab:
Turns out that allowing `hosts=None` is not only ok, but required for `connection.setup(hosts=None, ...)` and `connection.default` to continue working. This adds integration tests for those functions and fixes the regression.

I think maybe these integration tests could be unit tests, so don't merge just yet.


mambocab:
> I think maybe these integration tests could be unit tests, so don't merge just yet.

Never mind -- these tests connect to the server, and I don't know if we want to mock them out too heavily.

aboudreault:
+1

PR #696, opened 2017-01-24
let users specify Cython version

mambocab:
Adds a `CASS_DRIVER_ALLOWED_CYTHON_VERSION` "escape hatch" env var that allows users to specify what version of Cython to use.

You can verify this behavior by installing the driver in an environment without Cython installed, then opening the installed driver with

```
view `python -c 'from cassandra import cluster ; print cluster.__file__'`
```

and searching for `cython`. If you install without `CASS_DRIVER_ALLOWED_CYTHON_VERSION`, the driver is installed with `0.24.1`. If you install with `CASS_DRIVER_ALLOWED_CYTHON_VERSION=0.25.2`, the driver is installed with `0.25.2`. (This holds even when you've installed `Cython==0.25.2` in the environment you're installing the driver.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -388,7 +388,11 @@ def run_setup(extensions):
         # 1.) build_ext eats errors at compile time, letting the install complete while producing useful feedback
         # 2.) there could be a case where the python environment has cython installed but the system doesn't have build tools
         if pre_build_check():
-            kw['setup_requires'] = ['Cython>=0.20,<0.25']
+            cython_dep = 'Cython>=0.20,<0.25'
+            user_specified_cython_version = os.environ.get('CASS_DRIVER_ALLOWED_CYTHON_VERSION')
+            if user_specified_cython_version is not None:
+                cython_dep = 'Cython==%s' % (user_specified_cython_version,)
+            kw['setup_requires'] = [cython_dep]
         else:
             sys.stderr.write("Bypassing Cython setup requirement\n")
 


setup.py
```diff
@@ -388,7 +388,10 @@ def run_setup(extensions):
         # 1.) build_ext eats errors at compile time, letting the install complete while producing useful feedback
         # 2.) there could be a case where the python environment has cython installed but the system doesn't have build tools
         if pre_build_check():
-            kw['setup_requires'] = ['Cython>=0.20,<0.25']
+            user_specified_cython_version = os.environ.get('CASS_DRIVER_ALLOWED_CYTHON_VERSION')
+            kw['setup_requires'] = ['Cython>=0.20,<0.25' if
+                                    user_specified_cython_version is None else
+                                    'Cython=={}'.format(user_specified_cython_version)]
```
aboudreault: for clarity if we need to add more deps in the future. I would do something like: 
```python
cython_dep = 'Cython>=0.20,<0.25'
user_specified_cython_version = os.environ.get('CASS_DRIVER_ALLOWED_CYTHON_VERSION')
if user_specified_cython_version:
    cython_dep =  'Cython==%s' % user_specified_cython_version

kw['setup_requires'] = [cython_dep]
```

Note that In the whole code base, we use the old style formatting ('%s' % 'hello'), so better to stick to that here as well.

mambocab:
Thanks for the review! I've addressed your comments; if they look good to you I'll squash and merge.

aboudreault:
+1

mambocab:
Thanks, merging :+1:

PR #700, opened 2017-02-07
fix ExponentialReconnectionPolicy may throw OverflowError problem

vipjml:
self.base_delay * (2 ** i) may throw OverflowError  when self.base_delay is float and i>2000

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -556,8 +556,13 @@ class ExponentialReconnectionPolicy(ReconnectionPolicy):
 
     def new_schedule(self):
         i = 0
+        delay = self.base_delay
         while self.max_attempts is None or i < self.max_attempts:
-            yield min(self.base_delay * (2 ** i), self.max_delay)
+            if delay > self.max_delay:
+                yield self.max_delay
+            else: 
+                yield delay
+                delay = delay * 2
             i += 1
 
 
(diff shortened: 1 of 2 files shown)


cassandra/policies.py
```diff
@@ -556,8 +556,13 @@ def __init__(self, base_delay, max_delay, max_attempts=64):
 
     def new_schedule(self):
         i = 0
+        delay = self.base_delay
         while self.max_attempts is None or i < self.max_attempts:
-            yield min(self.base_delay * (2 ** i), self.max_delay)
```
vipjml: self.base_delay * (2 ** i) may throw OverflowError when self.base_delay is float and i>2000

mambocab:
Thanks for the PR! I've filed [PYTHON-707](https://datastax-oss.atlassian.net/browse/PYTHON-707) for this and scheduled its review for 3.9.

cassandra/policies.py
mambocab: I believe your fix still can lead to `OverFlowError`s when `base_delay` is a float and `max_delay` is larger `sys.float_info.max`. I'll push a fix without this issue and tag you in the PR.

mambocab:
Superceded by 744 for the reason mentioned [here](https://github.com/datastax/python-driver/pull/700#discussion_r112958129):

> I believe your fix still can lead to `OverFlowError`s when `base_delay` is a float and `max_delay` is larger `sys.float_info.max`. I'll push a fix without this issue and tag you in the PR.

cassandra/policies.py
vipjml: i agree

PR #703, opened 2017-02-15
Fix RuntimeError caused by change dict size during iteration

vetal4444:
> ERROR:cassandra.cluster:Error preparing query on all hosts: 
  File "cassandra/cluster.py", line 2230, in cassandra.cluster.Session.prepare (cassandra/cluster.c:38698) 
  File "cassandra/cluster.py", line 2242, in cassandra.cluster.Session.prepare_on_all_hosts (cassandra/cluster.c:39022) 
RuntimeError: dictionary changed size during iteration

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2242,7 +2242,7 @@ class Session(object):
         Intended for internal use only.
         """
         futures = []
-        for host in self._pools.keys():
+        for host in tuple(self._pools.keys()):
             if host != excluded_host and host.is_up:
                 future = ResponseFuture(self, PrepareMessage(query=query), None, self.default_timeout)
 


mambocab:
Thanks for the PR! I've filed [PYTHON-708](https://datastax-oss.atlassian.net/browse/PYTHON-708) for this and scheduled its review for 3.9.

aboudreault:
Thanks @vetal4444 !

bjmb:
I wonder @aboudreault if we should do the same in other parts of cluster.py where the list is iterated such as [here](https://github.com/datastax/python-driver/blob/89aad9bf36f659d807af3f05a838811923a5bbd5/cassandra/cluster.py#L2458), [here](https://github.com/datastax/python-driver/blob/89aad9bf36f659d807af3f05a838811923a5bbd5/cassandra/cluster.py#L2499) and [here](https://github.com/datastax/python-driver/blob/89aad9bf36f659d807af3f05a838811923a5bbd5/cassandra/cluster.py#L1568), I'd say theorically it could happen in this places as well.

aboudreault:
@bjmb I'm +1 with your suggestion if there are chances the same issue happen.

bjmb:
Added [this](https://github.com/datastax/python-driver/commit/b044b34e809a975d72014a903dfce32e13962e0b) commit as part of this.

PR #707, opened 2017-02-28
add total_ordering to Time, Date, & Host

mambocab:
[PYTHON-714](https://datastax-oss.atlassian.net/browse/PYTHON-714)

This works because the decorated types implement `__eq__` and `__lt__` already.

We don't have to do the same for metadata.Token because it implements `__cmp__`


aboudreault:
LGTM!

PR #709, opened 2017-03-09
update six dependency

aholmberg:
table extensions use viewkeys


mambocab:
My mistake for ignoring this for so long. Integration tests pass, except for an expected failure. Merging.

PR #711, opened 2017-03-13
Added fix for gevent and eventlet

mambocab:
Addresses a bug (found by Jaume) in how the `EventletConnection` handles empty strings coming back from its socket.


bjmb:
+1

PR #712, opened 2017-03-14
Fixed docs import BatchQuery

ittus:


bjmb:
Thank you @ittus for the fix.

PR #715, opened 2017-03-14
Support for testing with a remote cluster

bjmb:


tests/__init__.py
```diff
@@ -41,3 +46,5 @@ def is_gevent_monkey_patched():
 
 def is_monkey_patched():
     return is_gevent_monkey_patched() or is_eventlet_monkey_patched()
+
+notwindows = unittest.skipUnless(not "Windows" in platform.system(), "This test is not adecuate for windows")
```
mambocab: `s/adecuate/adequate`

bjmb: +1 Thanks for the review, I'll merge so it can be included in 3.8.1

PR #716, opened 2017-03-15
PYTHON-673: Ensure Evenlet greenthreads exit gracefully

aboudreault:
This fixes the socket issue with eventlet. Apparently, there are some cleanup not properly done when the greenthread don't exist gracefully. This is an issue with evenlet+python2 only. It doesn't happen with python 3.

### DIFF
--- a/cassandra/io/eventletreactor.py
+++ b/cassandra/io/eventletreactor.py
@@ -19,6 +19,7 @@
 import eventlet
 from eventlet.green import socket
 from eventlet.queue import Queue
+from greenlet import GreenletExit
 import logging
 from threading import Event
 import time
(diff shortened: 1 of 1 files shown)


mambocab:
I'll finish reproducing this locally tomorrow, but in the meantime, I was wondering: do we know whether or not this can happen in the `gevent` reactor? It also uses greenlets, if I understand correctly, and I don't see this handled in that reactor.

aboudreault:
I confirm there is no issue with gevent.

On Mar 15, 2017 5:15 PM, "Jim Witschey" <notifications@github.com> wrote:

I'll finish reproducing this locally tomorrow, but in the meantime, I was
wondering: do we know whether or not this can happen in the gevent reactor?
It also uses greenlets, if I understand correctly, and I don't see this
handled in that reactor.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/datastax/python-driver/pull/716#issuecomment-286881691>,
or mute the thread

mambocab:
Looks great, works locally, I'm +1.

PR #717, opened 2017-03-20
PYTHON-665: make cluster.connect() fails when no connection succeed

aboudreault:


mambocab:
Ugh, some unit test failures -- which means it worked! but also, we should figure out how we want to fix that.

One option could be to expose a "private" configuration variable that allows you to create sessions using the old behavior. What do you think of that, @aboudreault?

aboudreault:
@mambocab I think the current behavior is correct unless I overlooked some special cases. That said, your 'private' configuration would disappeared in 4.0. @bjmb If you have some free cycles, can you check what are the failing tests? I'm wondering if this shouldn't be pushed to 4.0 ...

bjmb:
I'll take a look into this, also why it hasn't run on Jenkins, I think it has to do with 3.10 being in the build.yaml

mambocab:
I can see an argument that we should leave this for 4.0 -- I can't imagine an application it'd break, but it is a behavior change.

The unit test failures are all (as far as I can tell) expected failures -- some tests expect to be able to initialize `Session` objects without an available C\* instance. Whether we push off to 4.0 or not, we'll need to figure out how to test the behavior in those tests without C\* available.

cassandra/cluster.py
```diff
@@ -1975,6 +1975,12 @@ def __init__(self, cluster, hosts, keyspace=None):
         while futures.not_done and not any(f.result() for f in futures.done):
             futures = wait_futures(futures.not_done, return_when=FIRST_COMPLETED)
 
+        if not any(f.result() for f in self._initial_connect_futures):
```
mambocab: Are there use cases for passing no hosts? If so, maybe this should be `if (not self._inital_connect_futures) and (not any(...)):`. I think that'd also make the tests require fewer changes, since empty `hosts` lists would be allowed.

aboudreault: I also questioned myself as well about that. But.. I don't see any good reason a user would try to connect to a cluster without any host. So, there is no real reason .connect() should succeed IMO.

mambocab:
@bjmb I've made a couple small cleanup changes. I want to see the unit tests pass on CI. Let me know what `build.yaml` should look like and I can clean up the commit history and merge.

PR #719, opened 2017-03-20
Fixed some failing tests in Jenkins

bjmb:


build.yaml
```diff
@@ -14,7 +14,7 @@ cassandra:
   - 2.1
   - 2.2
   - 3.0
-  - 3.10
+  - 3.7
```
mambocab: Sorry I didn't ask further when you mentioned this earlier -- why not test against the latest version?

bjmb: This was giving some errors so I'm doing that in a separate branch: https://github.com/datastax/python-driver/tree/python-update-buildyaml

bjmb: Also it wasn't testing against 3.10 despite at some point changing 3.7 -> 3.10, don't know why Job Creator kept deploying against 3.7

tolbertam: You might need to put `3.10` in quotes (i.e. `'3.10'`) since the yaml parser may interpret it as a number and in that case `3.10` == `3.1`

tolbertam: Only reason I knew that was because I had also ran into this recently: https://github.com/datastax/java-driver/blob/3.1.x/build.yaml#L39-L44

tests/integration/standard/test_metrics.py
```diff
@@ -143,6 +143,8 @@ def test_unavailable(self):
         self.assertTrue(results)
 
         # Stop node gracefully
+        # Sometimes this commands continues with the other nodes having not noticed
+        # 1 is down, and a Timeout error is returned instead of an Unavailable
```
mambocab: Is this just a note for future triage?

Also, if `stop(wait=True, wait_other_notice=True)` returns without others noticing, that sounds like a CCM bug?

bjmb: Yeah, I'm not sure if it's something with the environment or something with CCM, I'll have to look a bit more into that, although I don't think it's something we should worry about.

build.yaml
bjmb: That could really be what was happening, now we're trying to run against 3.11, nevertheless I've done the same as in your link and will pass as strings all the versions.

PR #722, opened 2017-03-24
Allow min_length==0 for Column.Text fields

HaraldNordgren:
Current behaviour treats 0 as not specified at all. If this is not on purpose we should check for `None` explicitly instead to allow us to store empty strings in the database.

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -332,7 +332,7 @@ class Text(Column):
         :param int max_length: Sets the maximum length of this string, for validation purposes.
         """
         self.min_length = (
-            1 if not min_length and kwargs.get('required', False)
+            1 if min_length is None and kwargs.get('required', False)
             else min_length)
         self.max_length = max_length
 


mambocab:
Thanks for the PR! I'd like to add a test or two, but after that, I can merge after you've signed the CLA. I've created [a JIRA ticket](https://datastax-oss.atlassian.net/browse/PYTHON-735) and marked it for inclusion in 3.9, our next release.

mambocab:
Sorry, I've jumped the gun in my reply. We've had some discussion about how to address this and are discussing other possible APIs for this solution. We may not get to this before our next release.

Still, please sign the CLA so if we decide to implement this solution, we can merge quickly. Thanks!

HaraldNordgren:
Could someone merge or close this one?

cassandra/cqlengine/columns.py
```diff
@@ -332,7 +332,7 @@ def __init__(self, min_length=None, max_length=None, **kwargs):
         :param int max_length: Sets the maximum length of this string, for validation purposes.
         """
         self.min_length = (
-            1 if not min_length and kwargs.get('required', False)
+            1 if min_length is not None and kwargs.get('required', False)
```
bjmb: I think it should be `1 if min_length is None and kwargs.get('required', False)`

HaraldNordgren:
@bjmb Yes, I think your comment probably makes sense.

I haven't worked with this code in a long while, but looking at it now I'm guessing that was a simple mistake on my part.

HaraldNordgren:
@mambocab I like tests, but I'm not quite sure where to begin with this one.

Setting `self.min_length = 10000` and running `tox` proves to me that this logic is already untested as it succeeds without any problems.

aboudreault:
Thanks. Added a minimal value test here: https://github.com/datastax/python-driver/commit/32037f35a299b937ca71bc553bf8c71ba22d7935

PR #724, opened 2017-03-29
Allow for counter columns be static. Fixed in C* 2.0.7

ubombi:
[CASSANDRA-6827](https://issues.apache.org/jira/browse/CASSANDRA-6827) issue was resolver.
So we can enable static counters in ORM.

Tested manually on existing project.

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -485,12 +485,14 @@ class Counter(Integer):
 
     def __init__(self,
                  index=False,
+                 static=False,
                  db_field=None,
                  required=False):
         super(Counter, self).__init__(
             primary_key=False,
             partition_key=False,
             index=index,
+            static=static,
             db_field=db_field,
             default=0,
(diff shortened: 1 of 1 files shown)


mambocab:
I mentioned it on IRC, but this should be a workaround for now: 

```python
from cassandra.cqlengine import columns
from cassandra.cqlengine.models import Model

class MyModel(Model):
    ...  # define PK, etc
    static_counter = columns.Counter()
    # The `static` attribute already exists, inherited from columns.Column
    static_counter.static = True
```

Please let me know if this doesn't work.

As-is, this PR adds an argument to `columns.Counter.__init__`, and it's not a keyword argument at end of the args, so it's a breaking API change that needs to wait until a major revision.

mambocab:
I'm closing this PR. We will probably address it for the next major ([PYTHON-926](https://datastax-oss.atlassian.net/browse/PYTHON-926)), and we'll credit you in the Git history then.

PR #726, opened 2017-03-30
PYTHON-657 PYTHON-719 rework and fix state tracking issues

aboudreault:


tests/integration/cqlengine/model/test_updates.py
```diff
@@ -134,3 +134,162 @@ def test_primary_key_update_failure(self):
...
+        @since 3.9
+        @jira_ticket PYTHON-657
+        @expected_result column value should not change
+
+        @test_category object_mapper
+        """
+        initial = ModelWithDefault(id=1, mf={0: 0}, dummy=0)
+        initial.save()
+
+        self._assert_model_with_default_row(ModelWithDefault.objects.all().get(),
```
mambocab: I think these calls would be clearer if they were replaced with, e.g.,

```
self.assertEqual(ModelWithDefault.objects().all().get()._as_dict(),
                 {'id': 1, 'dummy': 0, 'mf': {0:1})
```
I think that should work -- point is, there's an existing method for getting a basic data structure representing the returned model.

tests/integration/cqlengine/model/test_updates.py
```diff
@@ -134,3 +134,162 @@ def test_primary_key_update_failure(self):
...
+        initial.update()
+
+        self._assert_model_with_default_row(ModelWithDefault.objects.all().get(),
+                                            [1, 42, {0: 0}])
+
+
+    def test_null_update_is_respected(self):
+        """
+        Check if the we try to update with None under particular
+        circumstances, ir works correctly
```
mambocab: `s/ ir / it /`

bjmb:
Thanks for the review! I'l merge it now.

tests/integration/cqlengine/query/test_updates.py
```diff
@@ -242,6 +242,34 @@ def test_map_update_none_deletes_key(self):
...
+        partition = uuid4()
+        cluster = 1
+
+        TestQueryUpdateModel.objects.create(
+            partition=partition, cluster=cluster)
+
+        obj = TestQueryUpdateModel.objects(
+            partition=partition, cluster=cluster).first()
+
+        self.assertFalse(any([obj._values[column].deleted for column in obj._values]))
```
mambocab: This would have better error reporting as `self.assertFalse({k: v for (k, v) in obj._values.items() if not v.deleted})` -- the failure message will show the key-value pair that was deleted.

PR #727, opened 2017-04-03
PYTHON-688: Add map item deletion in cqlengine

aboudreault:


mambocab:
I've added a couple tests and I'm +1 :+1:

PR #728, opened 2017-04-04
733: race condition between start_fetching_next_page and add_callback

mambocab:
PYTHON-733

Adam figured this one out -- if I understand correctly, there's a race between `add_callback`, which uses `_final_result` to determine if callbacks should be called or not, and `start_fetching_next_page`, which resets it.

Adam was able to deterministically trigger the hang by inserting a sleep here:

https://gist.github.com/bjmb/41999db20794dfbac91bff6264f032ff#file-poc_event-py-L54

which led to `start_fetching_next_page` being called before `add_callback`. With this modification, the script hangs consistently before the changes in this PR and successfully runs after it.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3881,6 +3881,9 @@ class ResponseFuture(object):
         """
         run_now = False
         with self._callback_lock:
+            # Always add fn to self._callbacks, even when we're about to
+            # execute it, to prevent races with functions like
+            # start_fetching_next_page that reset _final_result
             self._callbacks.append((fn, args, kwargs))
             if self._final_result is not _NOT_SET:
                 run_now = True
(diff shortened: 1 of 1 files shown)


bjmb:
But I think the callback could be run twice now, for example:
Thread1 --Cluster._set_final_result --> self._final_result = response (if the thread changed just after that and) Thread2 Cluster.add_callback and here `self._final_result is not _NOT_SET` would be True and it would be run in Thread2. Then when it went back to Thread1 it would also be run in Thread1.

I think this could be solved by saving the callbacks at this point (inside the lock): https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L3730 and then running them.

mambocab:
Good catch, @bjmb -- your proposed fix

bjmb:
Thanks @mambocab and @aholmberg , LGTM, merging and moving to testing.

PR #731, opened 2017-04-05
Python update unit tests

bjmb:


tests/unit/io/test_eventletreactor.py
```diff
@@ -18,50 +18,35 @@
...
-class EventletTimerTest(unittest.TestCase):
-
+class EventletTimerTest(unittest.TestCase, TimerConnectionTests):
+    # There are some issues with some versions of pypy and eventlet
+    @notpypy
     def setUp(self):
-        if EventletConnection is None:
-            raise unittest.SkipTest("Eventlet libraries not available")
-        if not is_eventlet_monkey_patched():
+        if not MONKEY_PATCH_LOOP:
```
bjmb: Yeah, I'll change this, but I'd rather leave the `Eventlet is None` comparison in case Eventlet is not installed

mambocab: Yup, perfect

tox.ini
```diff
@@ -7,13 +7,41 @@ deps = nose
...
+
+[testenv:py27]
+deps = {[base]deps}
+       gevent
+
+setenv = LIBEV_EMBED=0
+         CARES_EMBED=0
+         MONKEY_PATCH_LOOP=1
+changedir = {envtmpdir}
+commands =
```
bjmb: It'd be nice to find a way around this code duplication, the docs from tox seem to do something similar though: http://tox.readthedocs.io/en/latest/example/unittest.html

appveyor/run_test.ps1
```diff
@@ -12,6 +12,11 @@ python -c "import platform; print(platform.architecture())"
 
 $wc = New-Object 'System.Net.WebClient'
 nosetests -s -v --with-ignore-docstrings --with-xunit --xunit-file=unit_results.xml .\tests\unit
+
+$env:MONKEY_PATCH_LOOP=1
```
mambocab: Is it possible to run a single command with an environment variable set? I'm concerned that I might come along in the future and add test commands below this without realizing they're running with `MONKEY_PATCH_LOOP` set.

tests/unit/io/test_eventletreactor.py
```diff
@@ -18,50 +18,32 @@
...
+from eventlet import monkey_patch, kill
 
 try:
     from cassandra.io.eventletreactor import EventletConnection
 except ImportError:
     EventletConnection = None  # noqa
 
-
-class EventletTimerTest(unittest.TestCase):
+@unittest.skipUnless(EventletConnection is not None, "Skpping the eventlet tests because it's not installed")
```
mambocab: Small readability nit: Could you change this to `@unittest.skipIf(EventletConnection is None, ...`? It's just a little simpler to reason about without the double negation.

tests/unit/io/test_eventletreactor.py
```diff
@@ -18,50 +18,32 @@
...
+class EventletTimerTest(unittest.TestCase, TimerConnectionTests):
 
     def setUp(self):
-        if EventletConnection is None:
-            raise unittest.SkipTest("Eventlet libraries not available")
-        if not is_eventlet_monkey_patched():
-            raise unittest.SkipTest("Can't test eventlet without monkey patching")
+        self.connection_class = EventletConnection
+        # We only to patch the time module
+        monkey_patch(time=True)
```
mambocab: Could you explain in the comment above why we only monkeypatch `time`? It's not clear to me why we don't monkeypatch everything.

mambocab: Also, I mentioned this in my earlier review -- does all of this need to happen in each test method? Seems to me that we could monkeypatch/tear down in `setUpClass`/`tearDownClass`.

tox.ini
mambocab: I'm ok with the dupe -- not a big deal.

tests/unit/io/utils.py
```diff
@@ -103,3 +108,34 @@ def submit_and_wait_for_completion(unit_test, connection, start, end, increment,
...
+        callback = TimerCallback(timeout)
+        timer = self.connection_class.create_timer(timeout, callback.invoke)
+        timer.cancel()
+        # Release context allow for timer thread to run.
+        time.sleep(.2)
+        timer_manager = self.connection_class._timers
+        # Assert that the cancellation was honored
+        self.assertFalse(timer_manager._queue)
+        self.assertFalse(timer_manager._new_timers)
+        self.assertFalse(callback.was_invoked())
```
mambocab: If `utils` gets big in the future it'd be nice to move it to a `mixins` module. No action needed right now, but just something to keep an eye on.

mambocab:
Oh, also: ran `tox` locally, works great!

bjmb:
Good! About why we don't run everything in a single regular test job the reason is because overall the environment is not restored completely. In the case of gevent, the rest of the tests can finish although for example sys.module is still filled with patched stuff. In the case of eventlet one of the posterior tests using Event got hanged because it was patched as well. Overall I think is a good idea to run the tests with environment the driver is going to be used and the only way I know now of how to restore it completely is by starting another process.
This makes me think what is the point of having an unpatch function, as I would say the outcome is a bit undefined as the unpatch in not 100%.

appveyor/run_test.ps1
bjmb: A fast solution to this could be to unset MONKEY_PATCH_LOOP after those two command are run

tests/unit/io/test_eventletreactor.py
bjmb: +1 on this

tests/unit/io/test_eventletreactor.py
bjmb: Yeah, I think we should patch everything now, as that's the way the driver is going to be used. I did it when I was hoping I'd be able to unpatch it.
Sorry about that, the reason is in SetUp instead of SetUpClass is because when a class is skipped SetUpClass and tearDownClass are still executed, meaning the patching would occur even if MONKEY_PATCH_LOOP=0

mambocab: Ohhh, got it -- I'm a little surprised, but gtk. Could you mention that in a comment?

mambocab:
> This makes me think what is the point of having an unpatch function, as I would say the outcome is a bit undefined as the unpatch in not 100%.

Gotcha -- weird. I guess that's what tools like `tox` are for. Makes sense.

tests/unit/io/test_eventletreactor.py
bjmb: Sure, I've pushed a slightly different version using setUpClass, also I've removed the tearDown as I'd rather it is explicit that the unpatch hasn't happened. Please let me know if you see any issues with this.

tests/__init__.py
```diff
@@ -44,7 +45,22 @@ def is_gevent_monkey_patched():
...
+def is_eventlet_time_monkey_patched():
+    import eventlet
+    return eventlet.patcher.is_monkey_patched('time')
+
+
 def is_monkey_patched():
     return is_gevent_monkey_patched() or is_eventlet_monkey_patched()
 
+
+MONKEY_PATCH_LOOP = bool(os.getenv('MONKEY_PATCH_LOOP', False) == "1")
```
mambocab: One last nit: it looks like other env vars used as flags just check for its presence in the environment, not its value, e.g. `USE_CASS_EXTERNAL = bool(os.getenv('USE_CASS_EXTERNAL', False))` -- so, this should just do `bool(os.getenv('MONKEY_PATCH_LOOP', False))`

PR #732, opened 2017-04-06
python-705: raise error on primary key update only if its value changed

aboudreault:


mambocab:
Looks like this break the integration tesstsss, but in an expected way. +1 once that's fixed, looks great

PR #737, opened 2017-04-10
Updated tox.ini and build.yaml for testing with python 3.6

bjmb:


tests/unit/cython/utils.py
```diff
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 from cassandra.cython_deps import HAVE_CYTHON, HAVE_NUMPY
-from tests.integration import VERIFY_CYTHON
+try:
+    from tests.integration import VERIFY_CYTHON
+except ImportError:
+    VERIFY_CYTHON = False
```
coldeasy: Does this change imply the cython optimisations are not available for python3?

mambocab: https://github.com/datastax/python-driver/blob/25e6cf3d0a503149481eb9d7bd2d2084526f62c9/tests/integration/__init__.py#L116-L117

PR #739, opened 2017-04-15
PEP 289 (just 1 line)

ubombi:


mambocab:
Good catch on the generator expression. Does using `not any(<generator>)` over `all(<generator>)` gain us anything? `not any` and `all` both should short-circuit on finding an unchanged value, so I think `not any` sacrifices clarity for no benefit. Am I missing something?

If not, could you please change `not any` to `all`? After that, I'll run the integration tests and merge.

ubombi:
Ok, removed.
You didn't miss anything.

mambocab:
Integration tests passed, merging -- thanks!

PR #742, opened 2017-04-20
Removed C* version from .travis.yml, update tox.ini

bjmb:


bjmb:
@mambocab, the reason tox-travis is used instead of tox is because the [travis image still doesn't have py3.5 and py3.6](https://github.com/travis-ci/travis-ci/issues/4794).

PR #743, opened 2017-04-24
PYTHON-730 Resulset.get_query_trace returns empty trace sometimes

aboudreault:
Make sure all trace mutations are written before considering a trace as complete.


mambocab:
Could you add a comment noting why we check those? After that, though, I'm +1.

PR #744, opened 2017-04-24
fixes for ExponentialReconnectionPolicy

mambocab:
This does a couple things.

- It changes the tests for `ExponentialReconnectionPolicy` to use `float`s instead of `int`s. This is the documented type to use. This also triggers the failures seen in #700.
- It changes the logic in `ExponentialReconnectionPolicy` to handle cases like those in the tests where `float`s can overflow.

Would love to hear from @vipjml (person who filed #700) if you have a moment to have a look and confirm this also addresses the issue.

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -564,9 +564,17 @@ class ExponentialReconnectionPolicy(ReconnectionPolicy):
         self.max_attempts = max_attempts
 
     def new_schedule(self):
-        i = 0
+        i, overflowed = 0, False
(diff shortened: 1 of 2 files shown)


vipjml:
this is well.  I think this fix the problem.

aboudreault:
LGTM guys! Thanks

mambocab:
The integration tests ran when we were having some configuration problems in CI. I'm going to rerun them before merging, just to be safe.

PR #746, opened 2017-04-26
PYTHON-196: Raise TraceUnavailable if the future is not done

aboudreault:


mambocab:
This looks good to me. @ptnapoleon could I get a quick once-over to make sure it's what you meant?

ptnapoleon:
While this is no longer fresh in my mind, this does seem to resolve the issue I was having.

PR #749, opened 2017-04-28
Add python3.6 classifier in setup.py

coldeasy:
Looks like this was tested here PYTHON-737


mambocab:
Sorry for the delay in review, but looks great. Thanks! Merging.

PR #750, opened 2017-04-28
Appveyor test improve

bjmb:


mambocab:
LGTM. I've filed [PYTHON-783](https://datastax-oss.atlassian.net/browse/PYTHON-783) to track changing the skipped test a bit so the underlying behavior is tested on Windows as well, but I think that's low-priority.

bjmb:
Thanks for the review! I'll merge this then.

PR #752, opened 2017-05-02
PYTHON-743: UDT values are not updated correctly in CQLEngine

aboudreault:
This Pull Request contains multiple bug fixes related to UDT:

1. It removes the default value handling in Column.to_database. From what I've seen, our Model code base and all sub-column types do not need that handling at this level, in fact all other columns that overload to_database() don't deal with the default value at all. I noticed that bug when debugging a test... _as_dict() was returning the default value when I had explicitly deleted the value before.

2. It fixes the case that a UDT default value wasn't deeply copied when set to a model, which results that multiple model could modify the UDT values.
```python
class ModelWithDefaultCollection(Model):
    id = columns.Integer(primary_key=True)
    mf = columns.Map(columns.Integer, columns.Integer, default={2:2})
    dummy = columns.Integer(default=42)
    udt = columns.UserDefinedType(UDT)
    udt_default = columns.UserDefinedType(UDT, default=UDT(age=1, mf={2: 2}))
...

item = ModelWithDefaultCollection(id=1, dummy=0).create()
item.udt_default.mf = {42: 42}
# ModelWithDefaultCollection.udt_default.column.default is now {42: 42}
```

3. It fixes the *changed* function with a default_value

4. It fixes and improve the test_collection_with_default test.

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -948,7 +948,16 @@ class Map(BaseContainerColumn):
 class UDTValueManager(BaseValueManager):
     @property
     def changed(self):
-        return self.value != self.previous_value or (self.value is not None and self.value.has_changed_fields())
+        if self.explicit:
+            return self.value != self.previous_value
+
+        default_value = self.column.get_default()
+        if not self.column._val_is_null(default_value):
+            return self.value != default_value
+        elif self.previous_value is None:
+            return not self.column._val_is_null(self.value) and self.value.has_changed_fields()
+
+        return False
 
     def reset_previous_value(self):
         if self.value is not None:
@@ -993,6 +1002,9 @@ class UserDefinedType(Column):
             return
         val.validate()
         return val
+
+    def to_python(self, value):
+        return deepcopy(value)
 
 def resolve_udts(col_def, out_list):
     for col in col_def.sub_types:
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -235,8 +235,6 @@ class Column(object):
         """
         Converts python value into database value
         """
-        if value is None and self.has_default:
-            return self.get_default()
         return value
 
     @property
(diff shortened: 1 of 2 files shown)


mambocab:
I'm +1 if you rebase the last commit into the commit where the `print` was introduced. I'll let @bjmb have the final word though, since he owns the 505 branch.

cassandra/cqlengine/columns.py
```diff
@@ -994,6 +1001,9 @@ def validate(self, value):
         val.validate()
         return val
 
+    def to_python(self, value):
+        return deepcopy(value)
```
bjmb: I think we should call `Column.validate()` on each member of the UDT, like it's done in the other classes inheriting from `Column`.
Otherwise this script would give an error:
```
from cassandra.cqlengine import columns
from cassandra.cqlengine.models import Model
from cassandra.cqlengine import management
from cassandra.cqlengine import connection
from cassandra.cqlengine.usertype import UserType


KEYSPACE = "testkeyspace"

try:
    class UDT(UserType):
        time_col = columns.Time()

    class ModelWithDefault(Model):
        __keyspace__ = KEYSPACE
        id = columns.Integer(primary_key=True)
        time_col = columns.Time()
        udt = columns.UserDefinedType(UDT)

    connection.default()
    management.create_keyspace_simple(KEYSPACE, 1)
    management.sync_table(ModelWithDefault)

    item=ModelWithDefault(id=1)
    item.save()

    user_to_update = UDT()
    user_to_update.time_col = 10

    #user_to_update = UDT(time_col=10)

    item.update(time_col=10, udt=user_to_update)

    first_item = ModelWithDefault.filter(id=1).first()
    print(first_item.id, first_item.time_col,
          first_item.udt.time_col)

finally:
    management.drop_keyspace(KEYSPACE)
```

But if we uncomment the line before the `item.update` it finishes without error (because in this case `BaseUserType.__init__` calls each member's `to_python`)

bjmb:
Looks good to me @aboudreault, could you take a look at my last two commits?

aboudreault:
+1 @bjmb for your additions. Thanks. I have been to reproduce the issue you mentioned.

PR #753, opened 2017-05-04
remove uses of map that don't work in py3

mambocab:
[PYTHON-749](https://datastax-oss.atlassian.net/browse/PYTHON-749):

> There are a couple call sites for the `map` builtin that I believe weren't correctly converted for Python 3 -- they call `map` but don't consume the result. I... would like to see a test for the `LibevReactor` fail before we merge.


aboudreault:
+1

mambocab:
Thanks for the review; merging.

PR #757, opened 2017-05-10
Added test for PYTHON-749

bjmb:


tests/integration/standard/test_connection.py
```diff
@@ -398,3 +398,49 @@ def setUp(self):
...
+        import weakref
+        import copy
+
+        # conn._write_watcher and conn._read_watcher will be closed
+        # when the request is finished so this may no hit them active
+        # every time
+        session.execute_async("SELECT * FROM system.local LIMIT 1")
+        # We have to make a copy because the connections shouldn't
+        # be alive when we verify them
+        live_connections = copy.copy(LibevConnection._libevloop._live_conns)
```
mambocab: `_live_conns` is just a `set`, right? If so, you can just do `live_conections = set(...)` or `tuple(...)` or whatever instead. `copy` is usually only necessary for copying non-collection objects.

tests/integration/standard/test_connection.py
```diff
@@ -398,3 +398,49 @@ def setUp(self):
             raise unittest.SkipTest(
                 'libev does not appear to be installed properly')
         ConnectionTests.setUp(self)
+
+    def test_watchers_are_finished(self):
```
mambocab: Does this need to be decorated to only run with the `libev` reactor or anything like that?

tests/integration/standard/test_connection.py
```diff
@@ -398,3 +398,49 @@ def setUp(self):
...
+        @since 3.10
+        @jira_ticket PYTHON-747
+        @expected_result the watchers are closed
+
+        @test_category connection
+        """
+        cluster = Cluster(connection_class=LibevConnection)
+        session = cluster.connect(wait_for_all_pools=True)
+
+        from cassandra.io.libevreactor import _cleanup
```
mambocab: These imports should be at the top of the file. This one should probably be imported/used as

```from cassandra.io import libevreactor
...
libevreactor._cleanup(weakref.ref(LibevConnection._libevloop))```

File: tests/integration/standard/test_connection.py [hunk 4]
```diff
@@ -398,3 +398,49 @@ def setUp(self):
...
+        import weakref
+        import copy
+
+        # conn._write_watcher and conn._read_watcher will be closed
+        # when the request is finished so this may no hit them active
```
mambocab: `may not`

tests/integration/standard/test_connection.py
```diff
@@ -398,3 +398,49 @@ def setUp(self):
...
+        # every time
+        session.execute_async("SELECT * FROM system.local LIMIT 1")
```
mambocab: How do you mean "this may not hit them active every time"? What is "this" here?

tests/integration/standard/test_connection.py
bjmb: We run ConnectionTests with Libev and Asyncore, not sure why we don't do it with other event loop managers but might be worth to add them

tests/integration/standard/test_connection.py
bjmb: "this" means the test here. They might be closed by the loop before they are closed by _cleanup so the test would succeed even if there was something wrong. It doesn't usually happen, but it's a possibility. I'll add some more lines explaining this.

tests/integration/standard/test_connection.py
bjmb: Agree

PR #759, opened 2017-05-12
Python jenkin fixes

bjmb:


tests/integration/standard/test_metrics.py
```diff
@@ -146,7 +146,7 @@ def test_unavailable(self):
         # Sometimes this commands continues with the other nodes having not noticed
         # 1 is down, and a Timeout error is returned instead of an Unavailable
         get_node(1).stop(wait=True, wait_other_notice=True)
-
+        time.sleep(5)
```
bjmb: Not the ideal solution, bit this fixes the test.

tests/unit/io/test_libevreactor.py
```diff
@@ -296,3 +295,37 @@ def test_partial_message_read(self, *args):
 
         self.assertTrue(c.connected_event.is_set())
         self.assertFalse(c.is_defunct)
+
+    def test_watchers_are_finished(self, *args):
```
bjmb: This test has been moved to unit tests so we can patch more things, it was messing up with the environment in the integration tests.

tests/integration/standard/test_cluster.py
```diff
@@ -282,15 +292,15 @@ def test_connect_on_keyspace(self):
...
             """
-            INSERT INTO test3rf.test (k, v) VALUES (8889, 8889)
+            INSERT INTO test1rf.test (k, v) VALUES (8889, 8889)
             """)
         self.assertFalse(result)
 
-        result = session.execute("SELECT * FROM test3rf.test")
-        self.assertEqual([(8889, 8889)], result)
+        result = session.execute("SELECT * FROM test1rf.test")
+        self.assertEqual([(8889, 8889)], result, "Rows in ResultSet are {0}".format(result.current_rows))
```
bjmb: This assertion still fails, not sure why though.

tests/integration/__init__.py
```diff
@@ -695,16 +701,6 @@ def setUpClass(self):
         self.common_setup(2)
 
 
-class BasicSharedKeyspaceUnitTestCaseWTable(BasicSharedKeyspaceUnitTestCase):
```
bjmb: This class is unused.

tests/integration/standard/test_cluster.py
mambocab: Also, I thiiiink I'd expect this to fail sometimes at RF1 on a 3-node cluster.

tests/integration/standard/test_metrics.py
mambocab: Huh, that's frustrating. Maybe talk to someone to see if there's a possible `ccm` bug here? but for the time being, this is fine.

tests/integration/standard/test_cluster.py
bjmb: Yeah, not often, I'm betting on a race condition on the server, but the error message is something like `[(8889, 8889)] != ResultSet`, so hopefully next time we'll know a little bit more of information

tests/integration/__init__.py
bjmb: I'll check this

PR #762, opened 2017-05-16
Fix tests of Murmur hash algorithm on big endian platforms

frenzymadness:
Because Murmur algorithm is optimized for different platforms, it produces different results on little/big endian systems.
These changes only split results check for little/big endian systems.

### DIFF
--- a/tests/unit/test_metadata.py
+++ b/tests/unit/test_metadata.py
@@ -21,6 +21,7 @@ from binascii import unhexlify
 from mock import Mock
 import os
 import six
+import sys
 import timeit
 
 import cassandra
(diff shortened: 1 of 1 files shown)


mambocab:
Thanks for the PR, I appreciate it. However, I believe it's actually correct for the test, as it exists on `master`, to fail -- our `murmur3` implementation should return results consistent with the server's implementation regardless of the platform. I'm going to close and schedule [PYTHON-653](https://datastax-oss.atlassian.net/browse/PYTHON-653) for evaluation for version 3.11, as we don't have time to schedule it in sooner.

PR #764, opened 2017-05-17
PYTHON-734: ConnectionHeartbeat should mark down HostConnection on OTO

aboudreault:


bjmb:
My concern with this is that I don't think `HostConnection._replace` can be called now. The only place it can originally be called is [here](https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/pool.py#L393) (then it calls itself) and `is_down` has to be `False`. But this, with the current ConvictionPolicy will only happen when we get an `OperationTimedOut` in `connection.last_error` as computed [here](https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/pool.py#L376).

If `mark_host_down` is `True` it won't be called in `HostConnection.return_connection` regardless of the value of `is_down` in that function (in this case the `OperationTimedOut` would get risen [here]( https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/connection.py#L937)). Before the PR, this was the only scenario in which a `connection.last_error` was set to `OperationTimedOut` and`HostConnection._replace` was called.

If `mark_host_down` is `False` I don't know how a `OperationTimedOut` can get to `connection.last_error`. The only relevant place where it's risen is in [this function]( https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/connection.py#L504) . But it would never end I think into `connection.last_error`.

bjmb:
Sorry, I realized it can be called, although this only happens with connections in the ConnectionHeartbeat (which I think is fine), as [this call](https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/connection.py#L1009) can lead to calling all the callbacks and returning it [here](https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/cluster.py#L3520).

The only issue we could have with this is that we are going to return the connection again in the [next line](https://github.com/datastax/python-driver/blob/29ccd1564d28b32620d08d17a89c532781cf0f38/cassandra/connection.py#L1010) but since it's going to be shutdown as well I guess it's not a big issue.

bjmb:
Looks goods to me merging

PR #765, opened 2017-05-17
make _time_remaining a property

mambocab:
This addresses PYTHON-755, in which `_time_remaining` is updated incorrectly, by avoiding updating it at all.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3319,7 +3319,6 @@ class ResponseFuture(object):
         self.message = message
         self.query = query
         self.timeout = timeout
-        self._time_remaining = timeout
         self._retry_policy = retry_policy
         self._metrics = metrics
         self.prepared_statement = prepared_statement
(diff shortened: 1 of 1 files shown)


cassandra/cluster.py
```diff
@@ -3334,6 +3333,12 @@ def __init__(self, session, message, query, timeout, metrics=None, prepared_stat
         self.attempted_hosts = []
         self._start_timer()
 
+    @property
+    def _time_remaining(self):
+        if self.timeout is None:
+            return None
```
aboudreault: Should you return 0 here?

cassandra/cluster.py
```diff
@@ -3364,8 +3369,6 @@ def _on_speculative_execute(self):
         self._timer = None
         if not self._event.is_set():
             if self._time_remaining is not None:
```
aboudreault: I see.... you could remove this IF if you just return 0 in _time_remaining I think.

aboudreault:
minor comments, +1

mambocab:
`timeout`, and by extension `_time_remaining`, can be `None` to signify "never time out" -- see [this timeout-handling code](https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L3391).

There are some test failures -- some are transient Jenkins issues relating to pulling from the ASF's git server, and the other is also flaky on master.

Thanks for the review, merging.

PR #766, opened 2017-05-18
Added test for PYTHON-755

bjmb:


tests/integration/standard/test_policies.py
```diff
@@ -102,3 +111,26 @@ def test_speculative_execution(self):
...
+        @test_category metadata
+        """
+        # We mock this so no messages are sent, otherwise a reponse might arrive
+        # and we would not know how many hosts we queried
+        with patch.object(Connection, "send_msg") as mocked_send_msg:
+            mocked_send_msg.return_value = 100
+
+            statement = SimpleStatement("INSERT INTO test3rf.test (k, v) VALUES (0, 1);", is_idempotent=True)
+            response_future = self.session.execute_async(statement, execution_profile='spec_ep_brr_lim', timeout=2.2)
+            response_future._event.wait()
```
mambocab: I think you can do without this by using `.execute` instead of `.execute_async`.

tests/integration/standard/test_policies.py
```diff
@@ -102,3 +111,26 @@ def test_speculative_execution(self):
...
+        @jira_ticket PYTHON-750
+        @expected_result speculative retries be schedule every fixed period, during the maximum
+        period of the timeout.
+
+        @test_category metadata
+        """
+        # We mock this so no messages are sent, otherwise a reponse might arrive
+        # and we would not know how many hosts we queried
+        with patch.object(Connection, "send_msg") as mocked_send_msg:
+            mocked_send_msg.return_value = 100
```
mambocab: I think you could do this by passing `return_value=100` to the `patch.object` call

tests/integration/standard/test_policies.py
bjmb: The problem with this is that an OperationTimedOut is arisen and later it's harder to count the number of queried hosts

mambocab: Got it -- a comment about that would be helpful, then

PR #769, opened 2017-05-19
755 timeout error followup

mambocab:
@bjmb to review. Our fix for 755 makes us lose a race in setting `self._time_remaining`, but this change makes losing that not matter anymore.


bjmb:
I'm wondering if the `try finally` is necessary now that we don't access `self._current_host.address` if `self._current_host` doesn't exist, could other exceptions be generated inside the block?

mambocab:
It's probably unnecessary, at least for this particular data race, but it can't hurt to make sure we always set it?

PR #773, opened 2017-05-25
PYTHON-734: small refactor to be less invasive in the api

aboudreault:


cassandra/connection.py
```diff
@@ -1007,7 +1007,10 @@ def run(self):
                 for connection, owner, exc in failed_connections:
                     self._raise_if_stopped()
                     connection.defunct(exc)
-                    owner.return_connection(connection, mark_host_down=True)
+                    if not connection.is_control_connection:
+                        # Only HostConnection supports shutdown_on_error
+                        owner.shutdown_on_error = True
+                    owner.return_connection(connection)
```
aboudreault: I hesitated to create a owner.set_shutdown_on_error() method here, but since it's the only place we use it, I haven't. If your prefer that, let me know.

mambocab:
+1, lgtm

PR #775, opened 2017-05-26
Danger danger high voltage

mambocab:
Adds a [`Dangerfile`](danger.systems). Don't merge with the commit containing app changes, that's just for testing the checker.


mambocab:
@bjmb Are you ok with merging this as an experiment? Obviously I'll squash to one commit. If you look at the past few commits, I think they demonstrate that the check works.

PR #777, opened 2017-06-01
Fix DeprecationWarning when using lz4

vetal4444:
When use `python-driver` with latest `lz4` I got a lot of `DeprecationWarning`:

    /usr/local/lib/python3.5/dist-packages/lz4/__init__.py:18: DeprecationWarning: Call to deprecated function or method compress (use lz4.block.compress instead).
      def compress(*args, **kwargs):
    /usr/local/lib/python3.5/dist-packages/lz4/__init__.py:22: DeprecationWarning: Call to deprecated function or method decompress (use lz4.block.decompress instead).
      def decompress(*args, **kwargs):

### DIFF
--- a/cassandra/connection.py
+++ b/cassandra/connection.py
@@ -61,6 +61,12 @@ try:
 except ImportError:
     pass
 else:
+    # The compress and decompress functions we need were moved from the lz4 to
+    # the lz4.block namespace, so we try both here.
+    try:
+        lz4_block = lz4.block
+    except AttributeError:
+        lz4_block = lz4
 
     # Cassandra writes the uncompressed message length in big endian order,
     # but the lz4 lib requires little endian order, so we wrap these
@@ -68,11 +74,11 @@ else:
 
     def lz4_compress(byts):
         # write length in big-endian instead of little-endian
-        return int32_pack(len(byts)) + lz4.compress(byts)[4:]
+        return int32_pack(len(byts)) + lz4_block.compress(byts)[4:]
 
     def lz4_decompress(byts):
         # flip from big-endian to little-endian
-        return lz4.decompress(byts[3::-1] + byts[4:])
+        return lz4_block.decompress(byts[3::-1] + byts[4:])
 
     locally_supported_compressions['lz4'] = (lz4_compress, lz4_decompress)
 


mambocab:
Thanks for the PR! I've created [PYTHON-769](https://datastax-oss.atlassian.net/browse/PYTHON-769) to review it.

Just a heads-up, it may take some time to complete this review -- our queue for our next release is already fairly full, and we may have to do a little research on what versions of `lz4` are available on different Linux distros.

vetal4444:
I can change my PR to support old versions on lz4. Something like that:

    lz4_block = lz4.block if hasattr(lz4, "block") else lz4

    def lz4_compress(byts):
            # write length in big-endian instead of little-endian
            return int32_pack(len(byts)) + lz4_block.compress(byts)[4:]

        def lz4_decompress(byts):
            # flip from big-endian to little-endian
            return lz4_block.decompress(byts[3::-1] + byts[4:])

    locally_supported_compressions['lz4'] = (lz4_compress, lz4_decompress)

What do you think about this?

mambocab:
Ah, sure, that'd be reasonable with some documentation. @aboudreault Do you see any issues with the proposed solution?

aboudreault:
I think I would do more a try/except here since lz4.block will work mostly everytime in new envs.
```
try:
    lz4_block = lz4.block
except AttributeError:
    lz4_block = lz4
```

LGTM when the PR is fixed.

cassandra/connection.py
```diff
@@ -61,18 +61,22 @@
 except ImportError:
     pass
 else:
+    # Old versions of lz4 has not block attribute.
+    # In new versions need to use lz4.block, because functions from
+    # lz4 namespace are deprecated.
+    lz4_block = lz4.block if hasattr(lz4, "block") else lz4
```
aboudreault: @vetal4444 Can you make the change I mentioned in the PR conversation about this? Otherwise, LGTM.

mambocab:
Awesome, thank you! I'm going to rebase and merge. Thanks for catching this and working with us to fix it.

msenol86:
I am still getting this error on Ubuntu 14.04, Python 2.7 and cassandra-driver (3.10). I have installed cassandra-driver via pip

vetal4444:
> I am still getting this warning on Ubuntu 14.04, Python 2.7 and cassandra-driver (3.10). I have installed cassandra-driver via pip

Thats because this fix will be included in next release. See [changelog](https://github.com/datastax/python-driver/blob/master/CHANGELOG.rst) for more details.

kmatt:
I'm still seeing these warnings with cassandra-driver 3.12.0 and Python lz4 0.12.0 (on CentOS 7.7.4.1708). This is with cqlsh from the apache-cassandra-3.11.1-bin.tar.gz tarball.

bjmb:
That tarball is bundled with the driver version 3.10, not 3.11, it's probably using the bundled version instead of the one you have installed.

kmatt:
Even with RPM install. Adding `print cassandra.__file__` to cqlsh.py shows me `/usr/share/cassandra/lib/cassandra-driver-internal-only-3.10.zip/cassandra-driver-3.10/cassandra/__init__.py`

Moving cassandra-driver-internal-only-3.10.zip out of the installed path addresses the issue. Where does this driver locate the bundled version of the driver? An option to use one in the Python site-packages path would be handy.

PR #778, opened 2017-06-01
Python host filter policy PYTHON-761

mambocab:
I need to write more documentation before we merge, but I wanted to get discussion going on API choices.

I chose to model the API here almost exactly off the [Java driver implementation](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/policies/HostFilterPolicy.java). The two relevant choices here:

- the predicate is immutable, and
- host-filtering takes place at policy-population time, not at query-planning time.

Does anyone have strong feelings either way about those? Changing either of these is totally doable, but would add more complexity and make the class harder to test.

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -619,6 +619,7 @@ class WriteType(object):
     A lighweight-transaction write, such as "DELETE ... IF EXISTS".
     """
 
+
 WriteType.name_to_value = {
     'SIMPLE': WriteType.SIMPLE,
     'BATCH': WriteType.BATCH,
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -17,6 +17,7 @@ import logging
 from random import randint, shuffle
 from threading import Lock
 import socket
+from warnings import warn
 
 from cassandra import ConsistencyLevel, OperationTimedOut
 
@@ -396,6 +397,10 @@ class TokenAwarePolicy(LoadBalancingPolicy):
 
 class WhiteListRoundRobinPolicy(RoundRobinPolicy):
     """
+    |wlrrp| **is deprecated. It will be removed in 4.0.** It can effectively be
+    reimplemented using :class:`.HostFilterPolicy`. For more information, see
+    PYTHON-758_.
+
     A subclass of :class:`.RoundRobinPolicy` which evenly
     distributes queries across all nodes in the cluster,
     regardless of what datacenter the nodes may be in, but
@@ -405,12 +410,25 @@ class WhiteListRoundRobinPolicy(RoundRobinPolicy):
     https://datastax-oss.atlassian.net/browse/JAVA-145
     Where connection errors occur when connection
     attempts are made to private IP addresses remotely
+
+    .. |wlrrp| raw:: html
+
+       <b><code>WhiteListRoundRobinPolicy</code></b>
+
+    .. _PYTHON-758: https://datastax-oss.atlassian.net/browse/PYTHON-758
+
     """
     def __init__(self, hosts):
         """
         The `hosts` parameter should be a sequence of hosts to permit
         connections to.
         """
+        msg = ('WhiteListRoundRobinPolicy is deprecated. '
+               'It will be removed in 4.0. '
+               'It can effectively be reimplemented using HostFilterPolicy.')
+        warn(msg, DeprecationWarning)
+        # DeprecationWarnings are silent by default so we also log the message
+        log.warning(msg)
 
         self._allowed_hosts = hosts
         self._allowed_hosts_resolved = [endpoint[4][0] for a in self._allowed_hosts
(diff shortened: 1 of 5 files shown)


mambocab:
Also -- do we think the one integration test I wrote is enough?

cassandra/policies.py
```diff
@@ -405,12 +410,24 @@ class WhiteListRoundRobinPolicy(RoundRobinPolicy):
...
     def __init__(self, hosts):
         """
         The `hosts` parameter should be a sequence of hosts to permit
         connections to.
         """
+        msg = ('WhiteListRoundRobinPolicy is deprecated. '
+               'It will be removed in 4.0.')
+        warn(msg, DeprecationWarning)
+        # DeprecationWarnings are silent by default so we also log the message
+        log.warning(msg)
```
aboudreault: I am not sure it is our job to ensure people are seeing deprecated warnings. Are we doing that somewhere else in the code base?

cassandra/policies.py
```diff
@@ -441,6 +458,85 @@ def on_add(self, host):
...
+    is truthy. Hosts for which ``predicate(host)`` is falsey will be considered
+    :attr:`.IGNORED`, and will not be used in a query plan.
+    """
+
+    _predicate = None
+
+    def __init__(self, child_policy, predicate):
+        super(HostFilterPolicy, self).__init__()
+        self._child_policy = child_policy
+        self._predicate = predicate
```
aboudreault: we could probably check if the predicate is callable here and raise an error if not.

tests/unit/test_policies.py
```diff
@@ -19,6 +19,10 @@
 
 from itertools import islice, cycle
 from mock import Mock, patch
+try:
+    from thread import LockType  # py2
+except ImportError:
+    from _thread import LockType  # py3
```
aboudreault: six.moves._thread

cassandra/policies.py
```diff
@@ -441,6 +458,85 @@ def on_add(self, host):
...
 
 
+class HostFilterPolicy(LoadBalancingPolicy):
+    """
+    A :class:`.LoadBalancingPolicy` subclass that takes an instantiated
+    ``LoadBalancingPolicy`` as a child policy, and a single-argument predicate.
+    This policy defers to the child policy for hosts where ``predicate(host)``
+    is truthy. Hosts for which ``predicate(host)`` is falsey will be considered
+    :attr:`.IGNORED`, and will not be used in a query plan.
+    """
```
aboudreault: I think it might be worth adding an example here. ie. the old WhiteListRoundRobinPolicy done with this new one.

aboudreault:
LGTM Jim! Minor comments added.

tests/unit/test_policies.py
mambocab: Hah -- I swear I looked for this. Thanks

cassandra/policies.py
mambocab: We are -- for instance:

https://github.com/datastax/python-driver/commit/9421d36de79ede7fdb20873fe5b117ee2d05b2e9#diff-9e9c9d4aa3f0469896e27bf747ef9e82R126

IMO the default is crappy enough that we should work around it.

cassandra/policies.py
mambocab: I might slightly prefer `except TypeError` when calling `self._predicate` and raising an error with a more useful error message. It's slightly more idiomatic, with the whole whole "better to ask forgiveness than permission" thing. Do you think one has technical merits over the other?

mambocab:
Oh, yeah, Alan -- did you have feelings either way about the implementation choices I mentioned in my request?

The one problem I can think of with the Java driver's approach is that `distance` calculates `predicate(host) when it's called, while `make_query_plan` uses data calculated during `on_{up,down,add,remove}`. It's a little weird that the results can be different if the conditions that trip the predicate change.

cassandra/policies.py
aboudreault: Also, I think the parameters should be documented in the __init__ function, so that autodoc generate that properly.

aboudreault:
Jim, I think what we have is correct. The hosts list should only change if we call populate (or on_up etc.), not at the query plan time. I don't see where Java does this, can you point me to the right place?

mambocab:
> The hosts list should only change if we call populate (or on_up etc.), not at the query plan time. I don't see where Java does this, can you point me to the right place?

You're correct -- the host list only changes in those situations, and `make_query_plan` defers to the child policy, so it uses results calculated during those host list changes. However, `.distance` doesn't use this result; it calculates `predicate(host)` when called.

So, problems can happen if `host` changes in some way (other than being added/removed or going up/down), or if `predicate` depends on some other kind of state that changes. You could run into a situation where, for instance, a `predicate(host)` is `IGNORED`, but is returned in `make_query_plan`. Is this ok? (I also want to mention this to the Java driver devs to see what they think.)

mambocab:
The latest commit implements this:

https://github.com/datastax/python-driver/pull/778#discussion_r120480828

in what might be a pretty gross way -- could you let me know your thoughts, Alan?

The code could be simpler if we didn't care about providing a public, read-only `.predicate` attr, but I do want to provide that so clients can check the predicate themselves without accessing private attribute.

I believe I've addressed your other comments. I'd like to wait on a decision for  [JAVA-1489](https://datastax-oss.atlassian.net/browse/JAVA-1489) before finalizing our decision on when methods call `.predicate`.

cassandra/policies.py
```diff
@@ -441,6 +460,133 @@ def on_add(self, host):
...
+
+        @wraps(predicate)
+        def predicate_with_better_error_handling(*args, **kwargs):
+            try:
+                return predicate(*args, **kwargs)
+            except TypeError as te:
+                msg = (
+                    'caught the following error while calling ' +
+                    self.__class__.__name__ + " object's predicate method: " +
+                    repr(te.message)
```
aboudreault: I don't think te.message is available for Python3

mambocab: grrrrr

aboudreault:
@mambocab Is there anything that force us to use your public accessor for the predicate? ie:
```
self.predicate(hola)
```

IMO, I think it would be more clean to create a private function that basically call the predicate and do your error handling then use it internally. ie. self._call_predicate(hola). Now, I feel we are scarifying our code for a feature we aren't sure that will be used (the predicate property). What do you think Jim? Feel free to take the final decision on that.

mambocab:
> Now, I feel we are scarifying our code for a feature we aren't sure that will be used (the predicate property). What do you think Jim? Feel free to take the final decision on that.

I kinda like continuing to expose `.predicate` for debugging. I've removed that wild error-handling wrapping code, I'm not sure why I thought it was necessary. Errors thrown by the predicate will show that they were thrown while calling `predicate`.

mambocab:
@aboudreault Are you happy with that solution? I think the existing error reporting is actually ok.

aboudreault:
LGTM!

mambocab:
Thanks for the review. I've cleaned up an unused import and added a test for the deprecation warning, so now I'll squash and merge.

mambocab:
Rebasing https://github.com/datastax/python-driver/pull/778/commits/c220991940dc3c6a19c003847d81cf51ae003708 down to a pre-HFP whitespace commit and a commit for HFP.

PR #779, opened 2017-06-05
Ensure asyncore dispatchers are closed on shutdown

aboudreault:


aboudreault:
changelog entry is missing, I will add it shortly

mambocab:
Unfortunately, I wasn't able to repro this locally, so I can't verify that the change fixes the issue. However, based on [your description of the issue](https://datastax-oss.atlassian.net/browse/PYTHON-767?focusedCommentId=45301&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-45301), I buy that this fixes it.

I was briefly concerned that this could cause a race where an in-flight request that could complete in theory would not return -- but, if we're closing connections, any such requests would already be doomed to fail, right?

If that's the case, I'm +1 if you add a short comment noting that the `conn.close()` loop will error any in-flight requests and allow the asyncore dispatchers to be cleaned up.

aboudreault:
Indeed, at this point, all requests are already doomed anyway.

PR #780, opened 2017-06-09
PYTHON-762: All Heartbeat futures should have the same timeout

aboudreault:
Let's set a global timeout for *all* heartbeat futures rather than waiting one-by-one.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 3.11
 ====
 
+Features
+--------
+* Add idle_heartbeat_timeout cluster option to tune how long to wait for heartbeat responses. (PYTHON-762)
+
 Bug Fixes
 ---------
 * is_idempotent flag is not propagated from PreparedStatement to BoundStatement (PYTHON-736)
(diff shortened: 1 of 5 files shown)


akshaykumar90:
The current solution is an improvement, but I don't think it's the best. We will essentially still wait 30s for timeout, which I believe is just too long to wait for a timeout.

Can we just decouple the timeout value from the heartbeat frequency? They serve two different purposes and it would be useful to be able to tune them separately.

aboudreault:
@akshaykumar90 Ok. Let me take a look at it.

mambocab:
Should this wait for a major revision? `interval_sec` no longer configures how long to wait on a given connection. Or do we consider that a bug? Seems a bit of a stretch to do so.

In addition, this adds an argument in the middle of the positional args to `Cluster.__init__`, and that's a breaking API change.

mambocab:
We had some discussion offline that resolved my concern -- the existing behavior is a bug, so changing configuration to address it is a fix.

We'll want to move the new `Cluster.__init__` param to be close to the other heartbeat-related params in the next major version (barring [a big rework of the way config works](https://datastax-oss.atlassian.net/browse/PYTHON-541)). I'll make a note.

So, once the tests pass, I'm +1.

PR #781, opened 2017-06-12
Added a more complete pre build check to setup.py

bjmb:
The Windows compiler has the initialize method which is the one that raises an Exception when it doesn't find a valid C++ compiler.


mambocab:
I haven't tested this on Windows, but it installs fine on Linux; +1.

PR #782, opened 2017-06-15
Fix RuntimeError on cluster shutdown

vetal4444:
File "core/context.py", line 163, in shutdown
        self.cluster.shutdown()
      File "cassandra/cluster.py", line 1226, in cassandra.cluster.Cluster.shutdown (cassandra/cluster.c:18488)
        for session in self.sessions:
      File "python3.5/_weakrefset.py", line 60, in __iter__
        for itemref in self.data:
    RuntimeError: Set changed size during iteration

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -964,7 +964,7 @@ class Cluster(object):
                         "be returned when reading type %s.%s.", self.protocol_version, keyspace, user_type)
 
         self._user_types[keyspace][user_type] = klass
-        for session in self.sessions:
+        for session in tuple(self.sessions):
             session.user_type_registered(keyspace, user_type, klass)
         UserType.evict_udt_class(keyspace, user_type)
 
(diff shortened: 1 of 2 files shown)


bjmb:
Thanks for the catch and for the PR. We iterate in more places of cluster.py the sessions like
`for session in self.sessions` . I wonder if we should do the same in the rest of them

vetal4444:
Apply this fix to all places with iteration over `self.session`.

mambocab:
Thanks for the PR. I don't believe we'll get this in in the next release; we'd need more time for testing. However, I've created [a JIRA ticket](https://datastax-oss.atlassian.net/browse/PYTHON-793) for it, and currently plan to address it in the release following the next one.

cassandra/cluster.py
```diff
@@ -1235,7 +1235,7 @@ def shutdown(self):
 
         self.control_connection.shutdown()
 
-        for session in self.sessions:
+        for session in tuple(self.sessions):
```
bjmb: I wouldn't replace it here since I think adding/removing a session at the same time we call `cluster.shutdown` is a bug of the application, not of the driver

mambocab: I don't know that I'm convinced -- it's not a big cost to do this, and asynchronous operations could create a session during shutdown. Yes, that's a bug, but I think that the "can't edit sessions while iterating over it" error may not be the best way to communicate that. Am I on the right track?

bjmb: Yeah agree, that shouldn't be the error message.

bjmb:
Thanks for the contribution @vetal4444! I've squashed the two commits and will now merge.

PR #784, opened 2017-06-15
request_ids may not be returned to the pool (PYTHON-739)

mambocab:
Fixes [PYTHON-739](https://datastax-oss.atlassian.net/browse/PYTHON-739).

Like the comment says, `this is bad`. However, I think a big overhaul to the "ref-counting" code here is not in order yet.

This doesn't change the code [where `highest_request_id` can be incremented, but no new id issued](https://github.com/datastax/python-driver/blob/f60bdc38c4a3faba276864fb9d44e22a5816650f/cassandra/connection.py#L432). I believe this is ok, as `highest_request_id` doesn't track the same kind of crucial state that `in_flight` can.

### DIFF
--- a/cassandra/connection.py
+++ b/cassandra/connection.py
@@ -810,7 +810,29 @@ class Connection(object):
         When the operation completes, `callback` will be called with
         two arguments: this connection and an Exception if an error
         occurred, otherwise :const:`None`.
+
+        This method will always increment :attr:`.in_flight` attribute, even if
+        it doesn't need to make a request, just to maintain an
+        ":attr:`.in_flight` is incremented" invariant.
         """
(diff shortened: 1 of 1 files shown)


bjmb:
This fixes the issue, that was the function where the in_flight was being sent wrong and it was messing with other parts of the code where it was used.
I'd also put augment [this](https://github.com/mambocab/python-driver/blob/351aa6b64651cf419f4eaab58af6dfd34cdc6a02/cassandra/connection.py#L439) variable after the assert. No big deal since we never read it I think but just in case it's necessary for the future.

bjmb:
I don't think it's that far from the solution though, since a similar thing is done in the same function when the callback is executed async.

mambocab:
Closing in favor of #786

PR #786, opened 2017-06-16
request_ids may not be returned to the pool (PYTHON-739) (replaces #784)

mambocab:
#784 should have been on a `python-*` branch to run the integration tests.

### DIFF
--- a/cassandra/connection.py
+++ b/cassandra/connection.py
@@ -810,7 +810,29 @@ class Connection(object):
         When the operation completes, `callback` will be called with
         two arguments: this connection and an Exception if an error
         occurred, otherwise :const:`None`.
+
+        This method will always increment :attr:`.in_flight` attribute, even if
+        it doesn't need to make a request, just to maintain an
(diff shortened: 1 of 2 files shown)


mambocab:
Tests ran clean on c44917f, but there are conflicts in the changelog. Rebasing and merging.

cassandra/connection.py
```diff
@@ -436,9 +436,10 @@ def get_request_id(self):
         try:
             return self.request_ids.popleft()
         except IndexError:
-            self.highest_request_id += 1
+            new_request_id = self.highest_request_id + 1
             # in_flight checks should guarantee this
             assert self.highest_request_id <= self.max_request_id
```
bjmb: @mambocab I think the comparison here should be `new_request_id <= self.max_request_id`, otherwise `self.highest_request_id` will end up being `self.max_request_id + 1`

mambocab: Should be addressed in https://github.com/datastax/python-driver/pull/793

PR #788, opened 2017-06-20
Added tests for PYTHON-762

bjmb:


bjmb:
This doesn't produce a OTO.

PR #790, opened 2017-06-28
KEEP_TEST_CLUSTER env var

mambocab:
I had to jump through some hoops to keep a cluster around and examine `debug.log`. This simplifies that, adding a `KEEP_TEST_CLUSTER` env var.

@bjmb This is low-priority/non-urgent. When you get a chance, could you have a look?


bjmb:
LGTM @mambocab , I think this can be useful

PR #792, opened 2017-07-05
Python 785

bjmb:


mambocab:
Dig it, +1 if you add a comment to note that `_only_fields` and `_defer_fields` use the db name.

mambocab:
Also, if you could squash it down into 1 commit and leave a commit message indicating that you changed these collections to hold DB names, that'd be good.

bjmb:
Thanks, merging

PR #794, opened 2017-07-11
PYTHON-653: Fix murmur3 on big-endian systems

aboudreault:


mambocab:
Added a comment -- if you're +1, feel free to squash and merge.

mambocab:
(I should note, I think a comment is particularly important here because we don't test on big-endian platforms, so it'd be good to leave a note.)

PR #795, opened 2017-07-12
PYTHON-774: Ensure unused connections are closed if a Session is deleted by the GC

aboudreault:


mambocab:
Shouldn't this be running on Jenkins? It's on a `python-` branch.

But, if the tests pass, I'm +1.

aboudreault:
Looks like it ran yes. The only failure is not related.

PR #798, opened 2017-07-13
Replaced WhiteHost policy for FilterHost policy in tests

bjmb:
Replaced WhiteListRoundRobinPolicy for HostFilterPolicy + RoundRobinPolicy as part of replacing it. I've left the tests that actually test WhiteListRoundRobinPolicy until it's fully removed.


tests/integration/standard/test_cluster.py
```diff
@@ -69,7 +56,8 @@ def test_ignored_host_up(self):
 
         @test_category connection
         """
-        ingored_host_policy = IgnoredHostPolicy(["127.0.0.2", "127.0.0.3"])
+        ingored_host_policy = HostFilterPolicy(RoundRobinPolicy(),
```
mambocab: It's a small thing, but I think this should be left as is, just to avoid one potential source of bugs

mambocab:
I've also taken the liberty of removing the `.predicate` calls we discussed; feel free to remove that commit if you have other plans for dealing with that.

PR #800, opened 2017-07-17
Added integration tests around querying replicas

bjmb:


bjmb:
Thanks for the review, merging

PR #803, opened 2017-07-20
aws lambda with cassandra python driver problem

tejomayonline:
Unable to import module 'lambda_function': /var/task/cassandra/cluster.so: undefined symbol: PyFPE_jbuf
I am using aws lambda with
 python 3.6 and
 cassandra-driver 3.7.1
After some googling I found some answer with version incompatibility in cython for cassandra driver.Please help.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,6 +8,8 @@ A modern, `feature-rich <https://github.com/datastax/python-driver#features>`_ a
 
 The driver supports Python 2.6, 2.7, 3.3, and 3.4.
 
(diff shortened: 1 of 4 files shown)


aboudreault:
@tejomayonline What is this PR for?  I don't this it is related to the problem describe at all? Can you create a ticket in our Jira instead? https://datastax-oss.atlassian.net

mambocab:
Gonna close this PR for the time being. @tejomayonline It's not clear if or how the driver is responsible for the error you reported. I'm closing this for now, but feel free to update and reopen if you need to. If you simply have a bug report, our JIRA instance is a better forum for that, as Alan mentioned.

PR #805, opened 2017-07-28
Move calls to asyncore to the loop thread

bjmb:
We should remove the changes from the build.yaml before merging this PR.

### DIFF
--- a/cassandra/io/asyncorereactor.py
+++ b/cassandra/io/asyncorereactor.py
@@ -18,9 +18,10 @@ import logging
 import os
 import socket
 import sys
-from threading import Lock, Thread
+from threading import Lock, Thread, Event
 import time
 import weakref
+import sys
 
 from six.moves import range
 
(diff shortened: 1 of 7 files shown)


cassandra/io/asyncorereactor.py
```diff
@@ -239,6 +267,11 @@ def _run_loop(self):
     def add_timer(self, timer):
         self._timers.add_timer(timer)
 
+        # This function is called from a different thread than the
+        # event loop thread, so for this call is thread safe, we must wake up
```
mambocab: This phrasing confused me at first -- I expected to find a data race or something.

I think I used to understand `wake_loop` enough to rephrase, but I don't anymore :grimacing:

mambocab:
So, explaining each commit:


- `whitespace`
  - Just a little cleanup for readability, and adding back a file-terminating newline that shouldn't have been removed. Kinda surprised your editor allows this.
- `rename Handle->WaitableTimer`
  - `Handle` makes it sound like this is a file handle or something. I think `WaitableTimer` captures the purpose of this class -- it's a wrapper around `Timer` with a `.wait` method.
- `grammar correction` and `rewrap comment` -- self-explanatory
- `change decorator name`
  - `notbigclockgranularity` is hard to parse -- eventually it makes sense, "this test should _not_ run if the clock granularity is big", but it took me a minute. I don't know if I like the name I landed on; `skipIfLargeClockGranularity` might be more in keeping with `unittest` conventions.

cassandra/io/asyncorereactor.py
```diff
@@ -295,6 +328,10 @@ def handle_fork(cls):
     @classmethod
     def create_timer(cls, timeout, callback):
         timer = Timer(timeout, callback)
+        return cls.add_timer(timer)
+
+    @classmethod
+    def add_timer(cls, timer):
```
bjmb: Yep, makes sense, I had doubts about adding it.

PR #806, opened 2017-07-28
Simulacron addition to test harness

bjmb:


build.yaml
```diff
@@ -103,6 +103,22 @@ build:
         python setup.py build_ext --inplace --no-cython
       fi
 
+      echo "Installing simulacron"
```
tolbertam: It could be easier/faster to do:

```bash
wget https://github.com/datastax/simulacron/releases/download/0.5.0/simulacron-standalone-0.5.0.jar
```

Also we could bake a simulacron jar into our jenkins image if that makes it easier too.

bjmb: hmm, yeah but that way we would have to update the build.yaml for every new release, this way we would have the latest.
Just saw the `git checkout 0.3.1` which should be removed.

tolbertam: Good point

tests/integration/simulacron/utils.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def set_node(self, cluster_id, datacenter_id, node_id):
+        self.cluster_id = cluster_id
+        self.datacenter_id = datacenter_id
+        self.node_id = node_id
+
+        if self.cluster_id is not None:
+            self.path += "/{}".format(self.cluster_id)
+
+        if self.cluster_id is not None:
```
bjmb: :+1: this looks much nicer

bjmb:
Thanks for the review @mambocab , all good ideas.

PR #807, opened 2017-08-01
Added a test around SELECT JSON

bjmb:
This looks like a normal query test @mambocab, I'd leave it but I don't have strong feelings if we don't add it.


mambocab:
Cool -- a short test like this should be just fine.

Could you move the `consistency_level` assignment to object init? so rather than 

```python
insert_query = SimpleStatement("INSERT INTO test3rf.test(k, v) values (1, 1)")
insert_query.consistency_level = ConsistencyLevel.QUORUM
```

I was thinking 

```python
insert_query = SimpleStatement("INSERT INTO test3rf.test(k, v) values (1, 1)",
                               consistency_level=ConsistencyLevel.QUORUM)
```

Change that (unless you object) and then feel free to squash and merge.

PR #811, opened 2017-08-04
Python 630

bjmb:


tests/integration/simulacron/test_connection.py
```diff
@@ -114,3 +114,36 @@ def test_heart_beat_timeout(self):
...
+
+        prime_query(query_to_prime, then={"delay_in_ms": 2000})
+
+
+        count = itertools.count()
+        future = session.execute_async(query_to_prime, timeout=1)
+        future.add_callbacks(callback, errback, callback_args=(count, ), errback_args=(count, ))
+        self.assertRaises(OperationTimedOut, future.result)
+
+        assert_quiescent_pool_state(self, cluster)
```
bjmb: This fail before the fix because not all the request ID used to make where the OTO happens it's not returned.

tests/integration/simulacron/test_connection.py
```diff
@@ -114,3 +114,36 @@ def test_heart_beat_timeout(self):
...
+
+        # We sleep until the response comes, it's schedule for
+        # two seconds
+        time.sleep(3)
+        self.assertEqual(next(count), 1)
```
bjmb: This would also fail because callback and errback are both called

cassandra/connection.py
```diff
@@ -586,18 +586,23 @@ def process_io_buffer(self):
...
     def process_msg(self, header, body):
+        self.msg_received = True
         stream_id = header.stream
         if stream_id < 0:
             callback = None
             decoder = ProtocolHandler.decode_message
             result_metadata = None
         else:
-            callback, decoder, result_metadata = self._requests.pop(stream_id)
+            try:
```
mambocab: I haven't decided whether I like this change or not. I understand why it works, but I don't like special-casing stream handling here for exactly OTOs.

cassandra/connection.py
```diff
@@ -586,18 +586,23 @@ def process_io_buffer(self):
...
+                callback, decoder, result_metadata = self._requests.pop(stream_id)
+            # This can only happen if the stream_id was
+            # removed due to an OperationTimedOut
+            except KeyError:
+                return
+
             with self.lock:
                 self.request_ids.append(stream_id)
 
-        self.msg_received = True
```
mambocab: What's this removal about? It doesn't seem related to the rest of the changes.

bjmb: I moved it at the beginning of the function since there is a return if there's a KeyError

mambocab: Ah whoops, I see it. :+1:

mambocab:
After some discussion in Slack, we've decided that it'd be nice to have stream-returning handled in exactly one place, but it's not clear how to do that without a major refactoring.

I think a refactoring is in order, but not now. I'll file a ticket and link it here.

Jaume, I'm +1 on merging, but please squash this down to one commit first.

PR #812, opened 2017-08-07
Set params as default when set after the Cluster is created

bjmb:


bjmb:
The reason it wasn't working is because the LBP is never populated (it's not working with any LBP not just DCAwareRoundRobinPolicy)
Despite being in `LEGACY` mode `self.profile_manager.default` is used in [some places](https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L2664). With this PR I think it should be equivalent now to setting the LBP in the arguments for `Cluster` since in this case it's added to the `DEFAULT` profile [here](https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L833)
Now that it's added to `cluster.profile_manager` it would be populated [here](https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L1176)

mambocab:
Closing in favor of #814.

PR #814, opened 2017-08-10
populate default lbp on connection (PYTHON-781)

mambocab:
This fixes the issue described in PYTHON-781, though I haven't been able to write an integration test to demonstrate this yet. Putting it up for high-level review first.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@
 Bug Fixes
 ---------
 * Both _set_final_exception/result called for the same ResponseFuture (PYTHON-630)
+* Use of DCAwareRoundRobinPolicy raises NoHostAvailable exception (PYTHON-781)
 
 
 3.11.0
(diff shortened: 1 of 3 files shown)


bjmb:
My only concern about this one is that we will be calling `self.load_balancing_policy.populate` twice if it's passed as an argument to cluster (once in the new line added and once when `self.profile_manager.populate` is called).

mambocab:
> My only concern about this one is that we will be calling self.load_balancing_policy.populate twice if it's passed as an argument to cluster (once in the new line added and once when self.profile_manager.populate is called).

I hear you, but I tried to fix this and ran into problems because [the later call to `check_supported`](https://github.com/datastax/python-driver/blob/cb09eb3ff2f0984e78a6d7348ed52719877f5a6c/cassandra/cluster.py#L1198) depends on `_cluster_metadata` [here](https://github.com/datastax/python-driver/blob/cb09eb3ff2f0984e78a6d7348ed52719877f5a6c/cassandra/policies.py#L344), and that's only assigned when the profiles are `populate`d.

In principle it'd be nice to fix this, but I don't think it's worth the time.

mambocab:
Thanks for the fix and the review! I've squashed to one commit, merging.

PR #815, opened 2017-08-16
Prevent reconnecting while reconnecting in the control connection

beltran:


bjmb:
> I worry about race conditions where, e.g., a connection could be permanently marked as reconnecting
I think this could only happen if it was really reconnecting the whole time

I agree something is up with the tests, I've been seeing this lately not only on this PR. It looks like on node goes down and doesn't come up and then it fails for 30 tests straight, which is pretty bad looking. I guess this must be some change in ccm or the some change in the way we start the cluster.

mambocab:
Closing, as per discussion on the ticket

PR #818, opened 2017-08-24
[PYTHON-517] Change the value of self._limit in cqlengine/query.py.

sasayabaku:
Chenge the value of self._limit in cqlengine/query.py
self._limit = 10000 ->  self._limit = None.

In the test code, I can get more 10000 rows data without limit() function.


aboudreault:
@sasayabaku Hello! Why do you want to change to default limit value? It is there for a reason: avoiding bad queries that fetch too many rows and fill the memory. Can you explain more the need here?

sasayabaku:
Hi, I saw the ticket [PYTHON-517].

This ticket’s issues are 
1. Make the limit default to None
2. Fix the limit function so that limit(None) works in Python 3

Issue 2 has already fixed. But the limit default still remain 10000.
When we get the columns without limit(), we want to get all columns but we get only 10000 columns.
Therefore, I made the limit default to None.

sasayabaku:
Hi, I failed 1 check but error parts have no relation with my changes.
How I should to do?

tests/integration/cqlengine/query/test_limit.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2013-2017 DataStax, Inc.
```
bjmb: There's no need to create a new file, this can be included in the file `test_queryset`

tests/integration/cqlengine/query/test_limit.py
```diff
@@ -0,0 +1,96 @@
...
+class test_table(Model):
+
+    __keyspace__ = "test_keyspace"
+    user = columns.Integer(primary_key=True)
+    prop = columns.Text()
+
+# Test
+class LimitTest(unittest.TestCase):
+
+    def setUp(self):
```
bjmb: If this tests was included in the file and in the appropriate class like TestQuerySetOperation it would change quite a bit how it's done, you can see how other tests are done there but there wouldn't be need for all this

tests/integration/cqlengine/query/test_limit.py
```diff
@@ -0,0 +1,96 @@
...
+        self.c = connection.setup(['127.0.0.1'], 'cqlengine', protocol_version=3)
+
+    def create_database(self):
+        print("----- CREATE KEYSPACE & TABLE -----")
+
+        create_keyspace_simple(self.KEYSPACE, replication_factor=3)
+        self.session.set_keyspace(self.KEYSPACE)
+
+        try:
+            self.session.execute("CREATE TABLE " + self.TABLE + " (user int, prop text, PRIMARY KEY (user))")
```
bjmb: The way to create a table in CQLEngine is with `sync_table(test_table)`

mambocab:
Thank you for your efforts on this. We won't be reviewing this in the next month or so, so I will close this PR for now, just to keep our issues and PRs tidy. We'll reopen it when we start spending more time working on the next major.

PR #820, opened 2017-08-29
Change protocol to allow sending key space independent of query string (CASSANDRA-10145)

mambocab:
This deals with [PYTHON-678](https://datastax-oss.atlassian.net/browse/PYTHON-678)/[PYTHON-804](https://datastax-oss.atlassian.net/browse/PYTHON-804).

I honestly don't know what the user API looks like for this, so I don't have integration tests yet. I'll consult with folks tomorrow.

### DIFF
--- a/cassandra/protocol.py
+++ b/cassandra/protocol.py
@@ -508,6 +508,7 @@ _WITH_PAGING_STATE_FLAG = 0x08
 _WITH_SERIAL_CONSISTENCY_FLAG = 0x10
 _PROTOCOL_TIMESTAMP = 0x20
 _WITH_KEYSPACE_FLAG = 0x80
+_PREPARED_WITH_KEYSPACE_FLAG = 0x01
 
 
 class QueryMessage(_MessageType):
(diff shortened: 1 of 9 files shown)


aboudreault:
This PR looks good to me Jim, nice job. As you mentioned, I think what's left is the user API. From my understanding, this feature would be available as an optional param in the according Statement subclasses. ie.

```python
s = SimpleStatement('...', keyspace='sensors')
# or
p = PreparedStatement('...', keyspace='sensors')
```
And the QueryMessage construction would be handled in `_create_response_future` as normal.

That said, this is just my understanding... I haven't validated my thought with others.

cassandra/protocol.py
```diff
@@ -768,14 +780,38 @@ class PrepareMessage(_MessageType):
...
+        self.keyspace = keyspace
 
     def send_body(self, f, protocol_version):
         write_longstring(f, self.query)
+
+        flags = 0x00
+
+        if self.keyspace is not None:
+            if ProtocolVersion.uses_keyspace_flag(protocol_version):
+                flags |= _WITH_KEYSPACE_FLAG
```
tolbertam: There's a small bug here, I only noticed it because I made the same mistake in java driver.  When sending a prepare message, the flag value for keyspace is `0x01` not `0x80` like it is for other message types.

mambocab: Whoooooops, thanks Andy

PR #822, opened 2017-09-07
Added tests for PYTHON-814

bjmb:


tests/integration/cqlengine/connections/test_connection.py
```diff
@@ -104,3 +106,23 @@ def test_connection_setup_with_setup(self):
     def test_connection_setup_with_default(self):
         connection.default()
         self.assertIsNotNone(connection.get_connection("default").cluster.metadata.get_host("127.0.0.1"))
+
+        connection.unregister_connection("default")
```
mambocab: Sorry to barge in without being tagged for review, but: this won't execute if the test fails.

I'm not sure where the right place to put it is. `addCleanup` would be the most explicit way to do it, but it's redundant and we can forget to add it. `unregister_connection` doesn't fail if the connection doesn't exist, so maybe we can just add it to `setUp` or `tearDown`?

I think we need some kind of cleanup in `test_connection_setup_with_setup` as well.

bjmb: No problem at all, thank you for taking a look! I'll add it to `tearDown` since it looks like it doesn't produce an error.

mambocab:
Looks great, +1!

PR #823, opened 2017-09-11
Added API and tests for CASSANDRA-10145

bjmb:


cassandra/cluster.py
```diff
@@ -1748,6 +1748,21 @@ def _prepare_all_queries(self, host):
                         if not success:
                             log.debug("Got unexpected response when preparing "
                                       "statement on host %s: %r", host, response)
+            """
```
bjmb: This is my opinion where we get the biggest improvement, we don't have to set the keyspace of the connection, we can just send the query with the keyspace parameter, not sure how to add it though but a big `if ProtocolVersion.uses_keyspace_flag(self._protocol_version)... else`

cassandra/cluster.py
```diff
@@ -2126,11 +2141,13 @@ def _create_response_future(self, query, parameters, trace, custom_payload, time
 
         if isinstance(query, SimpleStatement):
             query_string = query.query_string
+            statement_keyspace = query.keyspace
```
bjmb: The reason we don't compare here if 
`keyspace = query.keyspace if ProtocolVersion.uses_keyspace_flag(self._protocol_version) else None`
is because if the use creates a statement like 
`SimpleStatement("SELECT * from system.local".format(self.table_name), keyspace="system")` I think a exception should be raised. This breakes the current API, since some applications doing this (may be bugged applications) will now raise an Exception if they are using a lower protocol.

cassandra/cluster.py
```diff
@@ -2143,9 +2160,10 @@ def _create_response_future(self, query, parameters, trace, custom_payload, time
                     "BatchStatement execution is only supported with protocol version "
                     "2 or higher (supported in Cassandra 2.0 and higher).  Consider "
                     "setting Cluster.protocol_version to 2 to support this operation.")
+            statement_keyspace = query.keyspace if ProtocolVersion.uses_keyspace_flag(self._protocol_version) else None
```
bjmb: We need to compare here because a BoundStatment always has a keyspace set and this would raise an Exception under V5 protocol.

cassandra/cluster.py
```diff
@@ -2252,8 +2270,9 @@ def prepare(self, query, custom_payload=None):
             log.exception("Error preparing query:")
             raise
 
+        prepared_keyspace = message.keyspace if message.keyspace else self.keyspace
```
bjmb: This will always be `self.keyspace` since the message is built like `message = PrepareMessage(query=query)`. We can add the `keyspace` parameter to `session.prepare` but that's probably a different ticket.

tests/integration/__init__.py
```diff
@@ -182,8 +182,19 @@ def set_default_cass_ip():
         Cluster.__init__.__func__.__defaults__ = tuple(defaults)
 
 
-def get_default_protocol():
+def set_default_beta_flag_true():
```
bjmb: This set the flag `allow_beta_protocol_version` to `True` by default so we don't have to pass the argument. This is temporary since it looks like V5 won't be beta in Cassandra 4.0 although it is now.

cassandra/cluster.py
```diff
@@ -2262,21 +2281,22 @@ def prepare(self, query, custom_payload=None):
...
             try:
-                self.prepare_on_all_hosts(prepared_statement.query_string, host)
+                self.prepare_on_all_hosts(prepared_statement.query_string, host, prepared_keyspace)
             except Exception:
                 log.exception("Error preparing query on all hosts:")
 
         return prepared_statement
 
-    def prepare_on_all_hosts(self, query, excluded_host):
+    def prepare_on_all_hosts(self, query, excluded_host, keyspace=None):
```
bjmb: Same as with prepare, should keyspace be added here as a parameter? I'm more inclined to add it here but maybe the best is to add it together with `session.prepare` since it's only called from inside that function.

tests/integration/standard/test_query.py
```diff
@@ -1175,3 +1179,211 @@ def test_unicode(self):
...
+        session = cluster.connect(self.ks_name)
+        self.addCleanup(cluster.shutdown)
+
+        simple_stmt = SimpleStatement("SELECT * from {}".format(self.table_name), keyspace=self.ks_name)
+        # This raises cassandra.cluster.NoHostAvailable: ('Unable to complete the operation against
+        # any hosts', {<Host: 127.0.0.3 datacenter1>: UnsupportedOperation('Keyspaces may only be
+        # set on queries with protocol version 5 or higher. Consider setting Cluster.protocol_version to 5.',),
+        # <Host: 127.0.0.2 datacenter1>: ConnectionException('Host has been marked down or removed',),
+        # <Host: 127.0.0.1 datacenter1>: ConnectionException('Host has been marked down or removed',)})
+        with self.assertRaises(NoHostAvailable):
```
bjmb: Should setting the keyspace in `SimpleStatement` raise `UnsupportedOperation` or `UnsupportedOperation` for old versions? As mentioned in a previous comment, raising something would break the API, since I believe now the keyspace is just ignored-

bjmb:
Well, as for the second point, we are doing that in CQLEngine when we create `SimpleStatement` so I guess we shouldn't raise an exception and just ignore as until now.

cassandra/cluster.py
```diff
@@ -1748,6 +1748,21 @@ def _prepare_all_queries(self, host):
...
+
+            for ks_chunk in chunks:
+                messages = [PrepareMessage(query=s.query_string, keyspace=s.keyspace) for s in ks_chunk]
+                # TODO: make this timeout configurable somehow?
+                responses = connection.wait_for_responses(*messages, timeout=5.0, fail_on_error=False)
+                for success, response in responses:
+                    if not success:
+                        log.debug("Got unexpected response when preparing "
+                                  "statement on host %s: %r", host, response)
+            """
```
mambocab: Sorry if I've missed something -- is this meant to be uncommented on merge?

tests/integration/standard/test_query.py
```diff
@@ -1251,6 +1250,7 @@ def test_setting_keyspace_and_same_session(self):
 
 @greaterthanorequalcass40
 class SimpleWithKeyspaceTests(QueryKeyspaceTests, unittest.TestCase):
+    @unittest.skip
```
mambocab: Why are we skipping this now?

bjmb: Specifying the keyspace before V5 is allowed right now and raising an exception would break the API. The keyspace is ignored bu there may be people doing it (like us in CQLEngine), I added that as part of PYTHON-504, we can remove the skip after that.

cassandra/cluster.py
```diff
@@ -2232,7 +2228,7 @@ def _on_request(self, response_future):
         for fn, args, kwargs in self._request_init_callbacks:
             fn(response_future, *args, **kwargs)
 
-    def prepare(self, query, custom_payload=None):
+    def prepare(self, query, custom_payload=None, keyspace=None):
```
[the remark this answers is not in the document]
bjmb: I'm more inclined to document it, it shouldn't be long and it may be helpful for some users

mambocab:
I've expanded the new documentation a little -- other than that, looks ready to go!

PR #824, opened 2017-09-14
PYTHON-808: Include hash of result set metadata in prepared stmt id

mambocab:
PR for [PYTHON-808](https://datastax-oss.atlassian.net/browse/PYTHON-808) -- not ready to merge, waiting on [CASSANDRA-10786](https://issues.apache.org/jira/browse/CASSANDRA-10786).


mambocab:
Rebasing `db90e21` onto `master`. This fixes some test bugs and brings in the new keyspace-per-query changes.

cassandra/cluster.py
```diff
@@ -3771,8 +3772,11 @@ def _execute_after_prepare(self, host, connection, pool, response):
                 if self.prepared_statement:
                     # result metadata is the only thing that could have
                     # changed from an alter
-                    _, _, _, result_metadata = response.results
-                    self.prepared_statement.result_metadata = result_metadata
+                    (_, _, _,
+                     self.prepared_statement.result_metadata,
+                     new_metadata_id) = response.results
```
aboudreault: That is becoming a bit messy. We should probably create a follow-up ticket to create a better structure for response results.

cassandra/protocol.py
```diff
@@ -682,9 +686,14 @@ def recv_results_metadata(cls, f, user_type_map):
         else:
             paging_state = None
 
+        if flags & 0x0008:
```
aboudreault: better to create a constant var rather than using the hexa number directly here

cassandra/cluster.py
mambocab: We've already got a ticket for it, actually: [PYTHON-601](https://datastax-oss.atlassian.net/browse/PYTHON-601)

cassandra/protocol.py
mambocab: Yep, thanks for catching this

mambocab:
rebasing 57fe8fb onto master to deal with merge conflicts

cassandra/protocol.py
mambocab: Addressed :+1:

ifesdjeen:
Can we get this one pulled in and committed to the `cassandra-test` branch, as otherwise Cassandra dtests will be failing.

mambocab:
Squashing down 3759324 to a smaller commit history.

PR #827, opened 2017-09-20
The asyncore event loop shouldn't be awoken once closed, call manuall…

bjmb:
…y to close the remaining open connections

### DIFF
--- a/cassandra/io/asyncorereactor.py
+++ b/cassandra/io/asyncorereactor.py
@@ -279,6 +279,9 @@ class AsyncoreLoop(object):
         if not self._thread:
             return
 
+        # The loop shouldn't be woken up from here onwards since it won't be running
+        self._loop_dispatcher._notified = True
+
         log.debug("Waiting for event loop thread to join...")
         self._thread.join(timeout=1.0)
         if self._thread.is_alive():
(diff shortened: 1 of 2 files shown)


cassandra/io/asyncorereactor.py
```diff
@@ -279,6 +279,9 @@ def _cleanup(self):
         if not self._thread:
             return
 
+        # The loop shouldn't be woken up from here onwards since it won't be running
+        self._loop_dispatcher._notified = True
```
bjmb: I don't love this, but I'd rather do it this way than making more complex the logic `_AsyncoreUDPDispatcher` and `_AsyncorePipeDispatcher` by adding for example a `closed` variable.

cassandra/io/asyncorereactor.py
```diff
@@ -288,9 +291,12 @@ def _cleanup(self):
...
 
-        # Ensure all connections are closed and  in-flight requests cancelled
+        # Ensure all connections are closed and in-flight requests cancelled
         for conn in tuple(_dispatcher_map.values()):
             conn.close()
+        # The event loop should be closed, so we call remaining asyncore.close
+        # callbacks from here
+        self._timers.service_timeouts()
 
         log.debug("Dispatchers were closed")
```
bjmb: I believe this in an old bug, since if the `notify_loop` had been called just here, after all the dispatchers are closed and before completely exiting we would've have the same error

cassandra/io/asyncorereactor.py
```diff
@@ -288,9 +291,12 @@ def _cleanup(self):
 
         log.debug("Event loop thread was joined")
 
-        # Ensure all connections are closed and  in-flight requests cancelled
+        # Ensure all connections are closed and in-flight requests cancelled
         for conn in tuple(_dispatcher_map.values()):
             conn.close()
+        # The event loop should be closed, so we call remaining asyncore.close
+        # callbacks from here
+        self._timers.service_timeouts()
```
bjmb: I realized while fixing this that asyncore.close callbacks where added to `AsyncoreLoop._timers` to be called later by the event loop thread but by this point it had already been closed.

PR #830, opened 2017-09-22
Lock joining the libev event loop thread

bjmb:
I'd rather do this solution than add a lock around [this part](https://github.com/datastax/python-driver/blob/d3e960134f97198f244873aa1ce66078947c3a89/cassandra/io/libevreactor.py#L97) and the part where the thread is joined. It looks more simple to me.

### DIFF
--- a/cassandra/io/libevreactor.py
+++ b/cassandra/io/libevreactor.py
@@ -63,6 +63,7 @@ class LibevLoop(object):
         self._started = False
         self._shutdown = False
         self._lock = Lock()
+        self._lock_thread = Lock()
 
         self._thread = None
 
@@ -94,9 +95,11 @@ class LibevLoop(object):
                 should_start = True
 
         if should_start:
-            self._thread = Thread(target=self._run_loop, name="event_loop")
-            self._thread.daemon = True
-            self._thread.start()
+            with self._lock_thread:
+                if not self._shutdown:
+                    self._thread = Thread(target=self._run_loop, name="event_loop")
+                    self._thread.daemon = True
+                    self._thread.start()
 
         self._notifier.send()
 
@@ -126,8 +129,11 @@ class LibevLoop(object):
                     watcher.stop()
 
         self.notify()  # wake the timer watcher
-        log.debug("Waiting for event loop thread to join...")
-        self._thread.join(timeout=1.0)
+
+        # PYTHON-752 Thread might have just been created and not started
+        with self._lock_thread:
+            self._thread.join(timeout=1.0)
+
         if self._thread.is_alive():
             log.warning(
                 "Event loop thread could not be joined, so shutdown may not be clean. "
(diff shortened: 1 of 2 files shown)


cassandra/io/libevreactor.py
```diff
@@ -126,8 +126,11 @@ def _cleanup(self):
                     watcher.stop()
 
         self.notify()  # wake the timer watcher
-        log.debug("Waiting for event loop thread to join...")
-        self._thread.join(timeout=1.0)
+        # PYTHON-752 Thread might have just been created and not started
+        try:
+            self._thread.join(timeout=1.0)
+        except RuntimeError:
+            pass
```
mambocab: I worry a little about catching an error as general as `RuntimeError` -- would it be a good idea to check the message content and re-raising if it's not what we're looking for? I don't love it, but I worry about ignoring something important that subclasses `RuntimeError`.

bjmb: Hmm, yeah, what do you think of adding a debug inside the except like `log.warning(e.message)` or similar?

mambocab: I don't have a strong opinion either way on logging, but I don't think it solves the problem I'm worried about. To be clear, I meant something like:

```python
try:
    self._thread.join(timeout=1.0)
except RuntimeError:
    if 'cannot join thread before it is started' not in e.message:
        raise
```

What are your thoughts on that?

bjmb: Yeah agree, I don't like too much as well examining the error message. Let me see if I can think of better solution, if nothing comes up I think we can leave it like this.

bjmb: I think I'll go with this, I'd rather not add a lock for a thread that's run once at the very end of the execution.

mambocab: I don't think I object to the lock as much as you do, but works for me.

bjmb:
With the previous fix I was getting
```python
Exception in thread event_loop (most likely raised during interpreter shutdown):
  File "/usr/local/lib/python2.7/threading.py", line 810, in __bootstrap_inner
  File "/usr/local/lib/python2.7/threading.py", line 763, in run
  File "/home/jaume/workspace/python-driver/cassandra/io/libevreactor.py", line 113, in _run_loop
<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'debug'
```
Can you review again @mambocab ?

mambocab:
Oh, that's really bizarre. Hmm. Is something wrong with the logger at that point? Has it been torn down?

re: the new code: Does using the new lock fix the race? Seems to me that if we can `join` before starting/finishing starting, we can enter the lock around the `join` before entering the other. Is there some event we want to wait on? I don't like that solution either.

bjmb:
I'm not sure why that happens, must be something relative to the end of the execution, for some reason seems like `log` is not set.
You're right, I think this last commit should fix it.

mambocab:
Cool, ltgm. Not sure what's happening that's making the changelog checker fail, unfortunately.

bjmb:
The last commit changes libevrreactor.py but doesn't change the CHANGELOG, I did it this to make review easier, I'll squash and merge,thanks for the review!

PR #831, opened 2017-09-29
Moved test_getting_replicas to unit tests

bjmb:


mambocab:
Unless there's more integration between the `test_getting_replicas` mixin and the other tests, I think I'd prefer to see this as something like

```python
class GetReplicasTest(unittest.TestCase):
    def _get_replicas(self, token_class):
         ....

    def test_murmur3_tokens(self):
         self._get_replicas(Murmur3Token)

    ....
```

As-is, it feels bolted-on and it's not clear why the `setUpClass` is defined on the subclasses. 

That's a nit, though; feel free to merge if you disagree. Test looks good!

PR #833, opened 2017-10-02
PYTHON-810: Revert Deprecation of WhiteListRoundRobinPolicy

aboudreault:


aboudreault:
@bjmb Can you care of the tests part (no rush)? I think we should test the WhiteListRoundRobinPolicy as before: https://github.com/datastax/python-driver/pull/798/files

PR #834, opened 2017-10-03
WARN on Cluster.__init__ w contact points & no lbp

mambocab:
[PYTHON-812](https://datastax-oss.atlassian.net/browse/PYTHON-812)

I think this is the right way to implement this -- we check if `contact_points` is unset, not if it's the default value. That way, if a user has some weird configuration where they want to connect to `localhost`, but also want a non-default LBP, they should still be able to specify it.


aboudreault:
LGTM!

PR #837, opened 2017-10-03
PYTHON-794: add VIEW and CDC writetypes

mambocab:
[PYTHON-794](https://datastax-oss.atlassian.net/browse/PYTHON-794)

Had to do a little digging to find the writetypes, but I only found them in the documented cases. I can go dig the code back up if requested.

### DIFF
--- a/cassandra/policies.py
+++ b/cassandra/policies.py
@@ -731,6 +731,18 @@ class WriteType(object):
     A lighweight-transaction write, such as "DELETE ... IF EXISTS".
     """
 
+    VIEW = 6
+    """
+    This WriteType is only seen in results for requests that were unable to
+    complete MV operations.
+    """
+
+    CDC = 7
+    """
+    This WriteType is only seen in results for requests that were unable to
+    complete CDC operations.
+    """
+
 
 WriteType.name_to_value = {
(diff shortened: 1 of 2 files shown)


cassandra/policies.py
```diff
@@ -749,6 +749,18 @@ class WriteType(object):
...
+    This WriteType is only seen in results for requests that were unable to
+    complete MV operations.
+    """
+
+    CDC = 7
+    """
+    This WriteType is only seen in results for requests that were unable to
+    complete CDC operations.
+    """
+
```
aboudreault: There types should probably be added in WriteType.name_to_value:  https://github.com/datastax/python-driver/blob/24961b6bb3d978ee4d328374c50db75115280258/cassandra/policies.py#L765

Otherwise, the WriteFailureMessage class wont be able to use them at all: https://github.com/datastax/python-driver/blob/24961b6bb3d978ee4d328374c50db75115280258/cassandra/protocol.py#L319

mambocab:
Good catch, thanks -- could I get a quick re-review?

mambocab:
Squashing 236fe7a onto master and merging, thanks for the review :tada

PR #838, opened 2017-10-04
Add per-table slow-query timeout to table metadata

muru:
CASSANDRA-13002
PYTHON-835

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,7 @@
 
 Features
 --------
+* Add per-table slow_query_log_timeout_in_ms to control slow-query logging (PYTHON-835)
 * Send keyspace in QUERY, PREPARE, and BATCH messages (PYTHON-678)
 
 Bug Fixes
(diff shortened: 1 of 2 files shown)


mambocab:
Thanks for the PR! A couple procedural requests:

- Could you please add a `CHANGELOG` entry to the topmost `Features` section?
- Could you reorder the new option name so that the table options tuple is alphabetically sorted again?

Finally, we have a CLA that you can sign electronically. It seems to not be working at the moment, unfortunately, I'll get back to you when we figure it out. Shouldn't take too long.

After all that's settled, I don't see any reason not to merge this when/if the C* patch is merged. Thanks!

mambocab:
Looks great, thank you.

Thanks for stepping through the CLA -- I'll have to confirm internally that you've signed it, but then I believe this will be good to merge after the C* ticket. Could you please ping this PR after the C* ticket is merged?

mambocab:
This PR and the corresponding C* ticket have been inactive for over a year. Closing.

PR #841, opened 2017-10-11
Added query_by_keys

bjmb:


bjmb:
comparison

aboudreault:
Is the execute_concurrent shown in the Graph your new utility function? I guess not?

bjmb:
It `execute_concurrent` from `cassandra.concurrent`, the new function is `query_by_keys`, which is the red line

aboudreault:
ah, Ok thanks. I thought the IN clause was a raw IN query with all keys.

bjmb:
Yeah, I should've clarified a bit. For a reference [this](https://gist.github.com/bjmb/3feac7b3c08929653b93b2d4ff793996) is what I used for the graph. I misconfigured one thing. This is the real graph, the shape is the same but the values of the y-axis change:
comparison

bjmb:
This is a comparison of the function vs a query with a single IN clause:
comparison_2

This is the comparison of the function (which for this particular case sends three queries) vs sending three IN clauses with random keys:
comparison_3
There is no improvement in this case, so maybe targeting the replicas is not an improvement after all

aboudreault:
It looks like we aren't gaining much with this feature. However, It might be interesting to also check C* load with/without query_by_keys.

bjmb:
I tried with a bigger cluster with 9 nodes and there an improvement for the new function:
comparison

bjmb:
Closed in favor of https://github.com/datastax/python-driver/pull/886

PR #842, opened 2017-10-12
fix 3.11 failures after CASSANDRA-10786

mambocab:
This fixes some test failures; @bjmb can you review?

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -194,8 +194,10 @@ def get_supported_protocol_versions():
     3.10 -> 5(beta),4,3
 `   """
     version = Version(CASSANDRA_VERSION)
-    if version >= Version('3.10'):
+    if version >= Version('4.0'):
         return (3, 4, 5)
+    elif version >= Version('3.10'):
+        return (3, 4)
     elif version >= Version('3.0'):
         return (3, 4)
     elif version >= Version('2.2'):
--- a/tests/integration/long/test_failure_types.py
+++ b/tests/integration/long/test_failure_types.py
@@ -65,20 +65,12 @@ class ClientExceptionTests(unittest.TestCase):
         """
         Test is skipped if run with native protocol version <4
         """
-        self.support_v5 = True
         if PROTOCOL_VERSION < 4:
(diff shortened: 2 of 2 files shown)


tests/integration/long/test_failure_types.py
```diff
@@ -71,7 +71,7 @@ def setUp(self):
                 "Native protocol 4,0+ is required for custom payloads, currently using %r"
                 % (PROTOCOL_VERSION,))
         try:
-            self.cluster = Cluster(protocol_version=ProtocolVersion.MAX_SUPPORTED, allow_beta_protocol_version=True)
+            self.cluster = Cluster(protocol_version=ProtocolVersion.MAX_SUPPORTED)
```
bjmb: I'd expect this to fail still since now we will try to instantiate Cluster with `ProtocolVersion.V5` (because `ProtocolVersion.MAX_SUPPORTED = max(V5, V4, V3, V2, V1)`) but without setting `allow_beta_protocol_version=True`, I think we have to remove the whole `try ... except` and just do:
`self.cluster = Cluster(protocol_version=PROTOCOL_VERSION)`

tests/integration/__init__.py
```diff
@@ -195,7 +195,7 @@ def get_supported_protocol_versions():
 `   """
     version = Version(CASSANDRA_VERSION)
```
bjmb: This looks good to me, could we take the chance to update this and add:
```python
if version >= Version('4.0'):
    return (3, 4, 5)
if version >= Version('3.11'):
    return (3, 4)
...

tests/integration/long/test_failure_types.py
mambocab: The try/except logic makes the test pass, but you're right that there's no point to this now. I'll remove the try/except.

tests/integration/__init__.py
mambocab: can do :+1:

mambocab:
Made the requested changes, @bjmb could you have another look? Also did a small refactoring for readability.

bjmb:
Looks good @mambocab !

PR #844, opened 2017-10-17
Don't set keyspace in HostConnection if it doesn't exit

beltran:


cassandra/connection.py
```diff
@@ -799,9 +801,9 @@ def set_keyspace_blocking(self, keyspace):
                              consistency_level=ConsistencyLevel.ONE)
         try:
             result = self.wait_for_response(query)
-        except InvalidRequestException as ire:
+        except InvalidRequest as ire:
```
bjmb: This was a bug from before, here we received a `InvalidRequest` not an `InvalidRequestException`.

cassandra/cluster.py
```diff
@@ -2403,16 +2409,20 @@ def run_add_or_renew_pool():
...
                     self._lock.release()
                     set_keyspace_event = Event()
                     errors_returned = []
 
                     def callback(pool, errors):
                         errors_returned.extend(errors)
                         set_keyspace_event.set()
-
-                    new_pool._set_keyspace_for_all_conns(self.keyspace, callback)
+                    try:
```
bjmb: Issue from before, if an exception was risen here (like the connection was suddenly closed), the error risen was something like `Can't release lock if it isn't acquired`

absurdfarce:

PR #845, opened 2017-10-17
Python 838 warn lbp executionprofile

mambocab:
Addresses [PYTHON-838](https://datastax-oss.atlassian.net/browse/PYTHON-838).

I've given the `ProfileManger` the responsibility of iterating over profiles, which I think is the right call. I don't know if having the manager know about `_explicit_load_balancing_policy` is great, but I think it's better than having `ExecutionProfile` objects know or care about contact points.

Thoughts?


cassandra/cluster.py
```diff
@@ -772,24 +796,18 @@ def __init__(self,
...
-                     '(contact_points = {cp}, lbp = {lbp}'
-                     ''.format(cp=contact_points, lbp=load_balancing_policy))
-
         if contact_points is not None:
-            if contact_points is _UNSET_ARG:
-                contact_points = ['127.0.0.1']
+            if contact_points is _NOT_SET:
+                _contact_points_to_set = ['127.0.0.1']
+            else:
+                _contact_points_to_set = contact_points
```
aboudreault: Do we need that variable? Can't we just use `self.contact_points` and `contact_points`  accordingly?

mambocab: Is this what you're proposing?
 
```python
        if contact_points is not None:
            if contact_points is _NOT_SET:
                self.contact_points = ['127.0.0.1']
            else:
                self.contact_points = contact_points

            if isinstance(self.contact_points, six.string_types):
                raise TypeError("contact_points should not be a string, it should be a sequence (e.g. list) of strings")

            if None in self.contact_points:
                raise ValueError("contact_points should not contain None (it can resolve to localhost)")
            self.contact_points = _contact_points_to_set
```

Yeah, that seems better.

cassandra/cluster.py
```diff
@@ -259,6 +268,23 @@ class ProfileManager(object):
     def __init__(self):
         self.profiles = dict()
 
+    def _validate_contact_points(self, contact_points):
```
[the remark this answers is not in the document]
mambocab: How about this?:

- rename to `_check_for_lbps` or something like that
- remove the `contact_points` param and `if contact_points is _NOT_SET` from this method
- call this as `if contact_points is _NOT_SET: self.profile_manager._check_for_lbps` in `Cluster.__init__`

I'll make a note in `Cluster` where this would change.

cassandra/cluster.py
```diff
@@ -855,11 +873,24 @@ def __init__(self,
...
                                  "load_balancing_policy or default_retry_policy. Configure this in a profile instead.")
             self._config_mode = _ConfigMode.LEGACY
+
         else:
             if execution_profiles:
                 self.profile_manager.profiles.update(execution_profiles)
                 self._config_mode = _ConfigMode.PROFILES
 
+        if self._config_mode is _ConfigMode.PROFILES:
+            self.profile_manager._validate_contact_points(contact_points)
```
mambocab: Above, I'm proposing changing this to

```python
if self._config_mode is _ConfigMode.PROFILES:
    if contact_points is not _NOT_SET:
        self.profile_manager._check_for_lbps()
```

cassandra/cluster.py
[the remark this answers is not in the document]
aboudreault: `_check_for_lbps` or `_validate_lbps` sounds good.

cassandra/cluster.py
aboudreault: :+1:

cassandra/cluster.py
aboudreault: :+1:

mambocab:
Shuffled things around a bit so we can check in `add_execution_profile` too. I actually like this design quite a bit more too -- `Cluster` now just asks the `ProfileManager` for the names of EPs with a particular property and uses the names to format a message, which is better than the `ProfileManager` having some idea about what `contact_points` are and aren't acceptable. Alan, could you check the new work?

aboudreault:
LGTM Jim!

mambocab:
Adding docs, rebasing onto d72f0e0c, and merging.

PR #846, opened 2017-10-20
Add `NO_COMPACT` version to CQL protocol, patch for CASSANDRA-10857.

ifesdjeen:
cc @aholmberg @mambocab

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Features
 * WriteType.CDC and VIEW missing (PYTHON-794)
 * Warn on Cluster init if contact points are specified but LBP isn't (PYTHON-812)
 * Include hash of result set metadata in prepared stmt id (PYTHON-808)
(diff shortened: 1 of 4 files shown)


mambocab:
I've pushed a couple small changes:

- I've added a changelog entry, and
- I moved the new `kwarg` to the end of the `Connection.__init__` call signature to preserve the API for callers that use positional arguments.

Otherwise, looks good to me. If this is definitely the API we want, it looks like we can merge before CASSANDRA-10857 is merged. I'll confirm with the team, but @ifesdjeen let me know when you want this merged. Thanks for the PR!

mambocab:
(Tracking as [PYTHON-839](https://datastax-oss.atlassian.net/browse/PYTHON-839) in JIRA.)

ifesdjeen:
I'd say we can merge it before 10857 is merged, it's completely backwards-compatible.

mambocab:
Awesome, thanks! Merging.

PR #848, opened 2017-10-24
Added upgrade tests

bjmb:


build.yaml
```diff
@@ -139,6 +160,6 @@ build:
       EVENT_LOOP_MANAGER=$EVENT_LOOP_MANAGER CASSANDRA_VERSION=$CCM_CASSANDRA_VERSION VERIFY_CYTHON=$FORCE_CYTHON nosetests -s -v --logging-format="[%(levelname)s] %(asctime)s %(thread)d: %(message)s" --with-ignore-docstrings --with-xunit --xunit-file=standard_results.xml tests/integration/standard/ || true
 
       echo "==========RUNNING LONG INTEGRATION TESTS=========="
-      EVENT_LOOP_MANAGER=$EVENT_LOOP_MANAGER CASSANDRA_VERSION=$CCM_CASSANDRA_VERSION VERIFY_CYTHON=$FORCE_CYTHON nosetests -s -v --logging-format="[%(levelname)s] %(asctime)s %(thread)d: %(message)s" --with-ignore-docstrings --with-xunit --xunit-file=long_results.xml tests/integration/long/ || true
+      EVENT_LOOP_MANAGER=$EVENT_LOOP_MANAGER CASSANDRA_VERSION=$CCM_CASSANDRA_VERSION VERIFY_CYTHON=$FORCE_CYTHON nosetests -s -v --logging-format="[%(levelname)s] %(asctime)s %(thread)d: %(message)s" --exclude-dir=tests/integration/long/upgrade --with-ignore-docstrings --with-xunit --xunit-file=long_results.xml tests/integration/long/ || true
```
mambocab: If we're doing this, I don't see a great reason for the `upgrade` tests to be under `long`.

tests/integration/long/upgrade/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        return self.upgrade_done.is_set()
+
+    def wait_for_upgrade(self, timeout=None):
+        """
+        Waits until the upgrade has completed
+        """
+        self.upgrade_done.wait(timeout=timeout)
+
+    def _upgrade_one_node(self):
```
mambocab: I don't love how much state this needs to maintain -- maybe a better design would be `def upgrade_node(self, node_id): ...` and callers can use

```python
for node_id in self.nodes:
    self.upgrade_node(node_id)
    ...
```

bjmb: Hmm, yeah I did it this way because I figured it wouldn't be important which node is upgraded, you just want to get the next node upgraded. To me it's a trade-off of adding more complexity to be more flexible.

tests/integration/long/upgrade/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Decorator used to specify the upgrade paths for a particular method
+    """
+    def __init__(self, paths):
+        self.paths = paths
+
+    def __call__(self, method):
+        @wraps(method)
+        def wrapper(*args, **kwargs):
+            for path in self.paths:
```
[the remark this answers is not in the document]
bjmb: It's not ideal, but `method(*args, **kwargs)` may call some methods which depend on `UPGRADE_CLUSTER` (like `start_upgrade`, which eventually calls `_upgrade_one_node`) , and it'd be hard to pass that argument to every test method so it can pass it along when it calls a base method.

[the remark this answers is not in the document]
mambocab: Yeah -- I generally prefer the argument-passing style, even if it means adding the param to lots of methods. I think it's easier to debug and read than setting values in one place and reading in another. But, this still works!

tests/integration/long/upgrade/__init__.py
mambocab: Yeah, I guess we have different views of complexity -- but you're right that this is the only method that uses this state, so it's not too hard to read. :+1:

PR #850, opened 2017-10-31
Added tests for PYTHON-839

bjmb:


tests/integration/__init__.py
```diff
@@ -162,7 +162,7 @@ def set_default_cass_ip():
 
 def set_default_beta_flag_true():
     defaults = list(Cluster.__init__.__defaults__)
-    defaults = defaults[:-3] + [True] + defaults[-2:]
+    defaults = defaults[:-4] + [True] + defaults[-3:]
```
mambocab: We can future-proof this a little better with

`defaults[:27] + [True] + defaults[28:]`:

```python
In [1]: from cassandra.cluster import Cluster

In [2]: defaults = list(Cluster.__init__.__defaults__)

In [3]: (defaults[:-4] + [True] + defaults[-3:]) == (defaults[:27] + [True] + defaults[28:])
Out[3]: True
```

PR #858, opened 2017-11-14
PYTHON-629: Added a build.yaml for building wheels

ArunTejCh:


mambocab:
The new `adhoc` job looks good to me! Why are the regular CI jobs removed, though? I may have missed something in your description of how Jenkins works here, but I would have thought we want the `adhoc` job to be in the same build.yaml as the rest of CI.

ArunTejCh:
Hi Jim, Initially I wanted to keep wheel builds in a separate build.yaml. This will avoid any conflicts with the old CI jobs. I'll merge this build.yaml into master once macOS and conflicts are resolved.

mambocab:
Ah, I see -- looks good to me then; we'll clean up before merge to master.

@bjmb can you give a final review to make sure this will play well with existing Jenkins stuff?

build.yaml
```diff
@@ -1,144 +1,57 @@
...
+            for whl in wheelhouse/*cassandra_driver*.whl; do
+              auditwheel repair "$whl" -w /io/wheelhouse/
+            done
+            EOF
+
+            chmod +x build-wheels.sh
+            sudo docker run --rm -v `pwd`:/io quay.io/pypa/manylinux1_x86_64 /io/build-wheels.sh
+
+            # Downloading and running verification script
+            wget https://gist.githubusercontent.com/bjmb/0fa72e65f7f07bf34b813b213b9797c8/raw/112758c37f718c3d2886abd786074d526dbdb967/verify_driver_installation.py
```
bjmb: I'm wondering if we should add this file to the python driver instead of pulling it from my gist, maybe add it to the wheels folder in case we create it

build.yaml
```diff
@@ -1,144 +1,57 @@
...
+package:
+  include: # list of files and glob paths to include in the artifact, relative to the current working directory
+    - wheelhouse/*cassandra_driver*.whl
+release:
+    after:
+      each:
+        - script: |
+            set -e
+            sudo docker pull quay.io/pypa/manylinux1_x86_64
+            cat << 'EOF' > build-wheels.sh
```
bjmb: It may be useful to have this as a separate file. It'd make things easier if we want to build the wheels locally. Maybe we can create a new folder and put it inside, wdyt @ArunTejCh @mambocab ?

TheRealFalcon:
I independently started chasing PYTHON-629 recently, but ran into 
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=873791
Some background:
https://lists.gt.net/python/dev/1293958

TL;DR, they broke the extension ABI, and our driver is affected because it uses cython. As long as extensions are compiled using a python that was compiled without the '--with-fpectl' flag, we should be fine. That flag, along with the fpectl module, was ripped out of python entirely very recently, but most distros still have a python that was compiled with it. Extension modules will fail at runtime if they are run on a distro without it but were compiled on a distro that has it.

Currently, only bleeding edge distros like Debian testing and arch (and maybe that docker image?) will have the change, but it'll eventually propagate everywhere. Can we add a test to the verification script along the lines of:
```python
try:
    import fpectl
    raise Exception("Must use a python compiled without the '--with-fpectl' flag")
except ImportError:
    pass
```
If it fails, we need to recompile on a more up to date python.

bjmb:
I see, the problem is compiling with `--with-fpectl` and running on a python not compiled with `--with-fpectl`. But the other way around works fine. Good catch! I'll add that to the testing script

build.yaml
ArunTejCh: You can copy-paste the whole script portion of the build.yaml into a local script if needed. You guys have to decide if a new file is needed

ArunTejCh:
The original post in https://lists.gt.net/python/dev/1293958 mentions that the manylinux1 docker images are already setup in a way to avoid this issue. I hope we can add the suggestion by james to our testing script.

mambocab:
> I hope we can add the suggestion by james to our testing script.

I'm +1 to giving that a try -- does that work on the `manylinux1` docker image?

ArunTejCh:
It also seems like auditwheel can catch the fpectl setting. 
https://github.com/pypa/auditwheel/blob/master/auditwheel/main_show.py#L35
https://github.com/h2oai/h2o4gpu/issues/201

mambocab:
After some discussion off-PR, I think we all agree that the checker script should be in a subdirectory. Any thoughts on a name? Maybe `helpers` or `dev-scripts`? Off the top of my head, I can't think of a known convention, but to avoid confusion, we'll want to make clear that it does not contain stuff to help users.

ArunTejCh:
PYTHON-629 has been implemented using travis on top of the multibuild framework here https://github.com/riptano/python-dse-driver-wheels This PR is no longer needed

TheRealFalcon:
That URL 404's for me.

PR #861, opened 2017-11-17
PYTHON-861: query: make sure the retry_policy is passed to BoundStatements

iksaif:
Before this patch all prepared statements would end up with
the default RetryPolicy.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,10 @@
+3.13.0
+======
+
+Bug Fixes
+---------
+* Support retry_policy in PreparedStatement (PYTHON-861)
+
 3.12.0
 ======
 November 6, 2017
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -1,3 +1,10 @@
+3.13.0
+======
+
+Bug Fixes
+---------
+* Support retry_policy in PreparedStatement
```
bjmb: Can you add here the number of the ticket, like
Support retry_policy in PreparedStatement (PYTHON-861)

iksaif: done

iksaif:
Yes, the goal is to avoid explicitly binding prepared statements (since this happens implicitly in `execute`)

bjmb:
Thanks @iksaif!  Merging

PR #863, opened 2017-11-22
PYTHON-512: cqlengine: Add LIKE filter operator

aboudreault:


mambocab:
Could you add some docs? Should fit right in [here](https://github.com/datastax/python-driver/blob/master/docs/cqlengine/queryset.rst#filtering-operators).

docs/cqlengine/queryset.rst
```diff
@@ -191,6 +191,21 @@ Filtering Operators
...
+        The LIKE operator is available for text columns that have a SASI secondary index.
+
+        .. code-block:: python
+
+            q = Automobile.objects.filter(model__like='%Civic%').allow_filtering()
+
+        Limitations:
+        - Currently, cqlengine does not support SASI index creation. To use this feature,
+          you need to create the SASI index using the core driver.
+        - Queries using LIKE must use allow_filtering() since the *model* column has no
```
mambocab: Is this true? These docs demonstrate LIKE queries without ALLOW FILTERING:

https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useSASIIndex.html

aboudreault: This is a cqlengine limitation, not a cql limitation. If a column is not flagged with `index=True`, cqlengine will reject the execution. Here, adding a index=True is not a good solution since it might occasion the creation of undesired secondary indexes is sync_table is used.

mambocab: Ohhhh I see. Cool, I get it, ignore my comment :+1:

PR #865, opened 2017-11-24
PYTHON-813: __del__ method in Session is throwing a exception

aboudreault:


cassandra/cluster.py
```diff
@@ -2435,8 +2435,11 @@ def __exit__(self, *args):
         self.shutdown()
 
     def __del__(self):
-        # Ensure all connections are closed, in case the Session object is deleted by the GC
-        self.shutdown()
+        try:
+            # Ensure all connections are closed, in case the Session object is deleted by the GC
+            self.shutdown()
+        except:
```
mambocab: Should we be catching everything here? Or at least log the caught exception? I don't have strong feelings, just looking for your thoughts.

aboudreault: Thinking about it... I also don't want to log unuseful exceptions in the user log, e.g the TypeError related to connection shutdown.

aboudreault: I think I will leave things as is. My reasoning: 
* At this point in the code, the user that cares about cleanup shutdown has surely called cluster.shutdown() explicitly, so he will see any exception when doing so.
* If the user is not doing a clean shutdown explicitly, he probably doesn't care about any shutdown errors.

Let me know if you agree.

mambocab: Yeah, I agree, sounds good. Could you leave a comment about this, just so I don't come in and try to "fix" it later?

PR #873, opened 2017-12-04
PYTHON-836: Added lock to query plan for speculative execution

bjmb:


cassandra/cluster.py
```diff
@@ -3534,6 +3534,10 @@ def _make_query_plan(self):
         # calls to send_request (which retries may do) will resume where
         # they last left off
         self.query_plan = iter(self._load_balancer.make_query_plan(self.session.keyspace, self.query))
+        # Make iterator thread safe when there can be a speculative delay since it could
+        # from different threads
+        if self._spec_execution_plan is ResponseFuture._spec_execution_plan:
```
mambocab: I don't follow this identity check -- so, we make it threadsafe in the case where the execution plan is the `NoSpeculativeExecutionPlan`?

My gut says you mean `if self._spec_execution_plan is not ResponseFuture._spec_execution_plan`. Even with that behavior, I want to bring up a couple cases:

- The user passes a new `NoSpeculativeExecutionPlan` object -- in this case, we take a performance hit for no good reason. Maybe we don't care about that though, since the user should really just pass `None`.
- The user sets `ResponseFuture._spec_execution_plan` as a way to set a default. We also might not care about this, since it's an underscored attribute.

One thought is that we could get around that by checking `isinstance(self._spec_execution_plan, NoSpeculativeExecutionPlan)`?

mambocab:
Other than my little comment, which you can ignore if you don't think it's an issue, I'm +1.

cassandra/cluster.py
bjmb: :+1: you're right, good catch, I'll take you check, overall seems safer.

PR #875, opened 2017-12-06
PYTHON-862 asyncore loop may be None at shutdown

bjmb:


bjmb:
Closed in favor of https://github.com/datastax/python-driver/pull/879

PR #876, opened 2017-12-06
PYTHON-846: Fix DeprecationWarning of log.warn

aboudreault:


mambocab:
There are a couple test failures in the integration tests, but they don't look related to the changes. If you agree, I'm +1 :+1:

PR #879, opened 2017-12-11
PYTHON-862: use self.create_timer instead of AsyncoreConnection.creat…

bjmb:
…e_timer


mambocab:
LGTM -- I've copied logic from the ticket into a test that, locally, fails on master as described and passes on the fixed branch. Feel free to merge, I'm +1.

bjmb:
Awesome thanks for the test!

PR #882, opened 2017-12-13
PYTHON-877: Add some knowns deprecated warnings for 4.x

aboudreault:


mambocab:
Needs a changelog entry, but otherwise looks good. +1 if the unit tests pass after adding the changelog entry.

PR #884, opened 2017-12-15
Replaced event in ResponseFuture for lock

beltran:


mambocab:
Closing for now -- PYTHON-880 covers this issue, and we can reopen when we want to prioritize this.

PR #885, opened 2017-12-17
Allow to connect to a list of cassandra nodes with distinct port numbers

reixd:
Allow to connect to a list of cassandra nodes with distinct port numbers instead of using the default port number (9042) for all nodes.
If the contact_point did not specified a port number, the default port (e.g. 9042) is used.

This feature can be useful if for example in a dev environment in which a Docker cassandra cluster needs to be accessed externally. In this case the cluster is only reachable over a single (public) IP address, therefore requiring to specify different public port numbers for each cassandra node.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -813,10 +813,19 @@ class Cluster(object):
                 raise ValueError("contact_points should not contain None (it can resolve to localhost)")
             self.contact_points = contact_points
 
-        self.port = port
-
-        self.contact_points_resolved = [endpoint[4][0] for a in self.contact_points
-                                        for endpoint in socket.getaddrinfo(a, self.port, socket.AF_UNSPEC, socket.SOCK_STREAM)]
+        self.contact_points_resolved = []
+        self.contact_points_resolved_w_port = {}
+        # Allow to connect to a list of cassandra nodes with distinct port numbers.
+        # Iterate over all contact_points and split (if necessary) the IP address
+        # from the port number. Add the IP and port mapping to dict.
+        for a in self.contact_points:
+            if not ':' in a:
+                a += ':{0}'.format(port)
+            h, p = a.split(':')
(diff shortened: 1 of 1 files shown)


reixd:
Attached connection log:
[test_connection.log](https://github.com/datastax/python-driver/files/1565469/test_connection.log)

reixd:
Just forgot:
Usage example: `$Endpoint:$Port`

`cluster = Cluster(['192.168.0.1:9001', '192.168.0.2:9002', '127.0.0.1'])`

cassandra/cluster.py
```diff
@@ -1202,7 +1211,7 @@ def _make_connection_kwargs(self, address, kwargs_dict):
         if self._auth_provider_callable:
             kwargs_dict.setdefault('authenticator', self._auth_provider_callable(address))
 
-        kwargs_dict.setdefault('port', self.port)
+        kwargs_dict.setdefault('port', self.contact_points_resolved_w_port[address])
```
bjmb: This only works if you specify all the nodes in the contact list, otherwise when the new nodes are discovered it will give a KeyError

reixd: That's a good point. 
What would be a good solution to the problem? New Cluster subclass?

mambocab:
I'm going to close this. We decided recently that we don't want to support this feature -- see [RUBY-117](https://datastax-oss.atlassian.net/browse/RUBY-117?focusedCommentId=36802&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-36802):

> After some consideration with other driver engineers, we've decided to not implement this feature request.
> - It makes the api messy since port can now be specified in multiple ways.
> - network topology events (like new hosts being added to the cluster) won't be handled properly.
>   - C* itself assumes that all nodes run on the same port, so we have to ignore some of the metadata that C* returns related to nodes.
> - There is a fairly simple workaround: instead of creating ssh tunnels with client 'localhost', set up tunnels with different client ip addresses that alias localhost (e.g. 127.0.0.1, 127.0.0.2, etc.).

PR #886, opened 2017-12-18
Added group_keys_by_host

bjmb:


cassandra/metadata.py
```diff
@@ -2565,3 +2566,52 @@ def _cql_from_cass_type(cass_type):
...
+            routing_key = serialized_key[0]
+        else:
+            routing_key = b"".join(Statement._key_parts_packed(serialized_key))
+        all_replicas = cluster.metadata.get_replicas(keyspace, routing_key)
+        # First check if there are local replicas
+        valid_replicas = list(host for host in all_replicas if
+                              host.is_up and cluster._default_load_balancing_policy.distance(
+                                  host) == HostDistance.LOCAL)
+        if not valid_replicas:
+            valid_replicas = list(host for host in all_replicas if host.is_up)
```
aboudreault: This one is never used unless empty?

cassandra/metadata.py
```diff
@@ -2565,3 +2566,52 @@ def _cql_from_cass_type(cass_type):
...
+                # We will group under this statement all the keys for which
+                # we haven't found a valid replica
+                keys_per_host[NO_VALID_REPLICA].append(key)
+        else:
+            for replica in valid_replicas:
+                if replica in keys_per_host:
+                    keys_per_host[replica].append(key)
+                    break
+            else:
+                keys_per_host[valid_replicas.pop()].append(key)
```
aboudreault: I think it could be interesting to not always select blindly the last replica. It would be nice if we could have this done in a random way... but based on the previously selected replica. We can discuss that by voice so I can clarify what I'm thinking of.

cassandra/metadata.py
```diff
@@ -2565,3 +2566,52 @@ def _cql_from_cass_type(cass_type):
...
+
+    Example usage::
+        result = group_keys_by_host(session, "system", "peers",
+                ("peer", "data_center"), (("127.0.0.1", ), ("127.0.0.2", )))
+    """
+    cluster = session.cluster
+
+    partition_keys = cluster.metadata.keyspaces[keyspace].tables[table].partition_key
+
+    serializers = list(types._cqltypes[partition_key.cql_type] for partition_key in partition_keys)
```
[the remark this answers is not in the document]
bjmb: Huh, I thought that returned an iterator in python3, just checked it returns a list, I'll use [...]

cassandra/metadata.py
```diff
@@ -2565,3 +2566,52 @@ def _cql_from_cass_type(cass_type):
         return cass_type.subtypes[0].cql_parameterized_type()
     else:
         return cass_type.cql_parameterized_type()
+
+
+NO_VALID_REPLICA = object()
+
+
+def group_keys_by_host(session, keyspace, table, keys):
```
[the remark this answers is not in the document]
bjmb: I tried that initially but there a circular dependence problem. I got this exception due to adding
`from cassandra.pool import HostDistance`

```python
  File "/home/jaume/workspace/python-driver/z_poc/group_by_keys.py", line 1, in <module>
    from cassandra.util import group_keys_by_host
  File "/home/jaume/workspace/python-driver/cassandra/util.py", line 23, in <module>
    from cassandra.pool import HostDistance
  File "/home/jaume/workspace/python-driver/cassandra/pool.py", line 31, in <module>
    from cassandra.connection import ConnectionException
  File "/home/jaume/workspace/python-driver/cassandra/connection.py", line 42, in <module>
    from cassandra.protocol import (ReadyMessage, AuthenticateMessage, OptionsMessage,
  File "/home/jaume/workspace/python-driver/cassandra/protocol.py", line 35, in <module>
    from cassandra.cqltypes import (AsciiType, BytesType, BooleanType,
  File "/home/jaume/workspace/python-driver/cassandra/cqltypes.py", line 771, in <module>
    class SetType(_SimpleParameterizedType):
  File "/home/jaume/workspace/python-driver/cassandra/cqltypes.py", line 774, in SetType
    adapter = util.sortedset
```

cassandra/metadata.py
bjmb: When the algorithm gets the list of replicas first it tries to return a replica with `HostDistance.LOCAL`, if this fails it gets in this if and just returns any replica.

bjmb:
This is the behavior *right now* which we might want to change:

The return of this function is a dict `keys_per_host` which looks like:

```python
keys_per_host = {
    <Host 127.0.0.1>: [key1, key2, key3, ...]
    <Host 127.0.0.2>: [key4, key5, key6, ...]
    .
    .
    .
}
```

for each key the following is done:

1) If there's a local replica (which is up) `valid_replicas` is formed only of local replicas (which are up)
2) Otherwise `valid_replicas` is formed of the rest of the other replicas (which are up)
3) If no replica is found that key is added to the `NO_VALID_REPLICA` and we are done for this key (go to step 1 with the next key)
4) Once we have `valid_replicas` to choose one of the replicas from `valid_replicas` we choose one which already is in the return dict `keys_per_host`. This way we maximize the grouping
5) If none of the replicas in `valid_replicas` is in `keys_per_host` we randomly select one of the replicas to group it to and start a list with that key for that replica.

PR #887, opened 2017-12-20
PYTHON-884 Rename PreparedStatement.column_metadata to bind_metadata

bjmb:


tests/unit/test_parameter_binding.py
```diff
@@ -80,11 +80,11 @@ class BoundStatementTestV3(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        column_metadata = [ColumnMetadata('keyspace', 'cf', 'rk0', Int32Type),
+        bind_metadata = [ColumnMetadata('keyspace', 'cf', 'rk0', Int32Type),
```
mambocab: Could you realign the rest of this list?

bjmb:
Yes, I think the rest of the `column_metadata` appearances are good as they are. When it is decoded in protocol.py it is [already named correcty|https://github.com/datastax/python-driver/blob/master/cassandra/protocol.py#L683].
I'm not sure if it's worth to be included in the upgrade line since it would be basically the same line that it's in the CHANGELOG, but I can add it if we go for documenting there this kind of changes.

docs/upgrading.rst
```diff
@@ -4,6 +4,15 @@ Upgrading
 .. toctree::
    :maxdepth: 1
 
+Upgrading to 4.0
+----------------
+
+Execution API Updates
+^^^^^^^^^^^^^^^^^^^^^
+:attr:`~.PreparedStatement.column_metadata` is renamed to :attr:`~.PreparedStatement.bind_metadata`
```
bjmb: Is this the correct way to reference the parameters? It looks [like this](https://github.com/datastax/python-driver/blob/ea00fa4020413656b6ecfe8982c0325ca2d1710e/docs/upgrading.rst#attrpreparedstatementcolumn_metadata-is-renamed-to-attrpreparedstatementbind_metadata) which doesn't look to good but the rest of the document is like that.

mambocab: Yeah, it's gonna look bad on github, unfortunately

PR #889, opened 2017-12-21
PYTHON-863: Remove copyright dates from copyright notices

aboudreault:


mambocab:
I've pushed a commit with a some notices that I think you missed -- feel free to remove that commit if they're wrong, and feel free to squash my changes into yours if they're right.

The LICENSE file also includes copyright years -- is this intentional? I didn't change that since I don't know if it's a special case.

PR #894, opened 2017-12-28
PYTHON-888: Support cassandra.query.BatchType with cqlengine BatchQuery

aboudreault:


mambocab:
Tests look clean, I'm +1

PR #896, opened 2017-12-28
truncate microsecond using replace function

Usui22750:
It appears that the previous implementation did not work, test locally

### DIFF
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -531,7 +531,7 @@ class DateTime(Column):
             return
         if isinstance(value, datetime):
             if DateTime.truncate_microseconds:
-                return value.replace(microsecond=0)
+                return value.replace(microseconds=0, milliseconds=value.microseconds // 1000)
             else:
                 return value
         elif isinstance(value, date):
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -531,9 +531,7 @@ class DateTime(Column):
             return
         if isinstance(value, datetime):
             if DateTime.truncate_microseconds:
-                us = value.microsecond
-                truncated_us = us // 1000 * 1000
-                return value - timedelta(microseconds=us - truncated_us)
+                return value.replace(microsecond=0)
             else:
                 return value
         elif isinstance(value, date):


mambocab:
You may have the wrong impression of the behavior we want -- we don't want to remove microseconds entirely. We want to truncate the `microseconds` attribute to be a multiple of 1000. This is to ensure the result of `to_python` equals the `timestamp` value stored in the database, which [only has millisecond resolution](http://docs.datastax.com/en/cql/3.3/cql/cql_reference/timestamp_type_r.html).

Usui22750:
Obviously, I have missed something, but to use the replace function is still better no?
maybe a fix that can be use:
return value.replace(microsecond=0, millisecond=value.microsecond // 1000 )

The comportment that I was expected is to always have the same output for the datetime.toisoformat function, that specifiy that you can have two different output if the microsecond are different from 0 or not. Can we  add a parameter like truncate_microseconds but for milliseconds?

mambocab:
> to use the replace function is still better no?

Yeah, I agree with that! Much cleaner and probably worth doing.

> The comportment that I was expected is to always have the same output for the datetime.toisoformat function, that specifiy that you can have two different output if the microsecond are different from 0 or not. 

I don't follow this part. `truncate_microseconds` exists to work around between the mismatched resolutions between `datetime` objects and the database. Adding the ability to ignore microseconds entirely doesn't fix anything -- it just loses information, e.g., by failing to differentiate `datetime.datetime(2018, 1, 2, 10, 1, 49, 400000)` from `datetime.datetime(2018, 1, 2, 10, 1, 49, 500000)`. The `.isoformat` method does return different representations, but only when truncation doesn't lose information.

mambocab:
If you'd like to submit an improvement for using `.replace`, please note that the method doesn't have a `millisecond` parameter and that the value we want is `1000 * (value.microsecond // 1000)`. (The tests should catch that, just wanted to make sure you didn't waste your time mesing with it.)

Usui22750:
For sure you are right,  we will handle this point with marshmallow. I will edit the code to use the replace function.

mambocab:
Thanks for the changes. A couple things:

- Note my comments above -- `.replace` doesn't have a `millisecond` parameter, and the value we want is `1000 * (value.microsecond // 1000)`, as the unit is microseconds.
- This change will require a changelog entry.

I can make these changes when we get to properly reviewing this if you like.

After discussing this with the team, we're going to delay making this change. It is more readable, but the existing implementation may be optimized for Cython compilation. We'll want to check that before merging, so we'll delay that to a later release.

PR #905, opened 2018-01-17
PYTHON-506: Remove two-phase query when deleting values

aboudreault:
This PR now let None pass along the UPDATE statements. Map key deletions is a bit different.. so some code has been preserve to avoid the creation of tombstone on every update.


mambocab:
Sorry for the long delay on review here, but: code looks good. I think the new tests, which passed when they ran, demonstrate that the changes work. +1 from me!

PR #908, opened 2018-01-22
PYTHON-901: minor improvement in ExecutionProfile and Metrics docs

aboudreault:


mambocab:
Builds as expected locally, +1

PR #912, opened 2018-01-29
PYTHON-848 ResultSet.was_applied doesn't support batch with LWT state…

bjmb:
…ments

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -4328,10 +4328,10 @@ class ResultSet(object):
         Only valid when one of the of the internal row factories is in use.
         """
         if self.response_future.row_factory not in (named_tuple_factory, dict_factory, tuple_factory):
-            raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factsory,))
+            raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factory,))
 
         is_batch_statement = isinstance(self.response_future.query, BatchStatement)
-        if is_batch_statement and self.column_names[0] != "[applied]":
+        if is_batch_statement and (not self.column_names or self.column_names[0] != "[applied]"):
             raise RuntimeError("No LWT were present in the BatchStatement")
 
         if not is_batch_statement and len(self.current_rows) != 1:
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -4321,13 +4321,20 @@ class ResultSet(object):
         """
         For LWT results, returns whether the transaction was applied.
 
-        Result is indeterminate if called on a result that was not an LWT request.
+        Result is indeterminate if called on a result that was not an LWT request or on
+        a :class:`.query.BatchStatement` containing LWT. In the latter case either all the batch
+        succeeds or fails.
 
-        Only valid when one of tne of the internal row factories is in use.
+        Only valid when one of the of the internal row factories is in use.
         """
         if self.response_future.row_factory not in (named_tuple_factory, dict_factory, tuple_factory):
             raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factsory,))
-        if len(self.current_rows) != 1:
+
+        is_batch_statement = isinstance(self.response_future.query, BatchStatement)
+        if is_batch_statement and self.column_names[0] != "[applied]":
+            raise RuntimeError("No LWT were present in the BatchStatement")
+
+        if not is_batch_statement and len(self.current_rows) != 1:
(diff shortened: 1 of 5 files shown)


cassandra/cluster.py
```diff
@@ -4321,13 +4321,19 @@ def was_applied(self):
...
+        Only valid when one of the of the internal row factories is in use.
         """
         if self.response_future.row_factory not in (named_tuple_factory, dict_factory, tuple_factory):
             raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factsory,))
-        if len(self.current_rows) != 1:
+
+        if isinstance(self.response_future.query, BatchStatement) and self.column_names[0] != "[applied]":
+            raise RuntimeError("No LWT were present in the BatchStatement")
+
+        if not isinstance(self.response_future.query, BatchStatement) and len(self.current_rows) != 1:
```
aboudreault: LGTM for the logic. It's not critical but I would suggest to remove the duplicate isinstance call here.
```python
is_batch_statement = isinstance(self.response_future.query, BatchStatement)
if is_batch_statement and self.column_names[0] != "[applied]":
        raise RuntimeError("No LWT were present in the BatchStatement")

if not is_batch_statement and len(self.current_rows) != 1:
    raise RuntimeError("LWT result should have exactly one row. This has %d." % (len(self.current_rows)))
```
or with a single if/else block.

cassandra/cluster.py
```diff
@@ -4321,14 +4321,17 @@ def was_applied(self):
...
+        Result is indeterminate if called on a result that was not an LWT request or on
+        a :class:`.query.BatchStatement` containing LWT. In the latter case either all the batch
+        succeeds or fails.
 
-        Only valid when one of tne of the internal row factories is in use.
+        Only valid when one of the of the internal row factories is in use.
         """
         if self.response_future.row_factory not in (named_tuple_factory, dict_factory, tuple_factory):
             raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factsory,))
-        if len(self.current_rows) != 1:
```
bjmb: I finally removed this because it was failing for the batch test. We can have a query string with a batch were more than one row fails.

aboudreault: As we discussed last week, I think we should avoid any runtime behavior change. I think that it is fine to assume the user *has* to use a BatchStatement and not a query string if he wants this was_applied to work. We at least provide a small improvement for that case.

We can also create a follow-up ticket about this to see what we should change in the next major release.

bjmb: > I think that it is fine to assume the user has to use a BatchStatement and not a query string

I'd think as well, unfortunately however in CQLEngine we are using a string for the batch statement so [this assert](https://github.com/datastax/python-driver/pull/912/files#diff-8d5a882a817cdaaeeb0ce18fb53ef274R158) was failing in the test, this was the reason I changed the logic in was_applied. I think this is a bug in CQLEngine now since the user may wrongfully believe the batch statement was applied but we can handle this in a follow up ticket as well.

If you are +1 I'll do the following: leave the logic originally agreed in was_applied, create a bug ticket for CQLEngine and a follow up ticket for the major release.

bjmb:
I've left the logic like it was originally and added a warning in the cqlengine docs. I'll squash the commits before merging.

cassandra/cluster.py
```diff
@@ -4321,13 +4321,20 @@ def was_applied(self):
...
-        Result is indeterminate if called on a result that was not an LWT request.
+        Result is indeterminate if called on a result that was not an LWT request or on
+        a :class:`.query.BatchStatement` containing LWT. In the latter case either all the batch
+        succeeds or fails.
 
-        Only valid when one of tne of the internal row factories is in use.
+        Only valid when one of the of the internal row factories is in use.
         """
         if self.response_future.row_factory not in (named_tuple_factory, dict_factory, tuple_factory):
             raise RuntimeError("Cannot determine LWT result with row factory %s" % (self.response_future.row_factsory,))
```
aboudreault: just noticed this existing typo in the code... can you fix in this PR? `self.response_future.row_factsory`

cassandra/cluster.py
```diff
@@ -4321,13 +4321,20 @@ def was_applied(self):
...
-        if len(self.current_rows) != 1:
+
+        is_batch_statement = isinstance(self.response_future.query, BatchStatement)
+        if is_batch_statement and self.column_names[0] != "[applied]":
```
aboudreault: Question, can we execute an empty batch statement? If so, will `self.column_names[0]` fail?

bjmb: Yeah, this fails .. good catch!

PR #918, opened 2018-02-12
[WIP] poc of extendable components

aboudreault:
This is a simple proof of concept how the python driver could have extendable components. It is based on the java-driver-4 work done by Olivier.

***This does not represent any formal API or final version of anything. It is only for test purpose and to visualize a little bit better how things could be in 4.x**

I don't think there is any critical blocker that would restrict the python-driver to preserve its current dse fork approach. We can end up with something very clean and the dse-driver package will be much easier to maintain. Of course, this will require an important refactoring from the driver internal and mostly a creation of a new dse-driver based on it. This will take a good amount of time to come up with the final oss 4.x and 3.x dse releases. Java team has started this months ago.

Since DSE 6.0 is about to be released, it is a good time to start that work and this is certainly a task I'm interested in. I think we will be able to collaborate when we have some base functionalities ready and definitively on the entire dse-driver. 

The high level tasks we need to come up with a proper API and default/oss implementation for:

- refactor of ProtocolVersion, addition of CoreProtocolVersion (Enum of ProtocolVersion)
- ProtocolVersionRegistry
- TypeCodecs
- TypeCodecsRegistry (no more magic with metaclass? or in addition to..)
- New feature: the DriverContext (don't take attention to the fact that the context is passed to many places in this PR.... we will need to come up with a better way... which might imply to leverave a bit less classmethods and rather lazy-instantiate protocol and codec classes with context bundled or something)
- Chainable RequestProcessors (dse, graph etc.)
- refactor the protocol/messages to support MessageCodecs per protocol version: the single Message with many if/else statements will be gone. Each protocol will have the responsibility to provide its codec if something changed. 
- ConsistencyLevel and ConsistencyLevelRegistry: a bit strange at first sight... but java decided to be future-proof and make sure this can be extended if DSE needs it. (+BatchType, WriteType?)
* This will also add many additional tasks that will be determined while working on it. e.g we'ill need a DseSession that is aware of continuous paging etc. still unknown at the moment.
* metadata

### DIFF
--- a/cassandra/__init__.py
+++ b/cassandra/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+from functools import total_ordering
 
 
 class NullHandler(logging.Handler):
@@ -24,6 +25,15 @@ logging.getLogger('cassandra').addHandler(NullHandler())
 
 __version_info__ = (3, 12, 0, 'post0')
 __version__ = '.'.join(map(str, __version_info__))
+
+
+class ProtocolConstants(object):
+
+    class Version(object):
+
+        V3 = 3
+        V4 = 4
+        V5 = 5
 
 
 class ConsistencyLevel(object):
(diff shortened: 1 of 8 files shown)


mambocab:
I like this approach a lot. I don't even mind the explicit context-passing too much, though I agree, we could definitely get around it by, e.g., making the context a property of the registry from which we get the `Message`.

Some less general points:

>TypeCodecsRegistry (no more magic with metaclass? or in addition to..)

What do you mean about metaclasses here?

> the single Message with many if/else statements will be gone. Each protocol will have the responsibility to provide its codec if something changed.

This is an awesome side benefit IMO.

Is there any more specific feedback you'd like?

aboudreault:
Thanks Jim for the feedback.  No, I wasn't waiting more feedback really since it is only to show a bit the direction where I would go. We will discuss in detail  on the final api and implementation during the work.

About the magic, this is what I am referring to: https://github.com/datastax/python-driver/blob/master/cassandra/cqltypes.py#L107 . It is not really an approach we can take on custom/extended components since it would modify also the *core* driver... and might also create issues if the custom components  that were overriding some type codecs are imported before the core driver etc.

PR #922, opened 2018-02-27
PYTHON-947: Add one() function to the ResultSet API 

aboudreault:
PR for master (3.14)


mambocab:
Done with the review for this PR -- added a couple tests, made a small change to the index check:

> Integer identity for small integers works in CPython, but isn't guaranteed in other interpreters.

PR #931, opened 2018-03-21
PYTHON-956: Make protocol messages pluggable

aboudreault:
I tried to keep things as simple as possible, so this is more a *prepare to be pluggable* until I introduce the DriverContext (which will be my next task). Currently, there is some temporary code that takes care of the codecs registration.

Initially, I wanted more changes on the design ... but have not been able to justify this enough. This PR contains changes on the encoder/decoder API of the protocol messages, so we will easily plug that in the DriverContext later.

What is a message encoder?
========================

An encoder has the responsibility of the message translation to the native protocol representation. It is basically a simple function. No class needed. We just need to register it per-protocol and per-opcode. Its signature is:
```python
def encode(f, message, protocol_version):
```
So, an message encoder can be one of the following:
* A simple python  function
* A class staticmethod
* A class classmethod

What is a message decoder?
========================

An decoder has the responsibility of the message translation from the native protocol representation to the driver one. It is basically a simple function. No class needed. We just need to register it per-protocol and per-opcode. Its signature is:
```python
def decode(f, protocol_version, *args):
```
So, an message decoder can be one of the following:
* A simple python  function
* A class staticmethod
* A class classmethod

Class Codec
==========

You can see in the code that when the decoder was a bit more complex (like for the ResultMessage), I've created a simple ResultMessage.Codec class and set the main decoder to the classmethod `decode`. The only reason was to keep things related to the decoding together and not pollute the main ResultMessage class scope. We can discuss if we should do that even when the encoder/decoder is a simple function.

FAQ
====
Q: If we register an encoder/decoder per protocol version, why do you need it in its signature?
A: The protocol_version param basically gives the context to the encoder/decoder. Without it, we would need to write an encoder per protocol version. Now, we keep things as before since it's relatively simple enough and we avoid that case. The only reason we will need to replace an encoder/decoder is if we are in another package (e.g dse). Note that this is also done that way in the java4 driver.


aboudreault:
I started by doing a BaseCodec class as you suggest. this was like:
```python
class BaseCodec(object):
    opcode = None

    @classmethod
    def decode(cls, f, protocol_version, user_type_map, result_metadata, *args):
        raise NotImplemented...

    @classmethod
    def encode(f, message, protocol_version):
        raise NotImplemented...
```

But, I realized that:

1- The BaseCodec class does mostly nothing at all
2- We usually have Request codec... and Response codec. So both are mostly not related.

We can discuss on slack about it.

aboudreault:
@bjmb Feel free to comment your preference for tests and this PR.

PR #933, opened 2018-03-23
PYTHON-343 SSL support for Twisted

bjmb:


cassandra/io/twistedreactor.py
```diff
@@ -222,7 +265,7 @@ def close(self):
             self.is_closed = True
 
         log.debug("Closing connection (%s) to %s", id(self), self.host)
-        self.connector.disconnect()
+        reactor.callFromThread(self.connector.disconnect)
```
mambocab: What made this `callFromThread` necessary? And maybe this should use `twisted.internet.blockingCallFromThread` -- I believe that the bare `disconnect` call would block, but I don't think this does.

bjmb: This is independent from the ssl addition. `disconnect` should be called from twisted event loop thread since it's not threasafe. We do the same in [asyncore](https://github.com/datastax/python-driver/blob/de86392bd68b479f52510f631cca08c5c2302f4c/cassandra/io/asyncorereactor.py#L373). I don't think there's anything wrong with not waiting, we just schedule it and it will happen.

bjmb:
Yeah we weren't passing `OpenSSL.SSL.VERIFY_PEER` in the tests which would've failed if that package had been installing correctly.

cassandra/io/twistedreactor.py
```diff
@@ -198,18 +208,51 @@ def add_connection(self):
...
+        if self.ssl_options:
+            try:
+                import OpenSSL.crypto
+            except ImportError:
+                raise ImportError(
+                    'pyOpenSSL must be installed to enable SSL support with the Twisted event loop'
+                )
+            from twisted.internet import ssl
+
+            class SSLContextFactory(ssl.ClientContextFactory):
```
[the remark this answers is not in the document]
bjmb: That class is defined inside `def add_connection(self):` so I guess it's private but I'll add a `_` to be clearer

mambocab:
Also, maybe the SSL tests don't belong in `long`; they're about a minute and a half long total locally.

cassandra/io/twistedreactor.py
```diff
@@ -198,18 +208,51 @@ def add_connection(self):
...
-            timeout=self.connect_timeout)
-
-    def client_connection_made(self):
+        if self.ssl_options:
+            try:
+                import OpenSSL.crypto
+            except ImportError:
+                raise ImportError(
+                    'pyOpenSSL must be installed to enable SSL support with the Twisted event loop'
+                )
```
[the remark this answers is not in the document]
mambocab: Yeah, I'm +1 on your fix :+1:

cassandra/io/twistedreactor.py
mambocab: Got it :+1: sgtm

cassandra/io/twistedreactor.py
```diff
@@ -211,33 +211,50 @@ def add_connection(self):
...
-                    'pyOpenSSL must be installed to enable SSL support with the Twisted event loop'
+                    str(e) +
+                    ', pyOpenSSL must be installed to enable SSL support with the Twisted event loop'
                 )
             from twisted.internet import ssl
 
-            class SSLContextFactory(ssl.ClientContextFactory):
-                def __init__(self, ssl_options):
-                    self.ssl_options = ssl_options
+            class _SSLContextFactory(ssl.ClientContextFactory):
```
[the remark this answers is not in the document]
bjmb: :+1: right got it now and I agree

cassandra/io/twistedreactor.py
```diff
@@ -211,33 +211,50 @@ def add_connection(self):
...
+                def __init__(self_context, ssl_options):
+                    self_context.ssl_options = ssl_options
 
-                def getContext(self):
+                def getContext(self_context):
```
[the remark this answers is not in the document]
bjmb: It's because it was defined method and it was overriding the `self` from the method but since it's going to be moved out I'll rename it.

bjmb:
They don't take too long (around 7 minutes) but they are very few tests:
* long/test_ssl.py: Ran 7 tests in 120.567s
* standard/test_cluster.py: Ran 38 tests in 71.959s

I don't have strong feelings though, we can add 2 minutes to the short tests.

tests/integration/__init__.py
```diff
@@ -111,6 +111,8 @@ def _get_dse_version_from_cass(cass_version):
 KEEP_TEST_CLUSTER = bool(os.getenv('KEEP_TEST_CLUSTER', False))
 SIMULACRON_JAR = os.getenv('SIMULACRON_JAR', None)
 
+default_cassandra_version = '3.11'
```
mambocab: This duplicates [code a couple lines down](https://github.com/datastax/python-driver/pull/933/files#diff-437a31ded249825f4640cfd1c0608bf8R119) -- is this intentional?

test-requirements.txt
```diff
@@ -14,3 +14,4 @@ eventlet
 cython>=0.20,<0.28
 packaging
 asynctest; python_version > '3.4'
+pyOpenSSL
```
mambocab: We actually might be able to get away with `twisted[tls]` here, and maybe we should recommend that to users. (this knowledge earned yesterday while trying to get the duration tests to work :P)

tests/integration/__init__.py
bjmb: Nope, good catch

bjmb:
Awesome thank you for the review!

PR #936, opened 2018-03-27
PYTHON-940 Log warning when driver configures an authenticator, but s…

bjmb:
…erver does not request authentication


mambocab:
+1 after the Travis tests run clean with a changelog entry, looks good!

aboudreault:
LGTM Jim!

mambocab:
Code looks good. I've kicked off a new run of the tests, since it looks like something went wrong last time.

Once the tests run clean, I'm +1 here. Once it's merged to DSE, we should add a test for `DSEGSSAPIAuthProvider`; I'll add a subticket on JIRA.

bjmb:
:+1  test running clean thanks for the reviews, merging.

PR #938, opened 2018-04-04
PYTHON-966: Allow filter queries with fields that have an index managed outside of cqlengine

aboudreault:
This PR will require some DSE tests with a dse-search index.


mambocab:
I'm +1, on the assumption that [you can't create indexes on counter columns](https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useCountersConcept.html). Looks good!

PR #940, opened 2018-04-10
Added tests for PYTHON-966

bjmb:


tests/integration/cqlengine/query/test_queryset.py
```diff
@@ -754,15 +754,28 @@ def test_custom_indexed_field_can_be_queried(self):
         with self.assertRaises(query.QueryException):
             list(CustomIndexedTestModel.objects.filter(data='test'))  # not custom indexed
 
+        with self.assertRaises(QueryException):
```
bjmb: This assertions is failing, I have my doubts about what it should raise. Before it raised `QueryException`, but now it's raising `InvalidRequest`. I thought it'd be `QueryException` because we can detect it'll fail from cqlengine

PR #941, opened 2018-04-11
PYTHON-697 asyncore reactor should use a global variable instead of a class v…

bjmb:
… class variable for the event loop

### DIFF
--- a/tests/integration/standard/test_connection.py
+++ b/tests/integration/standard/test_connection.py
@@ -20,12 +20,13 @@ except ImportError:
 from functools import partial
 from six.moves import range
 import sys
+import threading
 from threading import Thread, Event
 import time
-import weakref
 
 from cassandra import ConsistencyLevel, OperationTimedOut
 from cassandra.cluster import NoHostAvailable, ConnectionShutdown, Cluster
+import cassandra.io.asyncorereactor
 from cassandra.io.asyncorereactor import AsyncoreConnection
 from cassandra.protocol import QueryMessage
 from cassandra.connection import Connection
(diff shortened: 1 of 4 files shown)


build.yaml
```diff
@@ -2,7 +2,7 @@ schedules:
   commit_master:
     schedule: per_commit
     branches:
-      include: [master]
+      include: [master, python-697]
```
bjmb: We should revert this line if merged

tests/integration/standard/test_connection.py
```diff
@@ -401,6 +402,28 @@ def setUp(self):
...
+        class C1(AsyncoreConnection):
+            pass
+
+        class C2(AsyncoreConnection):
+            pass
+
+        cassandra.io.asyncorereactor._global_loop._cleanup()
+        cassandra.io.asyncorereactor._global_loop = None
+
+        clusterC1 = Cluster(connection_class=AsyncoreConnection)
```
mambocab: Should this be `connection_class=C1`, and the one below be `connection_class=C2`?

bjmb: Yeah, good catch :+1:

bjmb: Hmm, I just realized with test happens the same with the libev one, it doesn't fail on master. Still is a good test to have IMO, wdyt?

bjmb: Sorry :worried:  , I had forgot to push, this only makes sense if `connection_class=C1` of course. 
By the libev test I was referring to my last comment here: https://github.com/datastax/python-driver/pull/950 . This test is not failing in master because somewhere before this test is ran we call `cluster = Cluster(connection_class=AsyncoreConnection)` and `C1` and `C2` use `AsyncoreConnection`'s loop, instead of creating one each. (It's better exapined in the libev PR)

mambocab:
Code looks good, both here and on #950. I've started the tests again though -- looks like there was a failure to install `mock`?

PR #942, opened 2018-04-12
PYTHON-968: Support IS NOT NULL operator in cqlengine 

bjmb:


tests/integration/cqlengine/operators/test_where_operators.py
```diff
@@ -48,3 +59,56 @@ def test_operator_rendering(self):
         self.assertEqual("<=", six.text_type(LessThanOrEqualOperator()))
         self.assertEqual("CONTAINS", six.text_type(ContainsOperator()))
         self.assertEqual("LIKE", six.text_type(LikeOperator()))
+
+
+class TestIsNotNull(BaseCassEngTestCase):
+    def test_is_not_null_to_cql(self):
```
mambocab: nit: Could this be a unit test?

bjmb: It definitely can, but we don't have unit tests for cqlengine, so I figured at some point we didn't want to create a new folder just for a few tests. They don't start C* so they don't add more time to the run for being there.
If we moved them to unit I'd do so in a future PR

bjmb:
They track count of how many parameters have been processed because the way cqlengine builds the queries is:
`SELECT clolumn1, column2, ... from keyspace.table WHERE A = %(0)s AND  B = %(1)s`, `parameters = {0: "parameter for A", 1: "parameter for B"}`. 
Using 0, 1, 2, ... to track the parameters.

So each clause need to know how many parameters the clauses from before have taken so they can put the "0" or the "1" or the number that it corresponds to them. Since `IS NOT NULL` uses zero `get_context_size` returns 0 and in `update_context` the context is not updated since it doesn't affect the global count nor uses parameters.

tests/integration/cqlengine/operators/test_where_operators.py
mambocab: :+1: awesome. filed a low-priority ticket: https://datastax-oss.atlassian.net/browse/PYTHON-975

mambocab:
Ohhh, cool. Thanks for the explanation.

mambocab:
Looks great, +1 for merging.

bjmb:
Awesome, thank you for the review!

docs/cqlengine/queryset.rst
```diff
@@ -199,6 +199,14 @@ In addition to simple equal to queries, cqlengine also supports querying with ot
 
         q = Automobile.objects.filter(model__like='%Civic%').allow_filtering()
 
+:attr:`IS NOT NULL (IsNotNull(column_name)) <query.QueryOperator.IsNotNullOperator>`
```
aboudreault: This should probably point to the Clause and not the Operator. In this case, I think the Operator can't be used directly.

bjmb: That's correct, the clause has to be used for this case. I'll change it

PR #943, opened 2018-04-13
Wait when checking the pool status in tests

bjmb:


mambocab:
I've pushed a small change that consolidates the methods using an optional kwarg and cleans up some whitespace. Feel free to squash it all together.

Is it a problem that the pool state stabilizes after `execute` returns? I think it's not, since the only downside is that it might reject potentially-valid `execute()` calls, and the upside is that we don't block on pool state stabilization unnecessarily.

bjmb:
Yeah I thought the same but I don't think it's  a problem since that's transparent to the user. 

I'd consider renaming `request_ids` to `_request_ids` for 4.0, I don't see why the user might want to access this. But it's true that there is a time window in which a request_id has landed and not yet returned to the queue. By the way I've realized that this only happens in dse (the tests were only failing in dse but I thought it was a common problem). There's an [explanation in the code for this](https://github.com/riptano/python-dse-driver/blob/master/dse/connection.py#L780). Still I'm not sure if this should be merged here, wdyt?

mambocab:
> Yeah I thought the same but I don't think it's a problem since that's transparent to the user.

Cool, thanks for confirming.

> I'd consider renaming request_ids to _request_ids for 4.0, I don't see why the user might want to access this.

That's a reasonable idea -- it's kind of hard to explain, and I agree that if you're messing with it, you're doing so at your own risk. [Filed a ticket for follow-up discussion/decision-making](https://datastax-oss.atlassian.net/browse/PYTHON-980).

> Still I'm not sure if this should be merged here, wdyt?

By "this" do you mean "this PR"? This PR seems fine to me.

PR #944, opened 2018-04-18
Token range ownership

damien-instaclustr:
According to Cassandra:

> A Range is responsible for the tokens between (left, right]. **

However, `TokenMap.get_replicas()` appears to implement [left, right).

This pull request aims to resolve this disagreement.

_** Source: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/dht/Range.java_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,12 @@
+3.15.0
+======
+Bug Fixes
+---------
+* Tokenmap.get_replicas returns the wrong value if token coincides with the end of the range (PYTHON-978)
+
 3.14.0
 ======
+April 17, 2018
 
 Features
 --------
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from binascii import unhexlify
-from bisect import bisect_right
+from bisect import bisect_left
 from collections import defaultdict, Mapping
 from functools import total_ordering
 from hashlib import md5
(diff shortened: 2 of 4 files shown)


bjmb:
This looks good to me, also add that the values in `self.ring` correspond to the end of the token range up to and including the value listed.

damien-instaclustr:
Thanks @bjmb 
I've updated the pull request to include the comment.

bjmb:
Thank you @damien-instaclustr , can you verify I've correctly explained the issue in the ticket: https://datastax-oss.atlassian.net/browse/PYTHON-978? Feel free to add something otherwise.

damien-instaclustr:
Thanks @bjmb for creating the ticket. I tried to create a ticket before submitting the PR, but wasn't able to because I was hitting the jira's account limit. Looks like the limit will also prevent me from directly contributing to the ticket aswell.
However, the description looks good to me.

bjmb:
Thank you @damien-instaclustr , added the changelog entry

damien-instaclustr:
Yes. It looks like there was another test I missed that assumed the token range ownership was [left, right).
I have pushed a change to the test which should correct this.

@mambocab - Do you mind reviewing again?

mambocab:
Great, thanks! I've rebased my other branch onto `master` to bring in some test fixes (see #948). Could you squash this history down to one commit? If the tests run clean, I'll merge. (I can also squash on my end; just let me know.)

mambocab:
Tests ran clean!

damien-instaclustr:
Thanks @mambocab. I have squashed my commits.

mambocab:
Many thanks!

PR #945, opened 2018-04-19
Fix unit tests

aboudreault:
tests was passing... but there was errors in logs


bjmb:
As discussed we'll get this in when we merge from master

PR #949, opened 2018-04-24
Read data from the Cassandra table which is updating realtime

aliartiza75:
The current version of the python-driver reads the data from the cassandra table till a particular row but what if the table is updating in real-time, so in order to read the new data from table we have to go through all the rows again.

For example:

A **USER** table have 100 rows in it. So using the script below:
```python
self.session.default_fetch_size = 10
result_set = self.session.execute("SELECT * FROM testKeyspace.USER")
    while(result_set.has_more_pages):
        for row in result_set.current_rows:
            print(row)
        page_state = result_set.paging_state
        result_set = self.session.execute("SELECT * FROM test3rf.test", paging_state=page_state)
```
we will be able to read the cassandra table till 100th. After some time the **USER** table gets updated with two new rows. So in order to read the two new rows we have to go through all the other 100 rows which is inefficient. 

So In this PR I will add a script, which will be able to read the data from the cassandra in real time.


mambocab:
https://github.com/datastax/python-driver/pull/949

Hi @aliartiza75, and thank you for the report. There are a couple procedural issues with this PR:

- New code should target `master`, `4.x`, or both; this PR proposes a merge from `4.x` into `master`, which we won't do until around 4.0's release.
- This PR adds no new code. If you're proposing a new feature, please [submit a ticket on JIRA](https://datastax-oss.atlassian.net/browse/PYTHON) describing the behavior you want. If you'd like help using JIRA, feel free to find me (username `mambocab`) on the [DataStax Academy Slack](https://academy.datastax.com/slack).

I should note a couple things about the behavior you're describing as well:

- The existing behavior you describe (results are set at a single point in time, paging through results returns the results fixed at that time) is intended. To our knowledge, it's standard behavior as well; we found similar behavior in Postgres.
- I don't know that C* supports the behavior you'd like, at least not as part of the normal execution API. This could be possible with CDC (for which there is no Python client at this time), or you could approximate it with clever use of timestamps. In your JIRA ticket, could you go into a little more detail into how you plan to implement this feature? 

Thanks!

PR #950, opened 2018-05-14
PYTHON-973: Use global variable for libev loops so it can be subclassed

bjmb:
This PR is very similar to https://github.com/datastax/python-driver/pull/941, so some of the review comments there might apply here as well @mambocab

### DIFF
--- a/tests/unit/io/test_libevreactor.py
+++ b/tests/unit/io/test_libevreactor.py
@@ -35,7 +35,6 @@ class LibevConnectionTest(ReactorTestMixin, unittest.TestCase):
 
     connection_class = LibevConnection
     socket_attr_name = '_socket'
-    loop_attr_name = '_libevloop'
     null_handle_function_args = None, 0
 
     def setUp(self):
@@ -62,7 +61,7 @@ class LibevConnectionTest(ReactorTestMixin, unittest.TestCase):
         Test for asserting that watchers are closed in LibevConnection
 
         This test simulates a process termination without calling cluster.shutdown(), which would trigger
-        LibevConnection._libevloop._cleanup. It will check the watchers have been closed
+        _global_loop._cleanup. It will check the watchers have been closed
         Finally it will restore the LibevConnection reactor so it doesn't affect
         the rest of the tests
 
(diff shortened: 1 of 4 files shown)


tests/unit/io/test_libevreactor.py
```diff
@@ -72,24 +71,25 @@ def test_watchers_are_finished(self):
 
         @test_category connection
         """
-        with patch.object(LibevConnection._libevloop, "_thread"),\
-             patch.object(LibevConnection._libevloop, "notify"):
+        from cassandra.io.libevreactor import _global_loop
```
bjmb: Seems like the import has to happen here so the it happens after `_global_loop` is initialized, if it's imported at the beginning, `_global_loop` is `None`

tests/unit/io/test_libevreactor.py
```diff
@@ -125,7 +125,8 @@ def create_timer(self):
 
     @property
     def _timers(self):
-        return self.connection._libevloop._timers
+        from cassandra.io.libevreactor import _global_loop
```
bjmb: Same here

mambocab:
Code looks good! 2 things:

- I should have asked this earlier, but could you add a basic test of the subclassing behavior?
- Any idea what's up with `pypy` on Travis? I can look into it if you like, I just haven't seen it until now.

bjmb:
It's hard to add in a test that fails on master and passes here because of the following:
## Case 1
```
class C1(LibevConnection):
    pass


class C2(LibevConnection):
    pass

clusterC1 = Cluster(connection_class=C1)
session1 = clusterC1.connect(wait_for_all_pools=True)
clusterC2 = Cluster(connection_class=C2)
session2 = clusterC2.connect(wait_for_all_pools=True)

# Two threads in master now, 1 in this PR
```
## Case 2
```
class C1(LibevConnection):
    pass


class C2(LibevConnection):
    pass

cluster = Cluster(connection_class=LibevConnection)
session = clusterC1.connect(wait_for_all_pools=True)

clusterC1 = Cluster(connection_class=C1)
session1 = clusterC1.connect(wait_for_all_pools=True)
clusterC2 = Cluster(connection_class=C2)
session2 = clusterC2.connect(wait_for_all_pools=True)

# 1 thread in both master and this PR. Because the subclasses use the father class loop
```

In our tests in `tests/integration/__init__.py` we do  `cluster = Cluster(connection_class=LibevConnection)`
Therefore the subclasses will use the event loop from the father.

I think we can test that `_global_loop` works fine but it's harder to test that it doesn't create more than one event loop.

As for the pypy issue looks like gevent is not being installed in pypy (this is happening in master as well). There was a release 4 days ago so it's probably related to this.

bjmb:
Yeah that's correct. That stops the libev loop and the tests verifies:
* If there's NO event loop -> Two subclasses are used in different clusters -> Only one new event loop is created

Before:

* If there's NO event loop -> Two subclasses are used in different clusters -> Two new event loop is created

But before also happened:

* If there's an event loop -> Two subclasses are used in different clusters -> Only ONE new event loop is created

bjmb:
I'll add a similar test to the asyncore one here

mambocab:
Rebased c8728b9aa05d741d6a88207ba3d35b9a69139d55 onto `master` at 29e7f8fd9ae5b5214c9bf2dc76a7eb12ff09a501.

mambocab:
Tests ran clean using libev; merging.

PR #951, opened 2018-05-17
PYTHON-893: fallback implementation for namedtuple row factory

mambocab:
[PYTHON-893](https://datastax-oss.atlassian.net/browse/PYTHON-893).

As mentioned on that ticket, we've decided it's better to degrade to a pseudonamedtuple (with a warning) than to simply fail on very long lists of columns.

### DIFF
--- a/cassandra/query.py
+++ b/cassandra/query.py
@@ -25,6 +25,7 @@ import struct
 import time
 import six
 from six.moves import range, zip
+import warnings
 
 from cassandra import ConsistencyLevel, OperationTimedOut
 from cassandra.util import unix_time_from_uuid1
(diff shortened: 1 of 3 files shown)


aboudreault:
LGTM Jim!

cassandra/query.py
```diff
@@ -83,6 +84,29 @@ def tuple_factory(colnames, rows):
...
+    def __getattr__(self, name):
+        return self._dict[name]
+
+    def __getitem__(self, idx):
+        return self._tuple[idx]
+
+
+def pseudo_namedtuple_factory(colnames, rows):
+    """
+    """
```
aboudreault: I'm, wondering if we really wants to make this public ... :thinking:

aboudreault: What's the benefit to suggest users to use this instead of the dict_factory if they have the issue?

cassandra/query.py
```diff
@@ -83,6 +84,29 @@ def tuple_factory(colnames, rows):
...
+    def __init__(self, ordered_dict):
+        self._dict = ordered_dict
+        self._tuple = tuple(ordered_dict.values())
+
+    def __getattr__(self, name):
+        return self._dict[name]
+
+    def __getitem__(self, idx):
+        return self._tuple[idx]
+
```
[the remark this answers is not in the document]
mambocab: oops, yep, should definitely do that.

cassandra/query.py
mambocab: This isn't given as a suggestion; [it's a fallback row factory](https://github.com/datastax/python-driver/pull/951/files#diff-6f3180b25cd5f084c6478485d9ca403eR156). I agree, there isn't any benefit to making this "public".

cassandra/query.py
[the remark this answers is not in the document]
mambocab: Fixed -- added `__iter__`.

cassandra/query.py
mambocab: Thoughts on this bit?

aboudreault:
LGTM Jim yep!

mambocab:
Thanks for the review! I've added a docstring for the new factory and sqashed. Will merge after a clean test run :+1:

mambocab:
Integration tests ran clean, travis failures will be addressed by the 3.3 removal recently merged to master. The Appveyor failures are known flakes.

PR #952, opened 2018-05-18
Fix LZ4 test

beltran:


bjmb:
As @aholmberg noticed, this was failing for older version of lz4 like `0.7.0`

mambocab:
Do we want the duplication here? Would it work to `from cassandra.connection import lz4_block`?

mambocab:
Another way we could deal with this:

https://github.com/datastax/python-driver/tree/replace-lz4-test

The benefit here is that the test and the dereference are right next to each other, so if one gets changed it'll be easier to remember to change the other.

mambocab:
To be clear: I think we could do that and remove the test in its entirety.

bjmb:
I like more your branch and removing the test. But I'd log something and then default to non-compression (like we do if lz4 is not installed) instead of raising an error.

mambocab:
Good idea, sounds like a plan.

PR #953, opened 2018-05-19
tox: remove 3.3

greut:
3.3 has been removed from Travis, and it could start running the tests on 3.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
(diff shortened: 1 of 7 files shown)


greut:
Let's hope lz4 is handled by #952

bjmb:
Just seen that py3.3 is also mentioned in ./docs/index.rst, ./docs/installation.rst, ./README.rst and ./README-dev.rst (and I think that's it!)

I gave it a shot before but run into several difficulties, lz4,  PyYaml were failing but can be skipped. But also nose was failing to get installed and it's harder to get around that. I believe https://github.com/datastax/python-driver/pull/952 is a different issue. 

I'd suggest opening different PRs for py3.3 and py3.7. Personally, I'd rather not have a build where we can't even install the dependencies.

greut:
> I'd suggest opening different PRs for py3.3 and py3.7

fair enough.

mambocab:
Thanks @greut, much appreciated! Merging.

PR #955, opened 2018-05-22
tox: enable python 3.7

greut:
Enables 3.7-dev tests on travis

also removes PyYAML which is not used by anyone and fails on 3.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
+dist: xenial
+sudo: false
+
 language: python
 python:
   - "2.7"
-  - "3.4"
   - "3.5"
   - "3.6"
-  - "pypy-5.4.1"
+  - "3.7"
+  - "pypy2.7-6.0"
+  - "pypy3.5"
 
-sudo: false
 env:
   - CASS_DRIVER_NO_CYTHON=1
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,10 @@
 [tox]
-envlist = py{27,34,35,36},pypy
+envlist = py{27,34,35,36,37},pypy
+skip_missing_interpreters = True
 
 [base]
 deps = nose
        mock<=1.0.1
-       PyYAML
        six
        packaging
        cython
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,7 +4,6 @@ nose
 mock!=1.1.*
 ccm>=2.0
 unittest2
-PyYAML
 pytz
 sure
 pure-sasl


greut:
`dist: xenial` has py3.7 but no pypy...

aboudreault:
Just linked your PR in https://datastax-oss.atlassian.net/browse/PYTHON-1016

boredstiff:
@greut Thanks for working to get this going on 3.7, definitely looking forward to using it.

agrybas:
Any idea when Python 3.7 support will be added? Thanks

aboudreault:
@agrybas FYI, python 3.7 should work fine. But I will put this in priority for the next release.

agrybas:
> @agrybas FYI, python 3.7 should work fine. But I will put this in priority for the next release.

We run into issues connecting to the cluster when using Python 3.7. The code hangs on the connect() call. 

When using behind gunicorn, we've noticed occasional timeouts - the client raises OperationTimedOut exception with the following error: `Client request timeout. See Session.execute[_async](timeout)` No errors/timeouts are reported on the database side which leads us to believe the driver is at fault.

We're running Apache Cassandra 3.11.4, we tried cassandra-driver versions 3.12 through 3.16. The same code works fine with Python 3.6 & cassandra-driver 3.12.

Thank you for looking into this!

aboudreault:
Will do. Also make sure you are using the latest version (3.17.1).

aboudreault:
Hello @greut, I'm about to merge this PR.... could you limit it to Python 3.7? (no 3.8)

.travis.yml
```diff
@@ -1,12 +1,19 @@
...
+
 language: python
 python:
   - "2.7"
-  - "3.4"
   - "3.5"
   - "3.6"
-  - "pypy-5.4.1"
+  - "3.7"
+  - "3.7-dev"
```
aboudreault: can we remove this?

.travis.yml
```diff
@@ -1,12 +1,19 @@
+dist: xenial
+sudo: false
+
 language: python
 python:
   - "2.7"
-  - "3.4"
   - "3.5"
   - "3.6"
-  - "pypy-5.4.1"
```
aboudreault: Any reason we remove 3.4 and pypy 5.4.1?

.travis.yml
```diff
@@ -1,12 +1,19 @@
...
+  - "3.7"
+  - "3.7-dev"
+  - "pypy3.5"
+
+matrix:
+  allow_failures:
+    -  python: "3.7-dev"
```
aboudreault: ^ same than above

tox.ini
```diff
@@ -1,10 +1,10 @@
 [tox]
-envlist = py{27,34,35,36},pypy
+envlist = py{27,35,36,37,38},pypy
```
aboudreault: remove 38 here

tox.ini
```diff
@@ -1,10 +1,10 @@
...
+skip_missing_interpreters = True
```
aboudreault: I guess we can remove this as well?

.travis.yml
greut: Because of Travis' support (also 3.4 has been EOL'd a while ago)

tox.ini
greut: it won't hurt.

.travis.yml
aboudreault: Python 3.4 is not available anymore with that xenial image? It is EOL since march, but the python driver still support it until we decide to remove it (next major probably).

tox.ini
aboudreault: So if an interpreter is removed in Travis, this will make our test suite to succeed silently? Is my understanding correct?

.travis.yml
greut: IIRC that's correct xenial doesn't have it.

greut: aha, and xenial doesn't have 3.7 :wink:

tox.ini
greut: Travis fails before that. https://travis-ci.com/datastax/python-driver/jobs/210891608

As a developer, it won't complain if you don't install all the supported interpreters though.

.travis.yml
aboudreault: Alright... so let's keep xenial and 3.7 then.

.travis.yml
```diff
@@ -1,12 +1,19 @@
...
+
 language: python
 python:
   - "2.7"
-  - "3.4"
   - "3.5"
   - "3.6"
-  - "pypy-5.4.1"
+  - "3.7"
+  - "pypy"
```
aboudreault: Can you try `pypy2.7` ? If it still doesn't work, this one should work I think: `pypy2.7-6.0`

.travis.yml
```diff
@@ -1,12 +1,19 @@
...
+  - "pypy3.5"
+
+matrix:
+  allow_failures:
+    - pypy
```
aboudreault: We will be able to remove this when my previous comment is fixed.

.travis.yml
aboudreault: https://travis-ci.community/t/confusing-pypy-versions/2505

aboudreault:
Thanks @greut !

boredstiff:
Thanks, was looking forward to this, starting a new project this week and this will allow me to use 3.7.

If we're fully Python3.7 ready, might want to update the README as well, it still ends at 3.6 and still lists 3.4.

PR #956, opened 2018-05-22
Add tests for customizing context

aboudreault:
This currently fails... but will be fixed in PYTHON-954


aboudreault:
Thanks Jaume for the tests!

PR #957, opened 2018-05-23
PYTHON-954: Make ProtocolVersion pluggable

aboudreault:


cassandra/registry.py
```diff
@@ -12,17 +12,87 @@
...
 
 from cassandra import ProtocolVersion
 from cassandra.protocol import *
 
 
+class ProtocolVersionRegistry(object):
+    """Default implementation of the ProtocolVersionRegistry"""
+
+    # default versions and support definition
+    protocol_version = ProtocolVersion
```
aboudreault: The ProtocolVersionRegistry keeps a trace of the ProtocolVersion support class for the driver purpose. In DSE, this will be set to DseProtocolVersion which additional versions and support checks. I don't this  the `ProtocolVersion` naming is accurate but kept it only because we had that before.

cassandra/protocol.py
```diff
@@ -129,25 +128,25 @@ class Codec(object):
         def __init__(self, error_decoders):
             self.error_decoders = {opcode: decoder for opcode, decoder in error_decoders}
 
-        def decode(self, f, protocol_version, *args):
+        def decode(self, f, protocol_version, *args, **kwargs):
```
mambocab: Here and elsewhere: why add `**kwargs` over a `protocol_version_registry` kwarg?

cassandra/registry.py
aboudreault: oh that was just a personal thought. I just feel a bit confusing that we use `protocol_version` everywhere in the code, which is always an integer... but that this object is called a `ProtocolVersion`... but it's only some support checks. Anyway.. it's not something I'm going to touch.

cassandra/registry.py
```diff
@@ -12,17 +12,84 @@
...
+    """A list of registered beta protocol versions"""
+
+    _supported_versions = None
+    """A list of supported protocol versions"""
+
+    def __init__(self, protocol_versions, beta_versions=None):
+        self._supported_versions = sorted(protocol_versions, reverse=True)
+        self._beta_protocol_versions = tuple(beta_versions or [])
+
+    def supported_versions(self):
```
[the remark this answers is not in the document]
aboudreault: I guess no arguments are expected, so yeah... will change that to properties.

cassandra/registry.py
```diff
@@ -12,17 +12,84 @@
...
+        Return the maximum protocol version supported by this driver.
+        """
+        return max(self.supported_versions())
+
+    def get_lower_supported(self, previous_version):
+        """
+        Return the lower supported protocol version. Beta versions are omitted.
+        """
+        try:
+            version = next(v for v in sorted(self.supported_versions(), reverse=True) if
```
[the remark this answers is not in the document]
aboudreault: I haven't touched to that code... will check if there is any  reason for the next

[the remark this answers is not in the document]
mambocab: Oh, sorry, missed that. Don't feel obligated to mess with it.

cassandra/protocol.py
aboudreault: It was mostly for the futur flexibility, but I have no objection to switch back to **kwargs if we need to if we hit a case.

mambocab: That sounds like a plan!

cassandra/registry.py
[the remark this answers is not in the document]
aboudreault: After some tests, I keep `next` here. We would have to use max for this use case which doesn't makes think clearer. And next handle well an empty sequence when no more version is available.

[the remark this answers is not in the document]
mambocab: > And next handle well an empty sequence when no more version is available.

Got it, perfect. Thanks for having a look.

tests/unit/test_connection.py
```diff
@@ -43,7 +43,7 @@ def make_connection(self):
         return c
 
     def make_header_prefix(self, message_class, stream_id=0,
-        version=driver_context.protocol_version_registry.max_non_beta_supported()):
+        version=driver_context.protocol_version_registry.max_non_beta_supported:
```
mambocab: Syntax error, right?

PR #958, opened 2018-05-30
Update PyPI URLs for new Python Package Index

mshuler:


mambocab:
Awesome, thanks for the PR!

PR #964, opened 2018-06-11
PYTHON-957: Make Types pluggable

aboudreault:
Important changes in this PR:

* There is no more mandatory `context` in various class APIs. `context` is now a kwarg  defaulted to None. When a component requires the context and context is None... it automatically get the new `DefaultDriverContext`

* The message protocol codec has been refactored based on our discussion. A Codec is now *always* a class that need to be instantiated at startup (done in the registry factory)

* Note that CQLEngine doesn't support driver context yet... it is hard-coded to DefaultDriverContext. there is a follow up ticket about that.


cassandra/protocol.py
```diff
@@ -122,32 +127,35 @@ def to_exception(self):
...
     error_code = None
 
-    @staticmethod
-    def decode_error_info(f, protocol_version, **kwargs):
-        pass
+    class Codec(object):
+        error_class = None
+
+        def __init__(self, error_class):
+            self.error_class = error_class
```
aboudreault: This is because I can't reference the outer class :(. We could also patch the class *after* it's definition with something like `MyErrorClass.Codec.error_class = MyErrorClass`

mambocab:
Looks like there are a couple tests that need to be fixed though.

cassandra/protocol.py
aboudreault: I like this approach Jim. +1 to apply it.

cassandra/context.py
```diff
@@ -47,17 +44,27 @@ def __call__(self):
...
 
     _protocol_version_registry = None
+    _type_registry = None
     _message_codec_registry = None
     # the default protocol handler
     _protocol_handler = None
 
     def __init__(self):
-        self._protocol_version_registry = SingletonProvider(ProtocolVersionRegistry.factory)
+        from cassandra.protocol import ProtocolHandler
```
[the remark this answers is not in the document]
aboudreault: I will add a comment. It is due to the DefaultDriverContext provider import in a few places.

cassandra/connection.py
```diff
@@ -258,19 +259,19 @@ class Connection(object):
...
+                 allow_beta_protocol_version=False, no_compact=False, context=None):
         self.host = host
         self.port = port
         self.authenticator = authenticator
         self.ssl_options = ssl_options.copy() if ssl_options else None
         self.sockopts = sockopts
         self.compression = compression
         self.cql_version = cql_version
-        self._context = context
+        self._context = context or DefaultDriverContext()
```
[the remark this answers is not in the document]
aboudreault: Exact

aboudreault:
@mambocab What tests are you referring to?

mambocab:
I was talking about the ones that failed here: http://jenkins-devtools.datastax.lan/job/datastax.python-driver.python-957.commit_branches/14/testReport/

aboudreault:
Those 2 failures are not related to this work... again unstable tests.

cassandra/protocol.py
```diff
@@ -122,32 +127,40 @@ def to_exception(self):
...
             if error_decoder:
-                return error_decoder(f, protocol_version, msg, **kwargs)
+                return error_decoder.decode(f, protocol_version, msg)
             return ErrorMessage(code=code, message=msg, info=None)
 
 
+class ErrorMessageSubTypeWithCodecMeta(type):
+    def __init__(cls, name, bases, clsdict):
+        super(ErrorMessageSubTypeWithCodecMeta, cls).__init__(name, bases, clsdict)
+        cls.Codec = type("Codec", (cls.Codec,), {})
```
aboudreault: @mambocab  This is the part I had to add to the metaclass... otherwise all subclassing was overriding the error_class of ErrorMessageSub.Codec (those that simply reuse the default one). I don't especially like that... do you see a workaround?

mambocab: Ohh interesting, whoops. Would this be better?

```python
# Make unique subclass of the superclass's `.Codec` if necessary
# e.g. if you're defining `class FooMessage(ErrorMessageSub)` without its own
# `.Codec`, this will subclass `ErrorMessageSub.Codec` so that
# `FooMessage.Codec` and `ErrorMessageSub.Codec` are different objects.
if 'Codec' not in clsdict:
    class Codec(cls.Codec):
        pass
    clsdict['Codec'] = Codec

super(ErrorMessageSubTypeWithCodecMeta, cls).__init__(name, bases, clsdict)
# If we hadn't created a Codec subclass above, we could be altering a `Codec` that
# `cls` inherited
cls.Codec.error_class = cls
```

It's not quite as succinct, but I think it's clearer. The `if 'Codec' not in clsdict` also prevents us from creating unnecessary new subclasses. Another option:

```python
super(ErrorMessageSubTypeWithCodecMeta, cls).__init__(name, bases, clsdict)

if 'Codec' not in clsdict:
    # Make unique subclass of the superclass's `.Codec` if necessary
    # e.g. if you're defining `class FooMessage(ErrorMessageSub)` without its own
    # `.Codec`, this will subclass `ErrorMessageSub.Codec` so that
    # `FooMessage.Codec` and `ErrorMessageSub.Codec` are different objects.
    class Codec(cls.Codec):
        error_class = cls
    cls.Codec = Codec
else:
    cls.Codec.error_class = cls
```

I don't know off the top of my head if these have different behavior or produce observably different classes, but I think I slightly prefer the first option -- all `Codec` classes produced this way will be assigned to `cls.Codec` in `super(...).__init__`, and then have `error_class` assigned after creation.

aboudreault:
Comments addressed based on our discussions.

mambocab:
Code looks good! I'm +1 if you are.

PR #965, opened 2018-06-19
PYTHON-960: Create a structure for protocol constants

aboudreault:


cassandra/protocol.py
```diff
@@ -54,13 +53,6 @@ class InternalError(Exception):
 HEADER_DIRECTION_TO_CLIENT = 0x80
 HEADER_DIRECTION_MASK = 0x80
```
aboudreault: umm... this doesn't seem to be used at all..

aboudreault: to be removed before merge

cassandra/__init__.py
```diff
@@ -26,67 +26,230 @@ def emit(self, record):
...
 
 
+class ProtocolConstants(object):
+
+    class Version(object):
+        V3 = 3
+        V4 = 4
+        V5 = 5
+
+    class Opcode(object):
```
mambocab: This might be a good opportunity to bring in some kind of `enum` for ease of use and free pretty-printing. Thoughts? We can even avoid introducing a dependency by vendoring `enum34`, which I think maintains compatibility with the stdlib `enum` and consists of a single file.

cassandra/cluster.py
```diff
@@ -2757,26 +2758,27 @@ def _refresh_nodes_if_not_up(self, addr):
...
         change_type = event["change_type"]
         addr = self._translate_address(event["address"][0])
-        if change_type == "NEW_NODE" or change_type == "MOVED_NODE":
+        if change_type == ProtocolConstants.TopologyChangeType.NEW_NODE or \
+                change_type == ProtocolConstants.TopologyChangeType.MOVED_NODE:
             if self._topology_event_refresh_window >= 0:
                 delay = self._delay_for_event_type('topology_change', self._topology_event_refresh_window)
                 self._cluster.scheduler.schedule_unique(delay, self._refresh_nodes_if_not_up, addr)
-        elif change_type == "REMOVED_NODE":
+        elif change_type == ProtocolConstants.TopologyChangeType.REMOVED_NODE:
```
mambocab: Yessssssssssssss, I love not comparing to random string literals

cassandra/__init__.py
aboudreault: This is a good suggestion. However, there is something I don't like from that Enum: we need to add `.value` everywhere in the code. Let me think a bit more about it..

aboudreault: I guess these kind of lines could:
```
 if kind == ProtocolConstants.ResultKind.VOID:
```
would become
```
 if ProtocolConstants.ResultKind(kind) == ProtocolConstants.ResultKind.VOID:
```

but what about this:
```
BATCH_LOG = ProtocolConstants.WriteType.BATCH_LOG
``
Should it become 
```
BATCH_LOG = ProtocolConstants.WriteType.BATCH_LOG
```
or not?

So my main worry is about thwhere the line should be drawn between the use of those enum versus the raw protocol value. Cause this has an impact as well on the API. e.g 
```
execute(..., consistency_level=??)  # ProtocolConstants.ConsistencyLevel.ONE or ProtocolConstants.ConsistencyLevel.ONE.value ? Supporting only the enum removes the ability to use a raw value directly.
```

Comment (mambocab) [reply cassandra/__init__.py]: Scrolling through, it looks like we could probably use `IntEnum` and `IntFlag` for the int codes and flags, respectively. If those values work, I _think_ that normal integers will work as well.

For strings, we might be able to do this:

```python
In [3]: class WriteType(str, Enum):
   ...:     SIMPLE = 'SIMPLE'
   ...:     BATCH = 'BATCH'
   ...:     
```

In particular, this gives us:
```python
In [4]: WriteType.SIMPLE == 'SIMPLE'
Out[4]: True
```

-- which would be `False` otherwise. I think that, in combination with `IntEnum` and `IntFlag`, makes stuff like `kind == ProtocolConstants.ResultKind.VOID` work for all the types we care about.

Another thing to keep in mind: we know all the entry points, right? We could have something like

```python
class ExecutionProfile(object):
    def __init__(self, load_balancing_policy=_NOT_SET, retry_policy=None,
                 consistency_level=ConsistencyLevel.LOCAL_ONE, ...):
         ...
         self.consistency_level = ConsistencyLevel(consistency_level)
```

so that all internals use our enums.

mambocab: > Sorry Jim if I sound super-negative about this change... that's not the goal

Not a problem! I'm glad we're talking through the idea.

> So, we get a better printing and a value_to_name mapping

Yeah, that's the benefit. Maybe I'm overvaluing it -- I think it'd be really nice for us, and I think it "modernizes" the codebase, but it's possible that it's not a big benefit for users.

> FYI, IntFlag and Flag are only in 3.6+... so we'll need to carry our vendor enum.py for 10 years.

Thanks for pointing that out, I missed that those were new. I'm not toooo concerned about vendoring `enum` -- it's in the stdlib, so we shouldn't have to change it too frequently? Right? :grimacing:

> Just another thing that I'm thinking is that currently, the driver is a no-brain. With that change.. it will crash if we call an unknown consistencyLevel (not a huge deal but still...)

This is a really good point. Does causing that crash break anything? My initial thought is that it doesn't, except when working with an unsupported protocol version.

Another option is to add our own enum subclass that provides a pass-through converter class method:

```python
@classmethod
def maybe(cls, value):  # don't actually call this method `maybe`
    try:
        return cls(value)
    except ValueError:
        return value
```

and use `self.consistency_level = ConsistencyLevel.maybe(consistency_level)`. This lets us use the enum internally wherever we can and fall back to builtin types when we can't.

aboudreault: Not sure what do to here Jim. I don't think I see the same benefits than you with that enum, especially that users don't care at all about the protocol internal values. My opinion is that we are just adding some complexities to manage all those raw values. I would be OK to vendor that enum module if I could see a pain we currently experience, but I can't at the moment. I also like the fact that there is no magic python code that tries to handle what is a good consistency leve and so on... the server already does that and produce good errors.

I think my suggestion would be to keep this as is for now. Maybe we will have more time for that kind of changes in the 5.x release. I mean, for more architectural changes... since 4.x is focused on the pluggability. I also don't have any objection if you want to give enums a try on your own in a follow-up ticket.

PR #966, opened 2018-06-20
Python fix keyspace test hang

mambocab:
Addresses [PYTHON-998](https://datastax-oss.atlassian.net/browse/PYTHON-998). See commit message:

> The old behavior led to Session objects not being registered with the cluster, and thus not cleaned up on Cluster.shutdown().


mambocab:
Following up with a style cleanup commit.

aboudreault:
LGTM!

mambocab:
Merging.

PR #976, opened 2018-08-15
Fix UDT changed behavior with updating fields of UDT

CharAct3:
### Issue: It doesn't work when only updating fields of UDT.

```python
item.udt.age = 1
item.save()

# or 
item.update(udt=item.udt)
```
In these cases, cqlengine won't treat `udt` as `changed`.


mambocab:
Closing for now due to the missing CLA. Please let us know if we should have another look, though, and thank you for the PR.

PR #977, opened 2018-08-15
Fire callbacks on ResponseFutures before unleashing result()

spladug:
When callbacks and blocking on .result() are mixed, it's currently a
race for which happens first. By flipping the order of operations here,
we guarantee that the callbacks are all complete before unleashing the
main thread. This makes it possible to reliably do stuff with callbacks
like collect metrics while the main application uses .result().

For reference, here's [the issue described in another context](https://github.com/reddit/baseplate/issues/100) and [the code in question](https://github.com/reddit/baseplate/blob/master/baseplate/context/cassandra.py#L134-L152).

I'm submitting this as a straw-man. I'm not sure what the downsides of flipping this would be but figured a diff would be the best way to show what I had in mind.

p.s. It appears the mailing list in the README is non-existent and the Slack
requires agreeing to various legal agreements for DataStax Academy to get
access. Sorry if jumping straight to a PR was bad form.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3948,11 +3948,15 @@ class ResponseFuture(object):
                 for (fn, args, kwargs) in self._callbacks
             )
 
-        self._event.set()
-
         # apply each callback
         for callback_partial in to_call:
             callback_partial()
+
+        # this should come after the callbacks are fired so that when mixing
+        # callbacks and .result() we know the callbacks are all done by the
+        # time .result() returns.
+        self._event.set()
+
 
     def _set_final_exception(self, response):
         self._cancel_timer()
@@ -3969,11 +3973,15 @@ class ResponseFuture(object):
                 partial(fn, response, *args, **kwargs)
                 for (fn, args, kwargs) in self._errbacks
             )
-        self._event.set()
 
         # apply each callback
         for callback_partial in to_call:
             callback_partial()
+
+        # this should come after the callbacks are fired so that when mixing
+        # callbacks and .result() we know the callbacks are all done by the
+        # time .result() returns.
+        self._event.set()
 
     def _retry(self, reuse_connection, consistency_level, host):
         if self._final_exception:


spladug:
:grimacing: The CLA app appears to have a broken SSL certificate.

mambocab:
Thanks for the report and PR, @spladug. Sorry for the technical issues :grimacing: We're working on getting good SSL certs for the CLA, and also talking with Google about why our Google Group disappeared. A PR is a perfectly good venue for this report; I understand if you don't want to agree to the DS Academy terms. I may end up linking to this report in our JIRA.

On to technical concerns: *tl;dr*: Unfortunately, `ResponseFuture` fundamentally can't do the chainable-future-like behavior you want. Because the list of callbacks is defined after `ResponseFuture.__init__`, and because `add_callback` can either add the callback to the list or execute it on the spot (ignoring the list), our `ResponseFuture` just can't provide any notion of "all callbacks are done".

More detailed thoughts:

- Blocking `_set_final_result()` is the wrong way to do this -- we want callback processing to be async so the main thread can use the result while callbacks do their thing in the event loop.
- In trying to come up with a workaround, I thought "oh, `@spladug` can just create a new callback with `add_callback`, and when it completes, that will signal that all callbacks have executed". This is incorrect, because, if the result is set, `add_callback` can execute its argument concurrently with the callbacks in `self._callbacks` and can finish before they've returned.
- We could add add a `_callbacks_done` event that we set after the callback iteration in `_set_final_exception` and provide a way for users to wait on it. You could imagine a new kwarg in `result()`, e.g. `def result(self, wait_on_callbacks=False):`. Unfortunately, this will *not* block on the completion of callbacks that are added after the final result is set, so it doesn't give a meaningful "all callbacks are done" signal.

So, this leads to the conclusion above: because new callbacks can be added to our `ResponseFuture` at any time, and because they can be executed in multiple places, we can't provide a way to block on all of them in the current model of callback-tracking. This definitely isn't ideal, but I don't think we have evidence to demonstrate this is important enough to change (or add complexity to) how our callback model works.

We're open to ideas, so please let us know if you have thoughts on this. For now, though, my answer is that any dependencies you have between callbacks need to be managed in your application.

spladug:
Thanks for the detailed feedback!

I suspected blocking the main thread would be a no no here :)

It does sound like it'd be rather difficult to get to full "all callbacks are done" given the flexibility of the system, but in my case I need something less rigorous: "let me know when you're about to tell the application this action is complete".  Since it's at a more invisible frameworky level where we're injecting observability stuff, it's hard for the application itself to manage the dependency ordering there. One option would be an explicit event system like SQLAlchemy's `before_cursor_execute` / `after_cursor_execute`. Alternatively, if there's a way for my code to wrap and fake out `.result()` so that it blocks on my one monitoring callback, that'd be fine too. (I played around with that a bit but I wasn't able to find a simple solution for it).

mambocab:
> Alternatively, if there's a way for my code to wrap and fake out .result() so that it blocks on my one monitoring callback, that'd be fine too. (I played around with that a bit but I wasn't able to find a simple solution for it).

This is the direction I had imagined. Any idea what the issue was? The simplest solution I can think of is something like:

```python
class CallbackManager(object):  # untested code; no guarantees :grimacing:
    def __init__(self, fn):
        self.fn, self. event = fn, threading.Event()
    def __call__(self, *args, **kwargs):
        try:
            result = self.fn(*args, **kwargs)
        except:
            pass  # error handling
        else:
            self.event.set()
        return result
    def done(self):
        return self.event.is_set()
```

which gives you a handle for checking if the specific task you care about is done.

mambocab:
Just realized I had an unstated assumption in the above, so I thought I'd explain my reasoning for proposing a wrapper around a callback:

Because of the race between result-setting and any `add_callback` calls, _a callback cannot know if another callback is done_.  You can only answer the question "is this particular callback done?".

spladug:
Apologies for the super long delay here. We've taken your suggestion and applied it to our use case so I'll close this out. Thank you for the advice!

PR #981, opened 2018-09-03
batch query async support

omrikiei:
I needed to use an asynchronous batch query for a project, originally creating a subclass of batch query. however I think it would be great to have this kind of support in the class itself.
I would like to add additional test cases in the future.


aboudreault:
Hello @omrikiei! CQLEngine async execution is a feature we have done for the next major version (4.x). All cqlengine apis will support async. 

* https://datastax-oss.atlassian.net/browse/PYTHON-605
* e.g: https://github.com/datastax/python-driver/blob/4.x/cassandra/cqlengine/query.py#L216

PR #983, opened 2018-09-06
Build and upload wheels for master CI builds

dgladkov:
This is an untested POC of how we can make creating release wheels easier. This also requires additional env configuration for Travis: https://docs.travis-ci.com/user/environment-variables/#defining-public-variables-in-travisyml

This feature can be also accompanied by a release automation script that does sdist, fetches appropriate wheels for a given release tag from S3 and then uploads sdist + wheels to PyPI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ python:
   - "pypy-5.4.1"
 
 sudo: false
-env:
-  - CASS_DRIVER_NO_CYTHON=1
 
 addons:
+  artifacts:
+    working_dir: dist
   apt:
     packages:
     - build-essential
@@ -32,6 +32,10 @@ script:
       echo "This PR includes code changes, but no CHANGELOG changes. Please include a CHANGELOG entry."
       exit 1
     fi
-  - tox
-  - tox -e gevent_loop
-  - tox -e eventlet_loop
+  - CASS_DRIVER_NO_CYTHON=1 tox
+  - CASS_DRIVER_NO_CYTHON=1 tox -e gevent_loop
+  - CASS_DRIVER_NO_CYTHON=1 tox -e eventlet_loop
+  - |
+    if [ "${TRAVIS_BRANCH}" = "master" ]; then
+      python setup.py bdist_wheel
+    fi


mambocab:
Thanks for the POC, @dgladkov.

> This feature can be also accompanied by a release automation script that does sdist, fetches appropriate wheels for a given release tag from S3 and then uploads sdist + wheels to PyPI

This infrastructure (plus that required to test the built wheels) is the stuff that we've chosen to not maintain for OSS -- I think the costs for us outweigh the benefits for users, especially because users can build and distribute their own wheels.

jezdez:
@mambocab I'm not sure I understand what prevents you from setting this up once (for free) and automate the release to PyPI via Travis' PyPI deploy system. It [allows building wheel files](https://docs.travis-ci.com/user/deployment/pypi/#uploading-different-distributions) if you configure it like that. What does "plus that required to test the built wheels" mean? Can you elaborate what the test case is?

With regard to your current strategy, I know that your users *can* build wheel files on their own using your docs but I'd argue that very few do, resulting in a wasteful build pattern which isn't really an awesome reason to pay you for the non-OSS product. I think not providing wheel files is not bringing you new customers, but is a missed chance to show that you follow best practices to the Python community. Given its continues growth, I'm not sure if that is a area worth missing.

To reiterate, the purpose of wheel files is to provide [a format that it easier to install than sdists](https://www.python.org/dev/peps/pep-0427/#rationale). Plus they also completely remove an attack vector of running arbitrary code in setup.py files since whl files are just unzipped during installation. Those two aspects seem great things to also provide to users that aren't paying customers.

If you don't have engineering resources free to do the one-time setup I'd like to suggest to ask for help from the OSS community (*waves*) to help. If you're comfortable with doing that and decide to reconsider your position, could you please elaborate what your constraints are, other than what you've shared so far?

mambocab:
Hi Jannis, thanks for your input. You're right that we're missing out on an opportunity to use Python best practices for the OSS releases.

However, we've discussed internally (among developers and management), and we see the cost/benefit analysis differently. Releasing wheels for the OSS driver exposes us to additional support and maintenance burden among a group of users who, broadly speaking, don't already have open support channels with us.

For the time being, we're going to continue distributing wheels for the DSE driver only.

jezdez:
Given that Datastax is profiting quite a bit from the Open Source community (read: other companies working on the Apache Cassanda project) I'd be interested to hear the specifics for the cost/benefit analysis that you've had internally.

@jbellis Hey there, would you mind elaborating?

tmehlinger:
For anyone stumbling upon this issue using versions earlier than 3.21.0, they now upload wheels.

https://pypi.org/project/cassandra-driver/3.21.0/#files

jezdez:
[Woohoo](https://github.com/datastax/python-driver/commit/e696a82b0c2d9cf5b99912e62b1fa0ffd458a349)! Thank you @aboudreault!

PR #984, opened 2018-09-14
Python 1008

mambocab:
Adds a failing test, and then a fix, for [PYTHON-1008](https://datastax-oss.atlassian.net/browse/PYTHON-1008). I explained the bug in a commit message, reproduced below:

>The CREATE CUSTOM INDEX statement initially generated in
`IndexMetadata.as_cql_query` is always a `unicode` object under Python 2,
because they are derived from CQL `text` fields in in
`system_schema.indexes`, and we serialize the bytes we get from
`text`-type values with `.encode('utf-8')`.

>The result of `cql_encode_all_types` is always a `six.binary_type`, i.e.
a `str` under Python 2.

>Notably, Python 2 `str` objects cannot be cast to `unicode` objects if
they contain non-ASCII characters, including those used to encode
Unicode characters in UTF-8. These casting errors can occur during
attempts to concatenate these objects:

```
>>> u"I'm a `unicode` object" + ", and I am a `str`."
u"I'm a `unicode` object, and I am a `str`."
>>> u"I'm a `unicode` object" + ", and I am a `str` with unicode code points \xf0\x9f\x98\xac"
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf0 in position 45: ordinal not in range(128)
```

>So, the error reported in CASSANDRA-14632 happens when we try to
concatenate a `str` containing escaped UTF-8 bytes, as generated by
`cql_encode_all_types`, onto `ret`, a `unicode`.

>Since we know `ret` is always a `six.text_type`, we can check if we need
to convert the result of `cql_encode_all_types`, and do so. We should
only ever need to under Python 2.

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -1435,10 +1435,8 @@ class IndexMetadata(object):
                 index_target,
                 class_name)
             if options:
-                opts_cql_encoded = _encoder.cql_encode_all_types(options)
-                # PYTHON-1008
-                if isinstance(opts_cql_encoded, six.binary_type):
-                    opts_cql_encoded = opts_cql_encoded.decode('utf-8')
+                # PYTHON-1008: `ret` will always be a unicode
+                opts_cql_encoded = _encoder.cql_encode_all_types(options, as_text_type=True)
                 ret += " WITH OPTIONS = %s" % opts_cql_encoded
             return ret
 
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -1435,7 +1435,7 @@ class IndexMetadata(object):
                 index_target,
                 class_name)
             if options:
-                opts_cql_encoded = Encoder().cql_encode_all_types(options)
+                opts_cql_encoded = _encoder.cql_encode_all_types(options)
                 # PYTHON-1008
                 if isinstance(opts_cql_encoded, six.binary_type):
                     opts_cql_encoded = opts_cql_encoded.decode('utf-8')
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -1435,7 +1435,11 @@ class IndexMetadata(object):
                 index_target,
                 class_name)
             if options:
-                ret += " WITH OPTIONS = %s" % Encoder().cql_encode_all_types(options)
+                opts_cql_encoded = Encoder().cql_encode_all_types(options)
+                # PYTHON-1008
+                if isinstance(opts_cql_encoded, six.binary_type):
+                    opts_cql_encoded = opts_cql_encoded.decode('utf-8')
+                ret += " WITH OPTIONS = %s" % opts_cql_encoded
             return ret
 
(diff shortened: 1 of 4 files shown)


cassandra/metadata.py
```diff
@@ -1435,7 +1435,11 @@ def as_cql_query(self):
                 index_target,
                 class_name)
             if options:
-                ret += " WITH OPTIONS = %s" % Encoder().cql_encode_all_types(options)
+                opts_cql_encoded = _encoder.cql_encode_all_types(options)
+                # PYTHON-1008
+                if isinstance(opts_cql_encoded, six.binary_type):
+                    opts_cql_encoded = opts_cql_encoded.decode('utf-8')
+                ret += " WITH OPTIONS = %s" % opts_cql_encoded
```
aboudreault: I think I have the same question than before, shouldn't encoders deal with that, rather than adding a hack-decoding in metadata.py? If there is no reason `options` (or another cql parts) to be encoded directly to `binary_type`.. we should probably create a new encoder function that is aware of this and always returns unicode.

mambocab: Ah, I understand your objection now. If the `cql_encode_*` methods were only used for generating CQL for display, I'd be firmly +1. However, these methods are also used internally for binding arguments to non-prepared statements, and I think this use case wants `six.binary_type` instance -- I'll have a look at that now to confirm.

Are you proposing an interface that always returns a `six.string_type` or one that allows callers to choose?

mambocab: See [`bind_params`](https://github.com/datastax/python-driver/blob/850f218b5cde1fd000338517372395f690097153/cassandra/query.py#L891), [used here, for example](https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L2239)

mambocab:
Continuing the discussion from the above thread:

What do you think of an API like `cql_encode_*(obj, as_unicode=False)`? To me, that's more or less equivalent to the conversion in this PR -- it's still an explicit request for the type. Of course, it's more succinct, so maybe that's a good thing.

Looking around, I don't see anywhere else where this conversion comes up, so I think the interface would be used exactly there.

mambocab:
Makes sense to me. I'll add that kwarg to `all_types`. I like your idea of restricting it to that method; the other methods aren't referenced outside `encoder`, so there probably isn't a need to add that check elsewhere.

One question:

>the extra encode/decode it will cause

What do you mean here? I don't think this causes any extra encode/decode compared to the existing code.

aboudreault:
No you are right, the extra decoding is already done in this PR. I was comparing to sending `as_unicode=True` to the cql_encode_str* functions so we could avoid it. But in this case.. it's not enough important.

mambocab:
Got it. Cool, thanks!

mambocab:
rebased e4af0701 onto master (2ace9322) to deal with CHANGELOG conflicts

PR #986, opened 2018-09-21
Python 1020 fix vtables against dse 6.0.x

mambocab:
Pretty simple fix for (PYTHON-1020)[https://datastax-oss.atlassian.net/browse/PYTHON-1020]. Modeled after [a similar case where we ignore failures to find a table](https://github.com/datastax/python-driver/blob/12b59b80cee69a1c41e57558787387b50eae1ef2/cassandra/metadata.py#L1785-L1789).

The second commit is a refactoring that simplifies the failure-ignoring logic. I definitely want to use it in the V4 parser, but I don't feel too strongly about it elsewhere; this code changes so infrequently that abstracting everything isn't strictly necessary.

I have a couple questions that came up when I looked at [cases like this](https://github.com/datastax/python-driver/blob/2ace9322870b38629e3235144f99470b1351c711/cassandra/metadata.py#L2115):

- Should the new `_handle_results` logic also allow the caller to pass a string to be logged if there's an exception? I don't feel strongly either way. I don't love complicating `_handle_results` further but more logging is good.
- If we choose to add debug messages to the new `_handle_results`, should we port those cases to the new logic? I think that it's not worth the risk of touching code that has worked for a long time, but I want to hear your thoughts.


cassandra/metadata.py
```diff
@@ -2529,12 +2542,21 @@ def _query_all(self):
...
+        )
+        self.virtual_tables_result = self._handle_results(
+            virtual_table_success, virtual_table_result,
+            expected_failures=InvalidRequest
+        )
+        self.virtual_columns_result = self._handle_results(
+            virtual_column_success, virtual_column_result,
+            expected_failures=InvalidRequest
+        )
+
```
aboudreault: My first idea would have been to simply do a `if virtual_ks_success`... but I think your way makes it more robust. :+1:

mambocab: That was my first thought too, but other places in the codebase check for an `InvalidRequest`, which seems like a good idea, since it maps to "this ks doesn't exist", which is the acceptable case.

Might be worth documenting that in `_handle_results`'s docstring, actually.

aboudreault: yep, I like your idea!

mambocab:
Rebased away CHANGELOG conflicts, tests should be running clean now. Could I get one last review?

PR #987, opened 2018-09-24
Improve error messages on persistent schema disagreements 

mambocab:
`wait_for_schema_agreement` doesn't raise an exception if the schema doesn't agree, so this change adds a check for its result. If we see failures as a result of persistent schema disagreements, they'll have the new error message instead of more confusing C* errors.

I made this change when I saw the failure in [this test run](http://jenkins-devtools.datastax.lan/job/datastax.python-driver.master.nightly_master/30/CASSANDRA_VERSION=3.11,CYTHON=CYTHON,OS_VERSION=ubuntu%2Fbionic64%2Fpython-driver,PYTHON_VERSION=3.5/testReport/junit/tests.integration.long.test_schema/SchemaTests/test_for_schema_disagreement_attribute/). If the test failed because the schema update failed entirely, then this change won't help improve the error. However, I think this change does improve the message in other failure cases.

### DIFF
--- a/tests/integration/long/test_schema.py
+++ b/tests/integration/long/test_schema.py
@@ -142,7 +142,7 @@ class SchemaTests(unittest.TestCase):
         rs = session.execute("DROP KEYSPACE test_schema_disagreement")
         self.check_and_wait_for_agreement(session, rs, False)
         cluster.shutdown()
-        
+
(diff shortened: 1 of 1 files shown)


mambocab:
Thanks, merging!

PR #989, opened 2018-10-05
PYTHON-1024: fix catch ssl err catching

mambocab:
Addresses [PYTHON-1024](https://datastax-oss.atlassian.net/browse/PYTHON-1024).

The new test for SSL error handling in the first commit is intended to fail on that commit -- Alan, could you check that it does?

I'm pretty confident that [removing this check from this test](https://github.com/datastax/python-driver/compare/long-python_catch-ssl-errs?expand=1#diff-b0e8d7ab6ad634ee6bddc3f68e7c013eL269) is fine. Could you poke through (in particular where `_current_frame` is used) and see if I'm missing anything?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 3.15.2
 ======
 
+Bug Fixes
+---------
+* Improve and fix socket error-catching code in nonblocking-socket reactors (PYTHON-1024)
+
 Other
 -----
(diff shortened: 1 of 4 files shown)


mambocab:
Thanks for the review! There was a flake on one Jenkins run (`test_cql_optional_params`), but it looks like one that simply can't be avoided all the time -- there was a failure to reach schema agreement.

PR #991, opened 2018-10-24
PYTHON-1015: Fix wrong resultset indexing use

aboudreault:
There are many resultset indexing use in tests. This is normal since all those fixes are in the 4.x branch. I did some attempts to ignore those warning when running tests without luck.


mambocab:
Re-running the tests

mambocab:
LGTM, +1!

PR #992, opened 2018-10-24
PYTHON-1026: ResultSet.one() fails if the row_factory is using a generator

aboudreault:


cassandra/cluster.py
```diff
@@ -4244,7 +4244,14 @@ def one(self):
         you know a query returns a single row. Consider using an iterator if the
         ResultSet contains more than one row.
         """
-        return self._current_rows[0] if self._current_rows else None
+        row = None
+        if self._current_rows:
+            try:
+                row = self._current_rows[0]
+            except Exception:  # generator object is not subscriptable, PYTHON-1026
```
mambocab:
```suggestion
            except TypeError:  # generator object is not subscriptable, PYTHON-1026
```

No reason not to specify IMO

aboudreault: totally... no idea why I put Exception there.

PR #996, opened 2018-11-05
Python 895 require only one resolved contact point

mambocab:
We can push this to the next release if timing necessitates it; I'm just PRing because I found this branch lying around.


aboudreault:
If our test suite passes, no objection to include this in 3.16.

mambocab:
Awesome, thanks. Merging then!

PR #1000, opened 2018-11-15
explicitly say about keyfile and certfile parameters for SSL

alexott:


mambocab:
Thanks for the PR! The SSL docs are generally a bit of a pain point for us, unfortunately, so we appreciate the contribution.

Here's the issue -- is sort of covered by the `These will be passed as kwargs to ssl.wrap_socket() when new sockets are created` docs above, though not explicitly. I'm going to talk with the team about this -- we don't want to completely copy the docs for `ssl` and `ssl.wrap_socket()` here, but we may be able to find some kind of compromise.

absurdfarce:

PR #1001, opened 2018-12-03
Fix python version in Contributing guide

BackEndTea:


mambocab:
Great catch, thanks for the fix! Merging.

BackEndTea:
Thank you @mambocab

PR #1002, opened 2018-12-14
Python 914: use datetime.fromtimestamp

mambocab:
See the commit message for the first commit for comments on the solution here. I've also made a multiplication change that makes an `int` call unnecessary.


aboudreault:
Is there any unit test we could add related to this change?

absurdfarce:

PR #1003, opened 2019-01-21
PYTHON-891: Improve handling of ssl errors when sending a message

aboudreault:
This is similar to the PYTHON-1024 fix. This PR improves how SSL WANT_READ/WANT_WRITE errors are handled. Basically, we just re-queue the chunk to complete the message instead of erroring the connection.

Note: The asyncio fix is not included in this PR.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,10 @@ Features
 * Add Cluster ssl_context option to enable SSL (PYTHON-995)
 * Allow encrypted private keys for 2-way SSL cluster connections (PYTHON-995)
 
+Bug Fixes
+---------
+* NoHostAvailable when all hosts are up and connectable (PYTHON-891)
+
 Other
 -----
 * Fail faster on incorrect lz4 import (PYTHON-1042)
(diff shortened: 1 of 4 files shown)


mambocab:
Code looks good, and the test fails when I remove the new code. +1.

In the logs for the old, failing behavior, I'm seeing ```2019-01-29 17:10:23,313 DEBUG [connection:401]: Defuncting connection (139697416395240) to 127.0.0.1:                                                                                                                                                         
  File "/home/mambocab/cstar_src/python-driver/cassandra/io/asyncorereactor.py", line 402, in handle_write                                                                                                                                                    
    sent = self.send(next_msg)                                                                                                                                                                                                                                
  File "/home/mambocab/.pyenv/versions/3.7.1/lib/python3.7/asyncore.py", line 360, in send                                                                                                                                                                    
    result = self.socket.send(data)                                                                                                                                                                                                                           
  File "/home/mambocab/.pyenv/versions/3.7.1/lib/python3.7/ssl.py", line 984, in send                                                                                                                                                                         
    return self._sslobj.write(data)                                                                                                                                                                                                                           
ssl.SSLWantWriteError: The operation did not complete (write) (_ssl.c:2324)                                                                                                                                                                                   
```

I think this would've been an easier problem to debug if we'd logged this above DEBUG -- maybe the error should be logged at WARN? We already log ```
2019-01-29 17:10:23,315 WARNING [cluster:1551]: Host 127.0.0.1 has been marked down
```
so it doesn't require adding a whole new WARN.

PR #1004, opened 2019-01-23
PYTHON-995: Allow encrypted private keys for 2-way SSL cluster connections

aboudreault:
Add Cluster.ssl_context option to enable SSL


CHANGELOG.rst
```diff
@@ -2,10 +2,20 @@
...
+
 Other
 -----
 * Fail faster on incorrect lz4 import (PYTHON-1042)
 
+Deprecations
+------------
+
+* Using Cluster.ssl_options to enable SSL is deprecated and will be removed in
+  the next major release, use ssl_context.
```
aboudreault: RFC: I don't know what you think of a this section in our CHANGELOG... but I've seen it in another project and I really like that. It makes it clear for users if they have important deprecation tasks to fix in their code. Note that there is no migration guide for minor releases.

CHANGELOG.rst
```diff
@@ -2,10 +2,20 @@
 ======
 NOT RELEASED
 
+Features
+--------
+* Add Cluster ssl_context option to enable SSL (PYTHON-995)
+* Allow encrypted private keys for 2-way SSL cluster connections (PYTHON-995)
+
```
aboudreault: RFC: I wrote 2 entries just for the discussion purpose. The first one is the one that makes it 100% clear to users what new feature/improvement we did in PYTHON-995. The second is simply the ticket title as we used to add in the changelog. I would like to know if you have any idea/preference how we should handle this in the future. I know we are not going to change all the jira ticket titles... but at the same time, I think it's very valuable to users to know on a quick changelog read what new features we added.

build.yaml
```diff
@@ -37,7 +37,6 @@ schedules:
       include: [/python.*/]
     env_vars: |
       EVENT_LOOP_MANAGER='libev'
-      EXCLUDE_LONG=1
```
aboudreault: will be remove before the merge, only for this PR purpose

build.yaml
```diff
@@ -118,7 +117,6 @@ cassandra:
   - '3.0'
   - '3.11'
   - 'test-dse'
-  - 'dse-6.7'
```
aboudreault: no idea why we had this before in OSS

CHANGELOG.rst
aboudreault: I think we should do what the java team do: creating an additional ticket for the purpose when appropriate: 
screenshot_20190123_153830

CHANGELOG.rst
aboudreault: Well, nobody checks the migration guide until they migrate to a major version. Note that this is not a migration guide replacement. This is to document changes so they can fix/be aware along the way of minor versions. That also means, no more surprises DeprecatedWarning showing in the logs. It's also useful to see everything to recheck when writing/finalizing the migration guide. Let me know what you think.

CHANGELOG.rst
aboudreault: Good. I'm going to follow this in the future then.

CHANGELOG.rst
mambocab: Ah, I see the difference. That sounds great, +1.

PR #1007, opened 2019-01-28
PYTHON-1068: send driver name/version in startup

mambocab:
The tests passed except for the `test-dse` tests. Test run 1 is the same code; the new force-pushed commit was just a commit message change.

You can verify that the feature works manually by connecting to a C* instance running `trunk` and checking `s.execute('SELECT * FROM system_views.clients'))`.

### DIFF
--- a/cassandra/connection.py
+++ b/cassandra/connection.py
@@ -106,6 +106,8 @@ else:
     locally_supported_compressions['snappy'] = (snappy.compress, decompress)
 
 
+DRIVER_NAME, DRIVER_VERSION = 'DataStax Python Driver', sys.modules['cassandra'].__version__
+
 PROTOCOL_VERSION_MASK = 0x7f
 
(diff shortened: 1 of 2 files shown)


cassandra/connection.py
```diff
@@ -721,7 +723,8 @@ def _handle_options_response(self, options_response):
     @defunct_on_error
     def _send_startup_message(self, compression=None, no_compact=False):
         log.debug("Sending StartupMessage on %s", self)
-        opts = {}
+        opts = {'DRIVER_NAME': DRIVER_NAME,
+                'DRIVER_VERSION': DRIVER_VERSION}
```
aboudreault: so that will just pass with older versions as well? Probably simply ignored?

mambocab: Yep, seems to just be ignored.

PR #1008, opened 2019-02-06
PYTHON-932: Add Session.get_execution_profile

aboudreault:


mambocab:
Looks good, and tests look good outside of unrelated test issues, +1!

PR #1016, opened 2019-05-06
PYTHON-1091: Improve version parsing to support a non-integer 4th component

aboudreault:
This will be removed when we remove the python2 support..


aboudreault:
Merging, feedback/comments are still welcome.

jorgebay:
Nice test coverage 👍

PR #1017, opened 2019-05-10
PYTHON-1064: Expose on_request_error method in the RetryPolicy

aboudreault:


cassandra/cluster.py
```diff
@@ -3925,20 +3926,15 @@ def _set_result(self, host, connection, pool, response):
                         self._metrics.on_unavailable()
                     retry = retry_policy.on_unavailable(
                         self.query, retry_num=self._query_retries, **response.info)
-                elif isinstance(response, OverloadedErrorMessage):
+                elif isinstance(response, (OverloadedErrorMessage,
+                                           IsBootstrappingErrorMessage,
+                                           TruncateError, ServerError)):
+                    log.warning("Host %s error: %s. retrying against a different host", host, response.summary)
```
jorgebay: NIT: Log message states that it's retrying and it is now up to the retry policy to decide.

cassandra/policies.py
```diff
@@ -839,6 +839,31 @@ def on_unavailable(self, query, consistency, required_replicas, alive_replicas,
         """
         return (self.RETRY_NEXT_HOST, None) if retry_num == 0 else (self.RETHROW, None)
 
+    def on_request_error(self, query, consistency, error, retry_num):
+        """
+        This is called when an unexpected error happens. This can be in the
+        following situations:
+
+        * On a connection error
+        * On server errors: overloaded, isBootstrapping, serverError, etc.
```
jorgebay: How `OperationTimedOut` are handled? will the `RetryPolicy` involved in the decision?

tests/integration/simulacron/test_policies.py
```diff
@@ -378,8 +390,62 @@ def test_setting_retry_policy_to_statement(self):
         self.session.execute(bound_stmt)
         self.assertEqual(next(counter_policy.write_timeout), 1)
 
-    def set_cluster(self, retry_policy):
-        self.cluster = Cluster(protocol_version=PROTOCOL_VERSION, compression=False,
-                          default_retry_policy=retry_policy)
-        self.session = self.cluster.connect(wait_for_all_pools=True)
-        self.addCleanup(self.cluster.shutdown)
\ No newline at end of file
+    def test_retry_policy_on_request_error(self):
```
jorgebay: Nice test using simulacron.

cassandra/policies.py
aboudreault: No. In our driver, we only have a global OperationTimedOut, so it is raised when that limit is reach (default 10s)

jorgebay: ah, makes sense.

aboudreault:
@jorgebay  Thanks for the review! Merged.

PR #1019, opened 2019-05-16
PYTHON-1082: accept legacy empty as column names

aboudreault:
Not much to do in the python-driver about this, only a minor fix in cqlengine. All tests are based on the CSHARP ones.


jorgebay:
Good coverage 👍

PR #1020, opened 2019-05-16
PYTHON-1087: make util.SortedSet handle uncomparable elements

aholmberg:


aboudreault:
Nice job! Looks like this solution will work Adam. Tests are currently running: https://jenkins-drivers.build.dsinternal.org/view/python-driver/job/datastax.python-driver.python-1087.commit_branches/

aboudreault:
Screenshot_20190517_112630

PR #1021, opened 2019-05-19
PYTHON-1065: Add jitter to ExponentialReconnectionPolicy

aboudreault:
@jorgebay I noticed when I was done implementing this feature that the nodejs impl does a bit more than the python and the Java ones. The nodeJS one applies only +0-15% when the delay is base_delay and only -0-15% when it's max delay. Do you think it's mandatory to modify?


jorgebay:
I wanted it to be strict on the Node.js implementation and it's documented that way:

> The jitter will never cause the delay to be less than the base delay, or more than the max delay.

I doubt that anyone will notice/care but in any case we should document what's the expected behaviour.

aboudreault:
Yeah, we have the same behavior about that, but we just floor/ceil invalid values to the base/max ones.

PR #1023, opened 2019-06-26
cqlengine: Allow QuerySets to be backed by CQL prepared statements

Asgavar:
So far it's rather a proof-of-concept than actual, ready to be merged, code (tests and DML queries support are not there yet), but before going any further I'd like to know your opinions on the subject.

Personally I feel like prepared statements are a quite important feature, especially for Cassandra, which boasts to be (and is typically used as) a 'performance-first' database and support for them directly in the ORM (or whatever cqlengine actually is, 'R' may be inappropriate) would be much welcome.

### DIFF
--- a/cassandra/cqlengine/query.py
+++ b/cassandra/cqlengine/query.py
@@ -15,6 +15,7 @@
 import copy
 from datetime import datetime, timedelta
 from functools import partial
+import re
 import time
 import six
 from warnings import warn
@@ -391,6 +392,7 @@ class AbstractQuerySet(object):
         self._if_exists = False
         self._fetch_size = None
         self._connection = None
+        self._as_prepared = False
 
     @property
     def column_family_name(self):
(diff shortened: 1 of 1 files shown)


aboudreault:
Hello @Asgavar and thanks for your PR! 

Prepared statements in CQLEngine are something we have investigate. Unfortunately.. it was not as worth as initially expected (especially in term of performance). I have no objection we revisit this need. 

Here is the ticket: https://datastax-oss.atlassian.net/browse/PYTHON-339

Asgavar:
Huh. I admit that the tests we conducted (quite promising) were ran via `cassandra-driver` directly, without using `cqlengine`.  
I'll try to measure the time difference in prepared vs unprepared statements in the wild and get back to you with the results ;]

aboudreault:
Great. Note that your PoC seems good. I think it's a good thing to let the user decide if he wants to use a prepared statement in cqlengine or not. We'll discuss the next step when you are done with your test.

Asgavar:
Okay, so, we ran a randomized test comparing:
1) 'standard' cqlengine usage
2) the experimental prepared statements from my patch posted here
3) SimpleStatements from cassandra-driver
4) PreparedStatements, as above

And, well, as you said, small performance gains are *sometimes* there, but due to deepcoyping QuerySet objects they are sometimes even slower than regular queries, without `as_prepared()` xD.
Then I think they are negligible enough to close this PR, at least with the way `cqlengine` currently works.

aboudreault:
Another point about PS is that they can also have an impact on the server (CPU utilization etc.). From your use case, do you have multiple clients that use this query (core driver and cqlengine)? Do you think it could still be good to propose the option in cqlengine?

PR #1027, opened 2019-07-04
PYTHON-1112: Add 2 concurrent execution examples

aboudreault:


examples/concurrent_executions/execute_async_with_queue.py
```diff
@@ -0,0 +1,65 @@
...
+
+start = time.time()
+futures = queue.Queue(maxsize=CONCURRENCY_LEVEL)
+
+# Chunking way, when the max concurrency level is reached, we
+# wait the current chunk of requests to finish
+for i in range(TOTAL_QUERIES):
+    try:
+        future = session.execute_async(prepared_insert, (uuid.uuid4(), str(i)))
+        futures.put_nowait(future)
```
TheRealFalcon: Nit: only the `futures.put_nowait(future)` line needs to be in the try block, right?

aboudreault: yes! will fix thanks!

PR #1028, opened 2019-07-11
[PYTHON-1100] Fix incorrect metadata for compact counter tables

TheRealFalcon:
https://datastax-oss.atlassian.net/browse/PYTHON-1100

I might have gone overboard with the tests, but I didn't find other tests around the compact storage flags, so figured it'd be good to add a couple other checks there too.

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -2326,9 +2326,9 @@ class SchemaParserV3(SchemaParserV22):
             table_meta.options = self._build_table_options(row)
             flags = row.get('flags', set())
             if flags:
-                compact_static = False
-                table_meta.is_compact_storage = 'dense' in flags or 'super' in flags or 'compound' not in flags
                 is_dense = 'dense' in flags
+                compact_static = not is_dense and 'super' not in flags and 'compound' not in flags
+                table_meta.is_compact_storage = is_dense or 'super' in flags or 'compound' not in flags
             elif virtual:
                 compact_static = False
                 table_meta.is_compact_storage = False
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Features
 
 Bug Fixes
 ---------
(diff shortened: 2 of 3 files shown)


cassandra/metadata.py
```diff
@@ -2326,7 +2326,7 @@ def _build_table_metadata(self, row, col_rows=None, trigger_rows=None, index_row
             table_meta.options = self._build_table_options(row)
             flags = row.get('flags', set())
             if flags:
-                compact_static = False
+                compact_static = 'dense' not in flags and 'super' not in flags and 'compound' not in flags
```
aboudreault: I don't know much about compact storage... have you validated with the C* spec that this is correct?

TheRealFalcon: If we follow the logic here,
old logic:
if we have flags, `compact_static` is set to `False`, otherwise `compact_static` is set to `True`. This is because `dense` and `super` directly affect the compact storage, and `compound` refers to CQL.

However, there's also the `counter` flag. When `counter` is set, `compact_static` can still be `True`, but we were incorrectly setting it `False` here. Instead of setting `compact_static` to `False` if there are any flags, we should only be setting it if we see one of the 3 flags that make it `False`.

aboudreault: Ok. 

nit: move `is_dense = 'dense' in flags` some line before and use it in the compact_static and table_meta.is_compact_storage assignments

TheRealFalcon:
@aboudreault Is this failure related?
https://jenkins-drivers.build.dsinternal.org/job/datastax.python-driver.PR.1028.commit_branches/31/CASSANDRA_VERSION=3.11,CYTHON=NO_CYTHON,OS_VERSION=ubuntu%2Fbionic64%2Fpython-driver,PYTHON_VERSION=2.7/testReport/junit/tests.integration.standard.test_metrics/MetricsTests/test_unavailable/

aboudreault:
LGTM. The failure is just an OperationTimedOut, not related.

PR #1029, opened 2019-07-11
[PYTHON-1117] Call ConnectionException with correct kwargs

TheRealFalcon:
https://datastax-oss.atlassian.net/browse/PYTHON-1117

Would an additional test add any value here?

### DIFF
--- a/tests/integration/standard/test_authentication_misconfiguration.py
+++ b/tests/integration/standard/test_authentication_misconfiguration.py
@@ -1,0 +1,47 @@
+# Copyright DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+from cassandra.cluster import Cluster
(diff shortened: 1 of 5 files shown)


aboudreault:
Maybe we could add small auth fail test to ensure that this code path is tested. Not sure why we don't have already one.. it would have caught this issue.

aboudreault:
Is your test different than the ones we find in test_authentication.py? What I had in mind was more just a new test in test_authentication that try to connect to one node with bad credentials... and we assert that an AuthenticationFailed exception is raised.

TheRealFalcon:
Yeah, sorry, was waiting for the tests to pass before having you take a look again. This particular code path only gets touched if there's one node that has different auth from the rest (e.g., node1 and node2 allow all, but node3 has password auth), and I didn't see any pre-existing tests that had that.

I put it in a separate file because the other one has a module setup, and it seemed easier to just create a new file than push down the CCM setup logic into each class.

tests/integration/standard/test_authentication_misconfiguration.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class MisconfiguredAuthenticationTests(unittest.TestCase):
+    """ One node (not the contact point) has password auth. The rest of the nodes have no auth """
+    @classmethod
+    def setUpClass(cls):
+        if not USE_CASS_EXTERNAL:
+            ccm_cluster = use_cluster(cls.__name__, [3], start=False)
+            node3 = ccm_cluster.nodes['node3']
+            node3.set_configuration_options(values={
```
TheRealFalcon: Notice we're only setting password auth on node3 here

aboudreault:
ah! I see. That explains the issue didn't get caught in existing auth tests. Thanks for the clarification!

TheRealFalcon:
Also, I ran the new test on master and the test failed

PR #1030, opened 2019-07-16
[PYTHON-1031] Remove "frozen" from metadata of function arguments

TheRealFalcon:
https://datastax-oss.atlassian.net/browse/PYTHON-1031

Functions don't accept frozen as part of the type arguments even though the collections get frozen after creation. To properly return the CQL statement we need to strip frozen from the type we receive from the server.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,7 +9,7 @@ Features
 * Remove invalid warning in set_session when we initialize a default connection (PYTHON-1104)
 Bug Fixes
 ---------
-* Fix 'as_cql_query UDF/UDA parameters incorrectly includes "frozen" if arguments are collections' (PYTHON-1031)
+* as_cql_query UDF/UDA parameters incorrectly includes "frozen" if arguments are collections (PYTHON-1031)
 * Fix incorrect metadata for compact counter tables (PYTHON-1100)
 * Call ConnectionException with correct kwargs (PYTHON-1117)
 * Can't connect to clusters built from source because version parsing doesn't handle 'x.y-SNAPSHOT' (PYTHON-1118)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Features
 * Remove invalid warning in set_session when we initialize a default connection (PYTHON-1104)
 Bug Fixes
 ---------
+* Fix 'as_cql_query UDF/UDA parameters incorrectly includes "frozen" if arguments are collections' (PYTHON-1031)
 * Fix incorrect metadata for compact counter tables (PYTHON-1100)
 * Call ConnectionException with correct kwargs (PYTHON-1117)
 * Can't connect to clusters built from source because version parsing doesn't handle 'x.y-SNAPSHOT' (PYTHON-1118)
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -28,6 +28,7 @@ the corresponding CQL or Cassandra type strings.
 # .from_cql_literal() and .as_cql_literal() classmethods (or whatever).
 
 from __future__ import absolute_import  # to enable import io from stdlib
+import ast
 from binascii import unhexlify
 import calendar
 from collections import namedtuple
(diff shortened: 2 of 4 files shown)


cassandra/metadata.py
```diff
@@ -2833,3 +2833,44 @@ def group_keys_by_replica(session, keyspace, table, keys):
...
+            elif token == '>' and in_frozen:
+                depth -= 1
+                if depth == 0:
+                    in_frozen = False
+                    skip_current_token = True
+            if not skip_current_token:
+                updated_tokens.append(token)
+        return _strip(updated_tokens)
+
+    return ''.join(_strip(scanned_tokens))
```
aboudreault: What about using regex? Could something like this work for this case?
```
FROZEN_REMOVAL_REGEX = re.compile(r'frozen<(.+)>')
def strip_frozen(cass_type):
    while 'frozen' in cass_type: 
        cass_type = re.sub(FROZEN_REMOVAL_REGEX, r'\1', cass_type)
    return cass_type
```

CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ Features
 
 Bug Fixes
 ---------
+* Remove "frozen" from metadata of function arguments (PYTHON-1031)
```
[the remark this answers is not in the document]
TheRealFalcon: So in general, anything in the changelog as a direct result of a ticket should use the ticket title?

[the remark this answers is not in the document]
aboudreault: In general we try to follow that. If the ticket title is wrong.. we fix it with the proper description. When a bug ticket result to a new feature or improvement, which fix the bug... we create another ticket and 2 changelog entries.

cassandra/metadata.py
TheRealFalcon: Unfortunately no, regex can't deal with arbitrary nesting. E.g.,
`frozen<tuple<int, frozen<tuple<int, text>>, int>>`
A non-greedy regex would remove the first end bracket, and a greedy one would remove the last one. Neither are correct.

CHANGELOG.rst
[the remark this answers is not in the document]
TheRealFalcon: Force pushed the update since it was just this line

cassandra/metadata.py
aboudreault: The function is a bit hard to read... not really because how it is done but mostly because it's something messy to do.  Maybe it we be possible to do something in a generic way and split it a bit more. e.g:

1. A function that only takes care to return a list of the types, grouped by container/collection
2. A recursive function that only takes care to return the inner of a cass type. tuple<THIS>, frozen<THIS>.

Just brainstorming.. we can discuss this more. I think it could be useful to have that for the future... since cql_types_from_string is not really smart.

Also could we move this stuff in cqltypes.py?

aboudreault:
Looks very clean, I like it. I will review when you assign me.

CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ Features
 
 Bug Fixes
 ---------
+* Fix 'as_cql_query UDF/UDA parameters incorrectly includes "frozen" if arguments are collections' (PYTHON-1031)
```
aboudreault: `as_cql_query UDF/UDA parameters incorrectly includes "frozen" if arguments are collections (PYTHON-1031)`

cassandra/cqltypes.py
```diff
@@ -121,6 +122,76 @@ def __new__(metacls, name, bases, dct):
...
+        (r'[, ]', lambda s, t: t),
+        (r'".*?"', lambda s, t: "'{}'".format(t)),
+    ))
+
+    scanned_tokens = scanner.scan(cql_string)[0]
+    hierarchy = ast.literal_eval(''.join(scanned_tokens))
+    if isinstance(hierarchy, str):
+        return [hierarchy]
+    else:
+        return list(hierarchy)
```
aboudreault: nit: `return [hierarchy] if isinstance(hierarchy, str) else list(hierarchy)`

cassandra/cqltypes.py
```diff
@@ -121,6 +122,76 @@ def __new__(metacls, name, bases, dct):
...
+        int -> ['int']
+        frozen<tuple<text, int>> -> ['frozen', ['tuple', ['text', 'int']]]
+    """
+    scanner = re.Scanner((
+        (r'[a-zA-Z0-9_]+', lambda s, t: "'{}'".format(t)),
+        (r'<', lambda s, t: ', ['),
+        (r'>', lambda s, t: ']'),
+        (r'[, ]', lambda s, t: t),
+        (r'".*?"', lambda s, t: "'{}'".format(t)),
+    ))
```
aboudreault: Interesting idea. I like the fact that everything inside  a `<>` are encapsulated as a python list.

cassandra/cqltypes.py
```diff
@@ -121,6 +122,76 @@ def __new__(metacls, name, bases, dct):
...
+def _strip_frozen_from_python(types):
+    """
+    Given a python list representing a cql type, removes 'frozen'
+    Example:
+        ['frozen', ['tuple', ['text', 'int']]] -> ['tuple', ['text', 'int']]
+    """
+    while 'frozen' in types:
+        index = types.index('frozen')
+        types = types[:index] + types[index + 1] + types[index + 2:]
+    new_types = [_strip_frozen_from_python(item) if isinstance(item, list) else item for item in types]
```
aboudreault: Just to be sure, will this also remove nested `frozen`?

TheRealFalcon: Yep, because of the recursion here
`new_types = [_strip_frozen_from_python(item) ... `

PR #1031, opened 2019-07-17
[PYTHON-1110] Fix ForcedHostSwitchPolicy not always choosing correct …

TheRealFalcon:
…host in prepared tests.
https://datastax-oss.atlassian.net/browse/PYTHON-1110

`ForcedHostSwitchPolicy` was causing failures if more than one query was happening between statement executions. Replaced it with a `ForcedHostIndexPolicy` that allows us to hardcode the node to send requests to, and updated the tests that used the old policy accordingly.

### DIFF
--- a/tests/integration/standard/test_query.py
+++ b/tests/integration/standard/test_query.py
@@ -24,7 +24,7 @@ from cassandra import ProtocolVersion
 from cassandra import ConsistencyLevel, Unavailable, InvalidRequest, cluster
 from cassandra.query import (PreparedStatement, BoundStatement, SimpleStatement,
                              BatchStatement, BatchType, dict_factory, TraceUnavailable)
-from cassandra.cluster import Cluster, NoHostAvailable, ExecutionProfile
+from cassandra.cluster import Cluster, NoHostAvailable, ExecutionProfile, EXEC_PROFILE_DEFAULT
 from cassandra.policies import HostDistance, RoundRobinPolicy, WhiteListRoundRobinPolicy
 from tests.integration import use_singledc, PROTOCOL_VERSION, BasicSharedKeyspaceUnitTestCase, \
     greaterthanprotocolv3, MockLoggingHandler, get_supported_protocol_versions, local, get_cluster, setup_keyspace, \
@@ -37,6 +37,7 @@ import random
 import re
 
 import mock
+import six
 
 
 log = logging.getLogger(__name__)
@@ -46,7 +47,7 @@ def setup_module():
     if not USE_CASS_EXTERNAL:
         use_singledc(start=False)
         ccm_cluster = get_cluster()
-        ccm_cluster.clear()
+        ccm_cluster.stop()
         # This is necessary because test_too_many_statements may
         # timeout otherwise
         config_options = {'write_request_timeout_in_ms': '20000'}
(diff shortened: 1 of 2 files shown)


tests/integration/standard/test_query.py
```diff
@@ -46,7 +47,7 @@ def setup_module():
     if not USE_CASS_EXTERNAL:
         use_singledc(start=False)
         ccm_cluster = get_cluster()
-        ccm_cluster.clear()
+        ccm_cluster.stop()
```
TheRealFalcon: Other tests use `.stop()` and I see no reason why we need to clear data here. It was preventing me from the running this module in isolation, because if you call `.clear()` with no pre-existing cluster data it'll throw an exception.

tests/integration/standard/test_query.py
```diff
@@ -495,33 +502,30 @@ def test_prepared_metadata_generation(self):
...
-        logger.addHandler(mock_handler)
-        select_statement = session.prepare("SELECT * FROM system.local")
-        session.execute(select_statement)
-        session.execute(select_statement)
-        session.execute(select_statement)
-        self.assertEqual(2, mock_handler.get_message_count('debug', "Re-preparing"))
-
+        select_statement = session.prepare("SELECT k FROM test3rf.test WHERE k = ?")
+        for host in clus.metadata.all_hosts():
+            session.execute(select_statement, (1, ), host=host)
```
TheRealFalcon: Since we can just hardcode the host here, this test didn't actually need the new policy in this test. The other two tests need the policy because we need to ensure prepares are happening on different nodes than the queries, and we can't pass a host to the prepare statement.

tests/integration/standard/test_query.py
```diff
@@ -551,9 +559,15 @@ def test_prepare_batch_statement(self):
...
         # is not prepared
+        policy.set_host(1)
         batch_statement = BatchStatement(consistency_level=ConsistencyLevel.ONE)
         batch_statement.add(insert_statement, (1, 2))
         session.execute(batch_statement)
+
+        # To verify our test assumption that queries are getting re-prepared properly
+        self.assertEqual(1, self.mock_handler.get_message_count('debug', "Re-preparing"))
+
+        policy.set_host(2)
```
aboudreault: What is the difference in this test, do we really need to use that custom policy? We could just use the `host` arg also?

tests/integration/standard/test_query.py
```diff
@@ -590,21 +608,29 @@ def test_prepare_batch_statement_after_alter(self):
 
         values_to_insert = [(1, 2, 3), (2, 3, 4), (3, 4, 5), (4, 5, 6)]
 
-        # We query the three hosts in order (due to the ForcedHostSwitchPolicy)
+        # We query the three hosts in order (due to the ForcedHostIndexPolicy)
         # the first three queries will have to be repreapred and the rest should
         # work as normal batch prepared statements
         for i in range(10):
+            policy.set_host(i % 3)
```
aboudreault: same question than above. can we do that cycling over the `metadata.all_hosts()` list and use `host`?

tests/integration/standard/test_query.py
TheRealFalcon: This particular line isn't necessary (I'll remove it), but the line on 562 is. The problem with this test is we need the batch statement to execute on a different host than the prepare statement. Unless there is a way to control the host that the prepare statement is going to, the policy is necessary for this test.

aboudreault: oh I see now. That's the part I missed. Indeed we need it to ensure it will be applied when preparing. Thanks for the explanation!

PR #1033, opened 2019-07-24
Connection execution profiles PYTHON-1009

TheRealFalcon:
This is NOT ready for review.

PR is currently to facilitate discussion

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1148,6 +1148,7 @@ class Cluster(object):
             raise ValueError("Cannot add execution profiles when legacy parameters are set explicitly.")
         if name in self.profile_manager.profiles:
             raise ValueError("Profile {} already exists".format(name))
+        self._config_mode = _ConfigMode.PROFILES
         contact_points_but_no_lbp = (
             self._contact_points_explicit and not
             profile._load_balancing_policy_explicit)
--- a/cassandra/cqlengine/connection.py
+++ b/cassandra/cqlengine/connection.py
@@ -17,12 +17,13 @@ import logging
 import six
 import threading
 
-from cassandra.cluster import Cluster, _NOT_SET, NoHostAvailable, UserTypeDoesNotExist, ConsistencyLevel
+
+from cassandra.cluster import Cluster, _NOT_SET, NoHostAvailable, UserTypeDoesNotExist, ConsistencyLevel, _ConfigMode, \
+    ExecutionProfile
 from cassandra.query import SimpleStatement, dict_factory
 
 from cassandra.cqlengine import CQLEngineException
 from cassandra.cqlengine.statements import BaseCQLStatement
-
 
 log = logging.getLogger(__name__)
 
@@ -39,6 +40,8 @@ _connections = {}
 # and because sessions may be replaced, we must register UDTs here, in order
 # to have them registered when a new session is established.
 udt_by_keyspace = defaultdict(dict)
+
+CQLENGINE_PROFILE = '_cqlengine'
 
 
 def format_log_context(msg, connection=None, keyspace=None):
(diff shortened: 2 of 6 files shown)


cassandra/cqlengine/connection.py
```diff
@@ -181,15 +230,13 @@ def register_connection(name, hosts=None, consistency=None, lazy_connect=False,
             raise CQLEngineException(
                 "Session configuration arguments and 'session' argument are mutually exclusive"
             )
-        conn = Connection.from_session(name, session=session)
-        conn.setup_session()
```
TheRealFalcon: `conn.setup_session()` is called after in `Connection.from_session()`

cassandra/cqlengine/connection.py
```diff
@@ -181,15 +230,13 @@ def register_connection(name, hosts=None, consistency=None, lazy_connect=False,
...
+        conn = Connection.from_session(name, session=session, execution_profile=execution_profile)
     else:  # use hosts argument
-        if consistency is None:
-            consistency = ConsistencyLevel.LOCAL_ONE
```
TheRealFalcon: Moved this to `Connection.setup_session` so it is set consistently

cassandra/cqlengine/connection.py
```diff
@@ -267,9 +314,10 @@ def set_session(s):
     """
     Configures the default connection with a preexisting :class:`cassandra.cluster.Session`
 
-    Note: the mapper presently requires a Session :attr:`~.row_factory` set to ``dict_factory``.
-    This may be relaxed in the future
+    Note: the mapper requires a Session :attr:`~.row_factory` set to ``dict_factory``
+     if the session is configured with legacy configuration instead of Execution Profiles
     """
+    # TODO: Can we link to the Execution Profiles page in that docstring?
```
TheRealFalcon: Is this possible?

tests/integration/cqlengine/connections/test_connection.py
```diff
@@ -127,3 +130,74 @@ def test_connection_session_switch(self):
         connection.set_session(self.session2)
         self.assertEqual(1, TestConnectModel.objects.count())
         self.assertEqual(TestConnectModel.objects.first(), TCM2)
+
+
+class LegacyConnectionsModel(Model):
```
TheRealFalcon: Not sure the best place for these tests. We have a `test_connection.py` and `test_connections.py`, and I'm not really sure the difference between them.

cassandra/cluster.py
```diff
@@ -1148,6 +1148,7 @@ def add_execution_profile(self, name, profile, pool_wait_timeout=5):
             raise ValueError("Cannot add execution profiles when legacy parameters are set explicitly.")
         if name in self.profile_manager.profiles:
             raise ValueError("Profile {} already exists".format(name))
+        self._config_mode = _ConfigMode.PROFILES
```
TheRealFalcon: Snuck this in in a different commit.

cassandra/cqlengine/connection.py
```diff
@@ -281,8 +329,6 @@ def set_session(s):
     if conn.session:
         log.warning("configuring new default connection for cqlengine when one was already set")
 
-    if s.row_factory is not dict_factory:
-        raise CQLEngineException("Failed to initialize: 'Session.row_factory' must be 'dict_factory'.")
```
TheRealFalcon: Moved this into the Connection class so it is called consistently. If using a different way to initialize the connection (e.g., `Connection.from_session()`), this check wasn't happening.

TheRealFalcon:
My original goal was to setup cqlengine to use execution profiles by default, but revert to legacy settings if they're specified (you know...because they're legacy...).

In the non-legacy case, regardless of if we're passed an `ExecutionProfile` or we create one ourselves, we create/clone one specific to cqlengine so we can set the `row_factory` independent of how the connection/session was configured.

One consequence is that running `get_session()` and running a query against it won't use the same execution profile that cqlengine is using. So a test like this fails because it's expecting a `dict_factory` result but getting a `named_tuple_factory` result instead:
https://github.com/datastax/python-driver/blob/python-1009-connection-execution-profiles/tests/integration/cqlengine/test_ttl.py#L171

Would that be considered a breaking change? If so, should we just plan on using legacy settings by default until the next major?

TheRealFalcon:
Closing this and using #1046 instead. Keeping this branch around for the next major.

PR #1034, opened 2019-07-25
[PYTHON-1111] Disallow multiple cluster objects sharing same executio…

TheRealFalcon:
…n profile


TheRealFalcon:
`Do a minimal class-level execution profile tracking in the ProfileManager class`
I thought about that, but wouldn't that introduce a breaking change? There's currently nothing preventing me from using the same execution profile more than once per cluster, and nothing _should_ be preventing me from doing that, right? Tracking duplicates across instances _only_ gets a lot more complicated, which is why I went this route.

aboudreault:
That is correct. But we can probably do a minimal tracking per instance in the ProfileManager class? We can put this one on hold and I will check deeper when I'm back. I think that if we cannot do something very simple to avoid this to happen.. I prefer to not fix since it's very not something critical and 99% of the application use only one Cluster instance. Some other thoughts:

- We could also ensure our custom built-in policies that `populate` is only called once, with an internal flag is_populated... https://github.com/datastax/python-driver/blob/master/cassandra/policies.py#L118
- Maybe we should end up to only improve the ExecutionProfile documentation...

TheRealFalcon:
@aboudreault I replaced the code with a small docs snippet about usage. Not sure I put it in the best place or if there's more that needs to be said, so let me know if you think it could be better worded or placed.

docs/execution_profiles.rst
```diff
@@ -9,6 +9,10 @@ The legacy configuration remains intact, but legacy and Execution Profile APIs
 cannot be used simultaneously on the same client ``Cluster``. Legacy configuration
 will be removed in the next major release (4.0).
 
+Instances of execution profiles and the settings used by them should be unique across ``Cluster`` instances.
+For example, an execution profile should not be applied to more than one ``Cluster`` instance,
+and a ``LoadBalancingPolicy`` should not belong to more than one execution profile.
+
```
aboudreault: suggestion:

`An execution profile and its parameters cannot be shared across multiple ``Cluster`` instances`. For example, an execution profile, its ``LoadBalancingPolicy`` and its ``RetryPolicy`` should not be applied to more than one ``Cluster`` instance.

PR #1035, opened 2019-07-26
[PYTHON-1104] Remove invalid warning in set_session when we initializ…

TheRealFalcon:
…e a default connection

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ Features
 --------
 * Add Python 3.7 support (PYTHON-1016)
 * Future-proof Mapping imports (PYTHON-1023)
-
+* Remove invalid warning in set_session when we initialize a default connection (PYTHON-1104)
 Bug Fixes
 ---------
 * Fix incorrect metadata for compact counter tables (PYTHON-1100)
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
_Forced pushed a clean rebase of master, but no code changes_

We're talking about a warning when calling `set_session()`, right? That's what the ticket says. But I'm not able to trigger any warning (and I don't see how I could) from calling `set_session()` other than the one I removed. Also, the warning I removed is in the call path every single time you call `set_session()` regardless of if a default connection existed because setting up any connection (including the default one) also initializes a session, so I figured it was an invalid warning.

```
>>> # master
>>> import logging
>>> from cassandra.cluster import Cluster
>>> from cassandra.cqlengine.connection import set_session
>>> 
>>> log = logging.getLogger()
>>> log.addHandler(logging.StreamHandler())
>>> log.setLevel('WARNING')
>>> # -----------------------------------------------------------------------
... 
>>> cluster = Cluster()
>>> session = cluster.connect()
>>> set_session(session)
configuring new default connection for cqlengine when one was already set
>>> 
```
```
>>> # branch: python-1104-invalid-set-session-warning
>>> import logging
>>> from cassandra.cluster import Cluster
>>> from cassandra.cqlengine.connection import set_session
>>> 
>>> log = logging.getLogger()
>>> log.addHandler(logging.StreamHandler())
>>> log.setLevel('WARNING')
>>> # -----------------------------------------------------------------------
... 
>>> cluster = Cluster()
>>> session = cluster.connect()
>>> set_session(session)
>>> 
```

PR #1041, opened 2019-08-13
Add docs clarification to host argument of Session.execute

TheRealFalcon:


cassandra/cluster.py
```diff
@@ -2245,8 +2245,9 @@ def execute(self, query, parameters=None, timeout=_NOT_SET, trace=False,
 
         `paging_state` is an optional paging state, reused from a previous :class:`ResultSet`.
 
-        `host` is the :class:`pool.Host` that should handle the query. Using this is discouraged except in a few
-        cases, e.g., querying node-local tables and applying schema changes.
+        `host` is the :class:`cassandra.pool.Host` that should handle the query. If the host specified is down or hasn't
+        finished initialization, the query will fail with :class:`NoHostAvailable`. Using this is
```
aboudreault: suggestion: `If the host specified is down or not yet connected` ...

PR #1044, opened 2019-08-19
PYTHON-1016: Python 3.7 eventlet reactor fix

aboudreault:
This is part of the python 3.7 support for the next release. Unfortunately, the Eventlet reactor is not compatible with the built-in concurrent.futures.ThreadPoolExecutor. 

This Pull Request basically detects this situation and raise an error to the user. To use Python 3.7 and Eventlet, the user has to install an additional package, so we can use a custom thread pool made for it.

@aholmberg Can you do a quick review?

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -1061,7 +1061,7 @@ class Cluster(object):
             HostDistance.REMOTE: DEFAULT_MAX_CONNECTIONS_PER_REMOTE_HOST
         }
 
-        self.executor = ThreadPoolExecutor(max_workers=executor_threads)
+        self.executor = self._create_thread_pool_executor(max_workers=executor_threads)
         self.scheduler = _Scheduler(self.executor)
 
         self._lock = RLock()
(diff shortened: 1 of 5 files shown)


cassandra/cluster.py
```diff
@@ -1076,6 +1076,43 @@ def __init__(self,
...
+        Python 3.7 and Eventlet cause the `concurrent.futures.ThreadPoolExecutor`
+        to hang indefinitely. In that case, the user needs to have the `futurist`
+        package so we can use the `futurist.GreenThreadPoolExecutor` class instead.
+
+        :param kwargs: All keyword args are passed to the ThreadPoolExecutor constructor.
+        :return: A ThreadPoolExecutor instance.
+        """
+        tpe_class = ThreadPoolExecutor
+        if sys.version_info[0] == 3 and sys.version_info[1] >= 7:
+            try:
```
aholmberg: Possibly use the internal utility function?: https://github.com/datastax/python-driver/blob/master/cassandra/cluster.py#L80-L84

cassandra/cluster.py
```diff
@@ -1076,6 +1076,43 @@ def __init__(self,
...
+
+        Python 3.7 and Eventlet cause the `concurrent.futures.ThreadPoolExecutor`
+        to hang indefinitely. In that case, the user needs to have the `futurist`
+        package so we can use the `futurist.GreenThreadPoolExecutor` class instead.
+
+        :param kwargs: All keyword args are passed to the ThreadPoolExecutor constructor.
+        :return: A ThreadPoolExecutor instance.
+        """
+        tpe_class = ThreadPoolExecutor
+        if sys.version_info[0] == 3 and sys.version_info[1] >= 7:
```
aholmberg: Maybe we're expecting this to be dead code by then, but does it make sense to make this `sys.version_info[0] == 3`?

cassandra/cluster.py
```diff
@@ -1076,6 +1076,43 @@ def __init__(self,
...
+                # Eventlet is not available or can't be detected
+                return tpe_class(**kwargs)
+
+            if is_eventlet:
+                try:
+                    from futurist import GreenThreadPoolExecutor
+                    tpe_class = GreenThreadPoolExecutor
+                except ImportError:
+                    # futurist is not available
+                    log.error(
```
aholmberg: Should we raise an exception with this message, instead of logging error and raising the less-specific `ImportError`?

cassandra/cluster.py
aboudreault: The reason why I do it like that is because the user can use eventlet in two ways:
1- monkey patching in their app before creating a cluster. In that case, the utility function would work.
2- Only specifying the connection_class at the cluster level, which will take care to monkey patch when the reactor is initialized. In that case, the utility function wont work since nothing is patched yet. In fact... that might be a oversight cause the documentation mentions that the implementation assume that all eventlet monkey patching is active but the code call monkey_patch. The GeventConnection in example doesn't care of to monkey patch at all.

cassandra/cluster.py
aboudreault: Yep, we can do that. +1

cassandra/cluster.py
aboudreault: Can you clarify this? My goal is to *only* affect versions that have the bug.

aboudreault: ah, you are talking about Python 4? Didn't even think about that... but yeah maybe it's safer to just use `>= 3`.

cassandra/cluster.py
aboudreault: I agree. I will create a ticket to remove that.

PR #1047, opened 2019-09-02
Fix twisted tests (PYTHON-1037)

TheRealFalcon:
Updated twisted dependency so windows install doesn't take 30 minutes. In testing the changes, also noticed a test failing for all twisted builds, so fixed the test as well.

Windows build:
https://ci.appveyor.com/project/DataStax/python-driver/builds/27114069

Jenkins build customized to use twisted:
https://jenkins-drivers.build.dsinternal.org/view/python-driver/job/datastax.python-driver.dev-python-1037-twisted.commit_branches_dev/

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -48,11 +48,11 @@ schedules:
     branches:
       include: [/dev-python.*/]
     env_vars: |
-      EVENT_LOOP_MANAGER='libev'
+      EVENT_LOOP_MANAGER='twisted'
       EXCLUDE_LONG=1
     matrix:
       exclude:
-        - python: [2.7, 3.4, 3.6, 3.7]
+        - python: [3.6, 3.7]
         - cassandra: ['2.0', '2.1', '2.2', '3.0', 'test-dse']
 
   release_test:
--- a/build.yaml
+++ b/build.yaml
@@ -48,11 +48,11 @@ schedules:
     branches:
       include: [/dev-python.*/]
     env_vars: |
-      EVENT_LOOP_MANAGER='twisted'
+      EVENT_LOOP_MANAGER='libev'
       EXCLUDE_LONG=1
     matrix:
       exclude:
-        - python: [3.6, 3.7]
+        - python: [2.7, 3.4, 3.6, 3.7]
         - cassandra: ['2.0', '2.1', '2.2', '3.0', 'test-dse']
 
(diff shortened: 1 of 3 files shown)


tests/integration/simulacron/test_endpoint.py
```diff
@@ -27,23 +27,24 @@
...
+        self._port = port
 
     @property
     def address(self):
         return self._address
 
     @property
     def port(self):
-        return None
+        return self._port
```
TheRealFalcon: port can't be None when using Twisted:
https://github.com/datastax/python-driver/blob/master/cassandra/io/twistedreactor.py#L269

tests/integration/simulacron/test_endpoint.py
```diff
@@ -99,7 +100,8 @@ def test_custom_endpoint(self):
         cluster = Cluster(
             contact_points=[AddressEndPoint('127.0.0.1')],
             protocol_version=PROTOCOL_VERSION,
-            endpoint_factory=AddressEndPointFactory()
+            endpoint_factory=AddressEndPointFactory(),
+            compression=False,
```
TheRealFalcon: Simulacron barfs on compression, which I had enabled locally. It passes on CI because compression isn't installed.

build.yaml
```diff
@@ -48,11 +48,11 @@ schedules:
     branches:
       include: [/dev-python.*/]
     env_vars: |
-      EVENT_LOOP_MANAGER='libev'
+      EVENT_LOOP_MANAGER='twisted'
       EXCLUDE_LONG=1
     matrix:
       exclude:
-        - python: [2.7, 3.4, 3.6, 3.7]
+        - python: [3.6, 3.7]
```
aboudreault: is this temporary changes?

test-requirements.txt
```diff
@@ -7,7 +7,8 @@ unittest2
 pytz
 sure
 pure-sasl
-twisted[tls]==19.2.1
+twisted; python_version >= '3.5'
+twisted[tls]==19.2.1; python_version < '3.5'
```
aboudreault: good solution!

build.yaml
TheRealFalcon: Yes, just temporary branch build changes.

PR #1048, opened 2019-09-16
[Python 1044] Driver hangs/deadlock if all connections dropped by heartbeat whilst request in flight and request times out

nschrader:
This PR concerns https://datastax-oss.atlassian.net/projects/PYTHON/issues/PYTHON-1044
It contains a fix, unit and integration tests. I had some problems with integration test configs, so 
6ebe0d8 addresses this.
Lastly, I would like to mention that this issue is all about an edge case and therefore extremely difficult to reproduce. However, some microservices of our business software got stuck in production several times in the last few months due to this.

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3722,11 +3722,14 @@ class ResponseFuture(object):
         if self._connection is not None:
             try:
                 self._connection._requests.pop(self._req_id)
-            # This prevents the race condition of the
-            # event loop thread just receiving the waited message
-            # If it arrives after this, it will be ignored
+            # PYTHON-1044
+            # This request might have been removed from the connection after the latter was defunct by heartbeat.
+            # We should still raise OperationTimedOut to reject the future so that the main event thread will not
+            # wait for it endlessly
             except KeyError:
-                return
+                key = "Connection defunct by heartbeat"
+                errors = {key: "Client request timeout. See Session.execute[_async](timeout)"}
+                self._set_final_exception(OperationTimedOut(errors, self._current_host))
 
             pool = self.session._pools.get(self._current_host)
             if pool and not pool.is_shutdown:
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3730,6 +3730,7 @@ class ResponseFuture(object):
                 key = "Connection defunct by heartbeat"
                 errors = {key: "Client request timeout. See Session.execute[_async](timeout)"}
                 self._set_final_exception(OperationTimedOut(errors, self._current_host))
+                return
 
             pool = self.session._pools.get(self._current_host)
             if pool and not pool.is_shutdown:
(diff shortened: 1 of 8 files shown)


aboudreault:
Thanks @nschrader for this PR. We are about to plan a new minor release and I will definitively add this one to the roadmap.

aboudreault:
@nschrader I am currently playing with this PR. Should I be able to reproduce the deadlock with your test? (without your fix)

cassandra/cluster.py
```diff
@@ -3722,10 +3722,14 @@ def _on_timeout(self, _attempts=0):
...
-            # event loop thread just receiving the waited message
-            # If it arrives after this, it will be ignored
+            # PYTHON-1044
+            # This request might have been removed from the connection after the latter was defunct by heartbeat.
+            # We should still raise OperationTimedOut to reject the future so that the main event thread will not
+            # wait for it endlessly
             except KeyError:
+                key = "Connection defunct by heartbeat"
+                errors = {key: "Client request timeout. See Session.execute[_async](timeout)"}
+                self._set_final_exception(OperationTimedOut(errors, self._current_host))
```
nschrader: Comment out to dead lock with unit and integration tests

aboudreault: That's what I meant by `without your fix`. Trying to reproduce the deadlock with simulacron...

nschrader: So I guess it's not deadlocking for you ? I tried on several machines with a freshly cloned repository using `simulacron-standalone-0.8.9.jar`. I already noticed that it's deadlocking only 99% of the time. Can you tell me more about your set-up ?

Is your simularcron configuration creating a cluster of 3 hosts (127.0.0.1-3) ? Maybe you can try to increase `lag = 0.05` to `lag = 0.1` or `lag = 0.2` within the integration test.

aboudreault: Thanks... I will make sure I'm using the same simulacron version and vary the lag to see.

aboudreault: What event loop are you using? asyncore, libev? What Python 2.7 version? Could you also attach the stacktrace of this test when it deadlocks?

aboudreault: Looks like I can reproduce the deadlock in our jenkins CI platform, but not locally. Better than nothing...

tests/integration/simulacron/test_connection.py
```diff
@@ -178,6 +178,79 @@ def test_callbacks_and_pool_when_oto(self):
         errback.assert_called_once()
         callback.assert_not_called()
 
+    def test_heartbeat_defunct_deadlock(self):
```
aboudreault: I can't reproduce this issue if the python driver is not build with cython. Add the `@cythontest` decorator here. --> from tests.unit.cython.utils import cythontest

PR #1049, opened 2019-09-29
Use 4.0 schema parser in 4 alpha and snapshot builds

clohfink:


TheRealFalcon:
Thanks for the contribution @clohfink ! This will be in our next release.

PR #1051, opened 2019-10-09
[PYTHON-1162] Allow passing ssl context for Eventlet

TheRealFalcon:
The reason for the instance checks in the eventlet code is that eventlet will monkeypatch the built in ssl module to use pyOpenSSL under the hood. We want to utilize that if we're relying strictly on ssl_options, but if we're passing an SSLContext, we must use the pyOpenSSL APIs.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -227,6 +227,7 @@ build:
           "tests/integration/standard/test_metrics.py"
           "tests/integration/standard/test_query.py"
           "tests/integration/simulacron/test_endpoint.py"
+          "tests/integration/long/test_ssl.py"
         )
         EVENT_LOOP_MANAGER=$EVENT_LOOP_MANAGER CCM_ARGS="$CCM_ARGS" CASSANDRA_VERSION=$CCM_CASSANDRA_VERSION MAPPED_CASSANDRA_VERSION=$MAPPED_CASSANDRA_VERSION VERIFY_CYTHON=$FORCE_CYTHON nosetests -s -v --logging-format="[%(levelname)s] %(asctime)s %(thread)d: %(message)s" --with-ignore-docstrings --with-xunit --xunit-file=standard_results.xml ${EVENT_LOOP_TESTS[@]} || true
         exit 0
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -78,12 +78,12 @@ from cassandra.datastax import cloud as dscloud
 try:
     from cassandra.io.twistedreactor import TwistedConnection
 except ImportError:
-    TwistedConnection = type(None)
+    TwistedConnection = None
 
 try:
     from cassandra.io.eventletreactor import EventletConnection
 except ImportError:
-    EventletConnection = type(None)
+    EventletConnection = None
 
 try:
     from weakref import WeakSet
@@ -925,10 +925,9 @@ class Cluster(object):
                 raise ValueError("contact_points, endpoint_factory, ssl_context, and ssl_options "
                                  "cannot be specified with a cloud configuration")
 
-            cloud_config = dscloud.get_cloud_config(
-                cloud,
-                create_pyopenssl_context=issubclass(self.connection_class, (TwistedConnection, EventletConnection))
-            )
+            uses_twisted = TwistedConnection and issubclass(self.connection_class, TwistedConnection)
+            uses_eventlet = EventletConnection and issubclass(self.connection_class, EventletConnection)
+            cloud_config = dscloud.get_cloud_config(cloud, create_pyopenssl_context=uses_twisted or uses_eventlet)
 
             ssl_context = cloud_config.ssl_context
             ssl_options = {'check_hostname': True}
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -37,16 +37,6 @@ from threading import Lock, RLock, Thread, Event
 import weakref
 from weakref import WeakValueDictionary
 
-try:
-    from cassandra.io.twistedreactor import TwistedConnection
-except ImportError:
-    TwistedConnection = None
-
-try:
-    from weakref import WeakSet
-except ImportError:
-    from cassandra.util import WeakSet  # NOQA
-
 from cassandra import (ConsistencyLevel, AuthenticationFailed,
                        OperationTimedOut, UnsupportedOperation,
                        SchemaTargetType, DriverException, ProtocolVersion,
@@ -84,6 +74,21 @@ from cassandra.query import (SimpleStatement, PreparedStatement, BoundStatement,
 from cassandra.timestamps import MonotonicTimestampGenerator
 from cassandra.compat import Mapping
 from cassandra.datastax import cloud as dscloud
+
+try:
+    from cassandra.io.twistedreactor import TwistedConnection
+except ImportError:
+    TwistedConnection = None
+
+try:
+    from cassandra.io.eventletreactor import EventletConnection
+except ImportError:
+    EventletConnection = None
+
+try:
+    from weakref import WeakSet
+except ImportError:
+    from cassandra.util import WeakSet  # NOQA
 
 
 def _is_eventlet_monkey_patched():
@@ -922,7 +927,7 @@ class Cluster(object):
 
             cloud_config = dscloud.get_cloud_config(
                 cloud,
-                create_pyopenssl_context=self.connection_class is TwistedConnection
+                create_pyopenssl_context=self.connection_class in [TwistedConnection, EventletConnection]
             )
 
             ssl_context = cloud_config.ssl_context
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -78,12 +78,12 @@ from cassandra.datastax import cloud as dscloud
 try:
     from cassandra.io.twistedreactor import TwistedConnection
 except ImportError:
-    TwistedConnection = None
+    TwistedConnection = type(None)
 
 try:
     from cassandra.io.eventletreactor import EventletConnection
 except ImportError:
-    EventletConnection = None
+    EventletConnection = type(None)
 
 try:
     from weakref import WeakSet
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -927,7 +927,7 @@ class Cluster(object):
 
             cloud_config = dscloud.get_cloud_config(
                 cloud,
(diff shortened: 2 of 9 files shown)


TheRealFalcon:
Normal (libev) test run here:
https://jenkins-drivers.build.dsinternal.org/job/datastax.python-driver.python-1162-ssl-context-eventlet.commit_branches/4/

Eventlet test run ran with a few failures in the test code that I just fixed, so the new run is here:
https://jenkins-drivers.build.dsinternal.org/job/datastax.python-driver.python-1162-ssl-context-eventlet.commit_branches/12/

aboudreault:
Sorry, can you clarify the scope of this PR? What feature do we add? Passing a Python SSLContext to evenlet? Passing a PyOpenSSL? Passing a GreenSSLContext?

TheRealFalcon:
pyOpenSSL. I added an example here:
https://github.com/datastax/python-driver/pull/1051/files#diff-435855f2b8c087a3bbf2c81a0b89bfd5R249

aboudreault:
So what is that GreenSSLContext?

aboudreault:
when monkeypatched.... a ssl.SSLContext becomes a GreenSSLContext ?

TheRealFalcon:
> when monkeypatched.... a ssl.SSLContext becomes a GreenSSLContext ?

Yes. Eventlet is using pyOpenSSL, but if you attempt to use the builtin ssl module, it will let you via monkeypatching, while converting the built in ssl calls (via these "green" objects) to pyOpenSSL. That's the only reason the existing code works and why I didn't add `eventlet` here:
https://github.com/datastax/python-driver/pull/1051/files#diff-e1e4ec7b2a2b7dffb3241b1360c97553R45

The monkeypatching isn't perfect though. It won't take a builtin SSL context, and it doesn't work with newer features of the ssl module like passing `check_hostname` to the wrapped socket. So we need to be able to accept a pyOpenSSL context while still supporting the old functionality.

build.yaml
```diff
@@ -29,11 +29,10 @@ schedules:
     branches:
       include: [/python.*/]
     env_vars: |
-      EVENT_LOOP_MANAGER='libev'
-      EXCLUDE_LONG=1
+      EVENT_LOOP_MANAGER='eventlet'
     matrix:
       exclude:
-        - python: [3.4, 3.6, 3.7]
+        - python: [3.4, 3.6]
```
TheRealFalcon: These build.yaml changes are a temporary commit so test eventlet on jenkins

TheRealFalcon:
@aboudreault Rebased and cleaned up the code a bit. Adding cloud connect was trivial since we already did it for twisted, so added that in here too. Should be ready to re-review.

tests/integration/long/test_ssl.py
```diff
@@ -42,7 +42,7 @@
 DRIVER_CERTFILE = os.path.abspath("tests/integration/long/ssl/driver.pem")
 DRIVER_CERTFILE_BAD = os.path.abspath("tests/integration/long/ssl/python_driver_bad.pem")
 
-USES_PYOPENSSL = "twisted" in EVENT_LOOP_MANAGER
+USES_PYOPENSSL = "twisted" in EVENT_LOOP_MANAGER or "eventlet" in EVENT_LOOP_MANAGER
 if "twisted" in EVENT_LOOP_MANAGER:
```
aboudreault: Do we need to include eventlet here?

cassandra/cluster.py
```diff
@@ -922,7 +927,7 @@ def __init__(self,
 
             cloud_config = dscloud.get_cloud_config(
                 cloud,
-                create_pyopenssl_context=self.connection_class is TwistedConnection
+                create_pyopenssl_context=self.connection_class in [TwistedConnection, EventletConnection]
```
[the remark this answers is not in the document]
TheRealFalcon: Can't we just fix it now with `create_pyopenssl_context=isinstance(self.connection_class, (TwistedConnection, EventletConnection))` ?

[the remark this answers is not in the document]
aboudreault: IIRC, self.connection_class is not an instance at this point... it's just the class. So we would have to use issubclass.. No objection if you want to fix it.

cassandra/io/eventletreactor.py
```diff
@@ -27,11 +26,24 @@
...
 
 log = logging.getLogger(__name__)
 
 
+def _check_pyopenssl():
+    if not _PYOPENSSL:
+        raise ImportError(
+            "{}, pyOpenSSL must be installed to enable "
+            "SSL support with the Twisted event loop".format(str(no_pyopenssl_error))
+        )
```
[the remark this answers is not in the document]
TheRealFalcon: Woah, good catch. Looked into it and behavior is different between 2.7 and 3.x. 2.7 this would work fine, but in 3.x the exception goes out of scope after the except clause unless it's been reassigned to a new variable.

cassandra/io/eventletreactor.py
```diff
@@ -89,6 +101,34 @@ def __init__(self, *args, **kwargs):
...
+
+    def _match_hostname(self):
+        if isinstance(self.ssl_context, eventlet.green.ssl.GreenSSLContext) or not self.ssl_context:
+            super(EventletConnection, self)._match_hostname()
+        else:
+            cert_name = self._socket.get_peer_certificate().get_subject().commonName
+            if cert_name != self.endpoint.address:
+                raise Exception("Hostname verification failed! Certificate name '{}' "
+                                "doesn't endpoint '{}'".format(cert_name, self.endpoint.address))
+
```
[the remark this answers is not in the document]
TheRealFalcon: Yes. test_ssl.py has some tests using `ssl_options` and other tests using `ssl_context`. The `ssl_options` tests will take the `if` branch while the `ssl_context` tests will take the `else` branch.

tests/integration/long/test_ssl.py
TheRealFalcon: No actually. Eventlet works fine with the built in `ssl.wrap_socket`, which is used when passing the old-style `ssl_options` arguments, but if constructing a context manually and wrapping it, you have to use pyOpenSSL. Clear as mud, right?

cassandra/io/eventletreactor.py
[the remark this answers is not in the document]
aboudreault: I was refering to different possible ssl context types. f I follow correctly, `self._wrap_socket_from_context()` is called only when there is a ssl_context. So there now 3 possibilities to test, right?
1. just ssl_options (legacy)
2. isinstance(self.ssl_context, eventlet.green.ssl.GreenSSLContext) == True
3. isinstance(self.ssl_context, eventlet.green.ssl.GreenSSLContext) == False

tests/integration/long/test_ssl.py
aboudreault: I think ;)

cassandra/io/eventletreactor.py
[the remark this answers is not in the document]
TheRealFalcon: Ah, right. 1 and 2 are the same. That first if condition you pointed out should never happen. I'll remove it.

How about I add something like a `self.uses_legacy_ssl_options = isinstance(self.ssl_context, eventlet.green.ssl.GreenSSLContext)` in the `__init__` and then check that instead. It would make the reading of those methods a little clearer.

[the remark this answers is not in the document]
aboudreault: +1 for that. Thanks

cassandra/io/eventletreactor.py
```diff
@@ -81,6 +94,7 @@ def service_timeouts(cls):
 
     def __init__(self, *args, **kwargs):
         Connection.__init__(self, *args, **kwargs)
+        self.uses_legacy_ssl_options = self.ssl_options and not self.ssl_context
```
TheRealFalcon: This is much easier than doing an instance check :)

aboudreault: I must admit that this helps my brain :D

cassandra/cluster.py
```diff
@@ -922,7 +927,7 @@ def __init__(self,
 
             cloud_config = dscloud.get_cloud_config(
                 cloud,
-                create_pyopenssl_context=self.connection_class is TwistedConnection
+                create_pyopenssl_context=issubclass(self.connection_class, (TwistedConnection, EventletConnection))
```
aboudreault: :+1:

TheRealFalcon:
@aboudreault One more small commit after doing some testing without the normal test requirements installed.
1. `issubclass()` requires classes, and None is not one, so have to use `type(None)` instead if imports fail.
2. Since we're not always creating the pyopenssl context anymore, if we can't import pyopenssl, I opted to fail with an error message rather than just ignore the import exception. Without it we were getting generic connection exceptions that are harder to debug.

cassandra/cluster.py
```diff
@@ -85,6 +75,21 @@
...
 
+try:
+    from cassandra.io.twistedreactor import TwistedConnection
+except ImportError:
+    TwistedConnection = type(None)
+
+try:
+    from cassandra.io.eventletreactor import EventletConnection
+except ImportError:
+    EventletConnection = type(None)
```
aboudreault: I think I'd prefer we keep that stuff where we use it. The reason is just because reading this is not very clear and we also don't understand why we assign to the None type rather than just None.

Note that I'm not sure if we should really call isubclass() with None types. What about just writing the logic at the place in the code with simple if statements?

TheRealFalcon: @aboudreault Howabout now?

cassandra/cluster.py
```diff
@@ -920,10 +925,9 @@ def __init__(self,
                 raise ValueError("contact_points, endpoint_factory, ssl_context, and ssl_options "
                                  "cannot be specified with a cloud configuration")
 
-            cloud_config = dscloud.get_cloud_config(
-                cloud,
-                create_pyopenssl_context=self.connection_class is TwistedConnection
-            )
+            uses_twisted = TwistedConnection and issubclass(self.connection_class, TwistedConnection)
+            uses_eventlet = EventletConnection and issubclass(self.connection_class, EventletConnection)
+            cloud_config = dscloud.get_cloud_config(cloud, create_pyopenssl_context=uses_twisted or uses_eventlet)
```
aboudreault: can't be more clear than that. :+1:

PR #1053, opened 2019-10-17
Fix broken link to map type documentation

oz123:
This fixes the previous link which is broken.

On a side note, maybe there should be `latest` docs website.
```
https://docs.datastax.com/en/dse/latest/cql/cql/cql_using/useMap.html
```
Should redirect to:
```
https://docs.datastax.com/en/6.7/latest/cql/cql/cql_using/useMap.html
```

And when the next version is released, latest should be updated, without updating links in the code.


aboudreault:
Thanks @oz123!

oz123:
@aboudreault  that was QUICK!

PR #1054, opened 2019-10-20
re-raising the CQLEngineException will fail on Python 3

hajs:
Instead a NameError will be thrown because unicode is unknown.
So I replaced unicode(exc) with six.text_type(exc) which will either use unicode or str.


aboudreault:
Hi @hajs, Can you create a ticket here and link your PR please? https://datastax-oss.atlassian.net

hajs:
Hi @aboudreault, you fand the related ticket here:
https://datastax-oss.atlassian.net/browse/PYTHON-1166

aboudreault:
Thanks @hajs !

PR #1057, opened 2019-11-06
Fix version parsing when two dashes in server version

TheRealFalcon:
PYTHON-1172.

Need to be able to parse versions like: 4.0.0-alpha2-SNAPSHOT


tests/unit/test_util_types.py
```diff
@@ -292,3 +292,5 @@ def test_version_compare(self):
         self.assertTrue(Version('4.0.0.build6-SNAPSHOT') > Version('4.0.0.build5-SNAPSHOT'))
         self.assertTrue(Version('4.0-SNAPSHOT2') > Version('4.0-SNAPSHOT1'))
         self.assertTrue(Version('4.0-SNAPSHOT2') > Version('4.0.0-SNAPSHOT1'))
+
+        self.assertTrue(Version('4.0.0-alpha1-SNAPSHOT') > Version('4.0.0-SNAPSHOT'))
```
tlasica: you should probably add Labs switch case here ;-)

tlasica: Also a reported problem info could be more comprehensive e.g. include problematic values and expected number.

PR #1061, opened 2019-12-07
Fix simple typo: versioni -> version

timgates42:
There is a small typo in cassandra/__init__.py.
Should read version rather than versioni.


aboudreault:
Thanks!

PR #1063, opened 2020-01-17
Move session id creation before insights reporter starts

TheRealFalcon:
When the insights reporter retrieves the startup data, it captures the session id. Since this runs in a separate thread, the session id is usually created by the time this capture actually runs. But it's a race, and sessionId can occassionally be captured as None.


TheRealFalcon:
Closing in favor of #1064

PR #1066, opened 2020-01-23
Test 68

TheRealFalcon:
Major changes mentioned in commits but summarized here:
 * Materialized view tests: You now can't create a MV with clustering order by unless you specify ALL of the clustering columns. So tests are updated to do that.
 * DuplicateRpcTest: We used to be able to update system.peers, but that's not allowed anymore, so converted it to a simulacron test. Updated another test to use mocks for the same reason.
 * Updated virtual keyspace test to not verify exact structure since that will change between versions. Rather, verify a few known values that should exist in all versions to verify we parsed the structure correctly.
 * Tombstone settings have moved under guardrails, and delete entire row intead of just cells in tombstone tests due to DB-2426.
 * Added a few optional env vars to allow management of an external cluster
 * DSE version check stuff

I'll squash down to one commit when merging

### DIFF
--- a/tests/integration/simulacron/test_cluster.py
+++ b/tests/integration/simulacron/test_cluster.py
@@ -16,14 +16,20 @@ try:
 except ImportError:
     import unittest  # noqa
 
-from tests.integration.simulacron import SimulacronCluster
-from tests.integration import (requiressimulacron, PROTOCOL_VERSION)
-from tests.integration.simulacron.utils import prime_query
+import logging
+from packaging.version import Version
+
+import cassandra
+from tests.integration.simulacron import SimulacronCluster, SimulacronBase
+from tests.integration import (requiressimulacron, PROTOCOL_VERSION, DSE_VERSION, MockLoggingHandler)
+from tests.integration.simulacron.utils import prime_query, start_and_prime_singledc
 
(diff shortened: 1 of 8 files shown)


tests/integration/simulacron/test_cluster.py
```diff
@@ -16,14 +16,20 @@
...
+from tests.integration import (requiressimulacron, PROTOCOL_VERSION, DSE_VERSION, MockLoggingHandler)
+from tests.integration.simulacron.utils import prime_query, start_and_prime_singledc
 
 from cassandra import (WriteTimeout, WriteType,
                        ConsistencyLevel, UnresolvableContactPoints)
-from cassandra.cluster import Cluster
+from cassandra.cluster import Cluster, ControlConnection
+
 
+PROTOCOL_VERSION = min(4, PROTOCOL_VERSION if (DSE_VERSION is None or DSE_VERSION >= Version('5.0')) else 3)
```
aboudreault: this is only a default value?

TheRealFalcon: Stole it from another simulacron test after getting protocol errors. Looks like simulacron barfs when we try to pass a higher protocol version.

PR #1067, opened 2020-02-07
PYTHON-1205: Fix Undefined column name native_transport_address" error when connecting to C* 4.0

aboudreault:


build.yaml
```diff
@@ -46,8 +46,8 @@ schedules:
       EXCLUDE_LONG=1
     matrix:
       exclude:
-        - python: [2.7, 3.4, 3.7, 3.8]
-        - cassandra: ['2.0', '2.1', '2.2', '3.0', 'test-dse', dse-4.8', 'dse-5.0']
+        - python: [2.7, 3.4, 3.7, 3.6, 3.8]
+        - cassandra: ['2.0', '2.1', '2.2', '3.0', '3.11', 'test-dse', 'dse-4.8', 'dse-5.0', 'dse-6.0', 'dse-6.7', 'dse-6.8']
```
aboudreault: TODO: remove those changes and add 4.0 at appropriate places

tests/integration/__init__.py
```diff
@@ -172,7 +172,7 @@ def _get_dse_version_from_cass(cass_version):
         cassandra_version = Version(mcv_string)
 
     CASSANDRA_VERSION = Version(mcv_string) if mcv_string else cassandra_version
-    CCM_VERSION = cassandra_version if mcv_string else CASSANDRA_VERSION
+    CCM_VERSION = mcv_string or cv_string
```
aboudreault: The changes here are to avoid using the packaging.Version with the CCM_VERSION. In this case, `4.0-alpha2` was converted to `4.0a2`

build.yaml
```diff
@@ -181,7 +182,7 @@ build:
       pip install --upgrade pip
       pip install -U setuptools
 
-      pip install git+ssh://git@github.com/riptano/ccm-private.git
+      pip install $HOME/ccm
```
TheRealFalcon: Is ccm already cloned here? What's this do?

tests/integration/__init__.py
TheRealFalcon: Can we use the `mcv_string if mcv_string else cv_string` syntax?

build.yaml
aboudreault: Yes. I wasn't aware but ccm-private is cloned when building driver images.

tests/integration/__init__.py
aboudreault: yeah, I'm fine if we prefer to use this more explicit syntax. will fix

PR #1068, opened 2020-03-12
Python 1207 transient parsing

TheRealFalcon:
Converted replication factors from int's to strings to support the parsing of transient replications.

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -450,18 +450,37 @@ class _UnknownStrategy(ReplicationStrategy):
         return {}
 
 
+def parse_replication_factor(input_rf):
+    """
+    Given the inputted replication factor, returns a tuple containing number of total replicas
+    and number of transient replicas
+    """
+    transient_replicas = None
+    try:
+        total_replicas = int(input_rf)
+    except ValueError:
+        try:
+            rf = input_rf.split('/')
+            total_replicas, transient_replicas = int(rf[0]), int(rf[1])
+        except Exception:
+            raise ValueError("Unable to determine replication factor from: {}".format(input_rf))
+    return total_replicas, transient_replicas
+
+
 class SimpleStrategy(ReplicationStrategy):
 
     replication_factor = None
     """
     The replication factor for this keyspace.
     """
+    transient_replicas = None
+    """
+    The number of transient replicas for this keyspace.
+    """
 
     def __init__(self, options_map):
-        try:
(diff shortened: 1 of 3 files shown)


cassandra/metadata.py
```diff
@@ -458,16 +475,14 @@ class SimpleStrategy(ReplicationStrategy):
     """
 
     def __init__(self, options_map):
-        try:
-            self.replication_factor = int(options_map['replication_factor'])
-        except Exception:
-            raise ValueError("SimpleStrategy requires an integer 'replication_factor' option")
+        self.replication_factor = options_map['replication_factor']
```
aboudreault: self.replication_factor is an integer since the beginning... it might be better to not change this. Could you add a `_raw_replication_factor` attribute and preserve `self.replication_factor` for the normal/total replica int value?

TheRealFalcon: That seems a little confusing to me. Are you worried about breaking existing behavior? The only place SimpleStrategy can get instantiated (other than tests) is here:
https://github.com/datastax/python-driver/blob/master/cassandra/metadata.py#L691

and nothing in the metadata parsing directly references instance attributes of the RF classes. If `replication_factor` was being accessed directly, you'd need to do a check to make sure you have a SimpleStrategy because other RFs don't have it.

Calling `export_as_string()` already prints it as a string, so no worries there.

aboudreault: `replication_strategy` is exposed in the KeyspaceMetadata, as you pointed. That means it is breaking change if the user is using the metadata to do anything. I understand it is not a critical breaking change but it is still one. No matter if they check what type is the strategy class... they expect an integer if they read that attribute.

Another option: Since the transient replication is not enabled by default, I'd would prefer we keep things as is since it's 99.99% of our current users and fallback to a `string` attribute if TR is detected when parsing. We can document this properly in the attribute docs. wdyt?

TheRealFalcon: I see. Since backwards compatibility is an issue, I think I prefer your first suggestion then. I thought having `replication_factor` not represent the "true" replication factor would be confusing, but if we have to do this to keep backwards compatibility, that's fine.

aboudreault: that was a valid point too. Should we expose a `replication_transient_factor` then?

TheRealFalcon: I added a separate `transient_replicas` field. Let me know if that looks good to you.

PR #1070, opened 2020-03-12
PYTHON-1226 Remove all user-facing references to Apollo or Constellation

TheRealFalcon:
Just to verify, we DON'T plan on changing this particular flag or references to it?
https://github.com/datastax/python-driver/blob/master/cassandra/datastax/cloud/__init__.py#L44

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -146,7 +146,7 @@ October 28, 2019
 
 Features
 --------
-* DataStax Apollo (Astra) Support (PYTHON-1074)
+* DataStax Astra Support (PYTHON-1074)
 * Use 4.0 schema parser in 4 alpha and snapshot builds (PYTHON-1158)
 
 Bug Fixes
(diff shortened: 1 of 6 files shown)


adutra:
> Just to verify, we DON'T plan on changing this particular flag or references to it?
> https://github.com/datastax/python-driver/blob/master/cassandra/datastax/cloud/__init__.py#L44

The string contents won't change. However you can change the variable name to something else that is more neutral (the Java driver for example names it `DATASTAX_CLOUD_PRODUCT_TYPE`).

CHANGELOG.rst
```diff
@@ -146,7 +146,7 @@ October 28, 2019
 
 Features
 --------
-* DataStax Apollo Support (PYTHON-1074)
+* DataStax Apollo (Astra) Support (PYTHON-1074)
```
adutra: I think you can simply remove "Apollo" even if the ticket was created with that word.

PR #1071, opened 2020-03-23
PYTHON-700: Support port discovery for C* 4.0

aboudreault:
This PR adds these features:
* The ability to specify a contact point as tuple: ('127.0.0.1', 9042) (so specifying the port by CP without having to create an explicit DefaultEndPoint. The main reason for this is that the OSS ccm (and Apple) is already using this. It was easy to support it anyway.
* system.peers_v2 support. I Implemented it the same way than the java-driver... we try to use it when we connect and initialize the control connection... otherwise we fallback to peers v1.
* With peersv2, ports are discovered per-host.

Next major: I want to be able to run the entire test suite using a single interface. e.g when testing C* 4.0+  (ticket to create). Too many changes were required at the moment since we use strings everywhere for node ip etc. and that would have been another mess to merge back in the 4.x branch.

### DIFF
--- a/cassandra/pool.py
+++ b/cassandra/pool.py
@@ -55,21 +55,60 @@ class Host(object):
 
     broadcast_address = None
     """
-    broadcast address configured for the node, *if available* ('peer' in system.peers table).
-    This is not present in the ``system.local`` table for older versions of Cassandra. It is also not queried if
-    :attr:`~.Cluster.token_metadata_enabled` is ``False``.
+    broadcast address configured for the node, *if available*:
+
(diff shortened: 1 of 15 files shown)


cassandra/pool.py
```diff
@@ -143,6 +182,11 @@ def address(self):
         # backward compatibility
         return self.endpoint.address
 
+    @property
+    def full_address(self):
```
aboudreault: not used anymore will be removed

aboudreault:
@TheRealFalcon this is ready for review... later I will rebase it on your 4.0 tests branch.

PR #1073, opened 2020-03-26
PYTHON-1207: transient replication support (phase 2)

aboudreault:

### DIFF
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -386,6 +386,7 @@ class ReplicationStrategyTypeType(type):
         return cls
 
 
+
 @six.add_metaclass(ReplicationStrategyTypeType)
 class _ReplicationStrategy(object):
     options_map = None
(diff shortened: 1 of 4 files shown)


cassandra/metadata.py
```diff
@@ -450,37 +450,82 @@ def make_token_replica_map(self, token_to_host_owner, ring):
...
+
+        return self.all_replicas == other.all_replicas and self.full_replicas == other.full_replicas
 
 
 class SimpleStrategy(ReplicationStrategy):
 
-    replication_factor = None
-    """
-    The replication factor for this keyspace.
+    replication_factor_info = None
```
aboudreault: `info` anything better to suggest?

aboudreault:
@TheRealFalcon Could you give this a first look before I continue? based on the new information received yesterday.. I think it's better to move the RF to a better data structure.

cassandra/metadata.py
TheRealFalcon: `info` is ok. I also thought maybe `full_replication_factor` or `complete_replication_factor`.

Do we need to expose this publicly? It'll be shown in the `export_as_string` methods. Since there's not really a consistent interface between the different ReplicationFactor implementations, I think it makes more sense to keep it private. I'd lean towards removing the class variable and just have a private `_replication_factor` instance variable.

cassandra/metadata.py
```diff
@@ -502,36 +547,40 @@ def export_for_schema(self):
...
         if not isinstance(other, SimpleStrategy):
             return False
 
-        return str(self._raw_replication_factor) == str(other._raw_replication_factor)
+        return str(self.replication_factor_info) == str(other.replication_factor_info)
 
 
 class NetworkTopologyStrategy(ReplicationStrategy):
 
+    dc_replication_factors_info = None
```
TheRealFalcon: Same comment here. Do we need to expose this?

TheRealFalcon:
Left comments but +1 overall. The class is a nice abstraction.

cassandra/metadata.py
aboudreault: The problem I see if we don't expose it is that for a transient replication, we are exposing an *incomplete* RF (since replication_factor only expose full_replicas). A user cannot even know if the rf is transient or not. export_as_string is not really convenient except for generating cql.

PR #1085, opened 2020-06-01
PYTHON-1248: tcp flow control for libevreactor

aboudreault:
* Fix the inflight max request id issue found while working in this ticket. You can ignore that stuff while reviewing.
* Add a better tcp flow control. 
  1. When a tcp send buffer is full, we mark the connection as non writable, so the query planner skip that host automatically on the next query.
  2. When the send buffer is writable again, we flush all the current write queue *then* mark the connection as writable again.

A good part here is the contribution of James Falcon. Unfortunately I had to rebase to clean and separate commits properly.


cassandra/connection.py
```diff
@@ -926,6 +927,8 @@ def send_msg(self, msg, request_id, cb, encoder=ProtocolHandler.encode_message,
             raise ConnectionShutdown("Connection to %s is defunct" % self.endpoint)
         elif self.is_closed:
             raise ConnectionShutdown("Connection to %s is closed" % self.endpoint)
+        elif not self._socket_writable:
+            raise ConnectionBusy("Connection %s is overloaded" % self.endpoint)
```
joao-r-reis, line 931: This is the only instance where the `_socket_writable` flag is checked so IIUC the python driver already handles TCP flow control correctly but this PR adds an optimization so that the connection throws this exception when the buffer is full? 

I wonder if it would be better to handle this on the LBP level (the new LBP checks the in flight counter for example). Otherwise you could probably just use the [THROW_ON_OVERLOAD startup option](https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L278) to handle this for you no?

aboudreault: Thanks Joao for your comments. Yes, that's correct. Before, we were just exiting the write handler so the write queue was still growing and there was more OTO. Now, we skip that host immediately and pass to the next one until the connection gets in better shape.

This will definitively be better with the new default LBP, but this one is not yet in the driver. our impl also checks the in_flight count. And at the moment, our LBPs dont have the session context to do that kind of logic. This is also something we improved in the PoC LBP ticket.

THROW_ON_OVERLOAD: What would be the benefit to use that? IIUC, the bad node will still be selected and the query will be sent, then we need to handle the exception and the retry_next_host logic will happen. Since python is already extremely slow... I tend to prefer to detect earlier and avoid any network cycle to that bad node. Am I missing something? However, I think it's a good idea to use it for the other reactors..

joao-r-reis: Yeah you're right the driver would still have to send the request which would be slower... 👍

cassandra/io/libevreactor.py
```diff
@@ -310,22 +310,28 @@ def handle_write(self, watcher, revents, errno=None):
...
+                    if err.args[0] in NONBLOCKING:
+                        self._socket_writable = False
                     with self._deque_lock:
                         self.deque.appendleft(next_msg)
                 else:
                     self.defunct(err)
                 return
             else:
                 if sent < len(next_msg):
+                    self._socket_writable = False
```
aboudreault, line 331: Do you think it's too aggressive? should I do this only if `send == 0`?

tests/integration/simulacron/test_connection.py
```diff
@@ -24,22 +24,22 @@
 from cassandra import OperationTimedOut
```
aboudreault, line 24: this file is all related to the max inflight stuff.

tests/unit/test_host_connection_pool.py
```diff
@@ -22,11 +22,14 @@
```
aboudreault, line 22: this file is all related to the max inflight stuff.

aboudreault:
@joao-r-reis If you can give this another review, that would be appreciated :) And no worries if you prefer I assign someone else for this python review.

cassandra/io/libevreactor.py
joao-r-reis: I'm not too familiar with the low level socket API but IIRC if `sent == 0` that would mean that the remote endpoint closed the connection right? So you would want to close the socket instead of setting the flag to false in this scenario.

I'm not sure if setting the flag immediately here is better than letting the loop continue and potentially receive a `EAGAIN` / `EWOULDBLOCK` in the next `send()`... I think that this is fine and maybe better than letting the loop continue.

When is the `writable` flag being set back to `true` again? Should the driver check if it is `false` after a `socket.send()` succeeds with all data being written and set it to `true`?

joao-r-reis:
> 
> 
> @joao-r-reis If you can give this another review, that would be appreciated :) And no worries if you prefer I assign someone else for this python review.

My review is very superficial, I still haven't had the time to setup the dev environment and take a deeper look at the python driver's code (and I'm not very familiar with the language 😅 ). Having said that, I can review it again no problem.

cassandra/io/libevreactor.py
joao-r-reis: oh I just read this on the PR description:

> When the send buffer is writable again, we flush all the current write queue then mark the connection as writable again.

This answers my question related to setting the `writable` flag 👍

tests/integration/simulacron/test_backpressure.py
```diff
@@ -0,0 +1,179 @@
...
+        buffer = '1' * 50000
+        for _ in range(100000):
+            future = session.execute_async(query, [buffer], **execute_kwargs)
+            futures.append(future)
+
+            total_blocked = 0
+            for pool in session.get_pools():
+                if not pool._connection._socket_writable:
+                    total_blocked += 1
+            if total_blocked == expected_blocked:
```
joao-r-reis, line 48: is there any scenario where `total_blocked` can be greater than `expected_blocked`? Might be safer to just change it to `total_blocked >= expected_blocked`

cassandra/io/libevreactor.py
aboudreault: Good point... I think it's not really adding any value. I will remove the sent check and wait egain/ewouldblock.

tests/integration/simulacron/test_backpressure.py
aboudreault: no chance since there is only 3 connections. But better to fix this yep.

aboudreault:
Created PYTHON-1249 for the memory issue.

TheRealFalcon:
When I was testing this, I remember sometimes seeing the socket wouldn't always raise `NONBLOCKING`, but rather just return 0 bytes sent, letting requests still pile up (https://github.com/datastax/python-driver/blob/bbdefc7e15f599c9179ea26f4fc31ce117e76b09/cassandra/io/libevreactor.py#L330). I remember certain event loops would always act a particular way, but I thought libev sometimes did both.

aboudreault:
Thanks @TheRealFalcon for raising this. How did you experienced this, just with the tcp backpressure tests? I would definitively assume that if 0 byte were sent for a request, all subsequent ones raise NONBLOCKING..

TheRealFalcon:
It's been a while, so the details are fuzzy. Just looked back at my previous PR and I didn't have it in there either, so it must have been a different event loop.

https://bugs.python.org/issue20951 is probably relevant.

aboudreault:
Thanks James! I will check this..

joao-r-reis:
> 
> 
> When I was testing this, I remember sometimes seeing the socket wouldn't always raise `NONBLOCKING`, but rather just return 0 bytes sent, letting requests still pile up

Oh then this changes things, specifically this discussion: https://github.com/datastax/python-driver/pull/1085#discussion_r433993217

aboudreault:
Yes, from what I've seen after a look at the python patch... it seems to be a case when using SSL. So I will add it back to be sure.

PR #1091, opened 2020-08-31
PYTHON-1258: Implement protocol v5 checksumming

aboudreault:
TODO:
- [ ]  Could be a good addition to add some unit io tests in tests/unit/io to test the connection._process_segment_buffer function


cassandra/segment.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @property
+    def header_length(self):
+        return self.COMPRESSED_HEADER_LENGTH if self.compression \
+            else self.UNCOMPRESSED_HEADER_LENGTH
+
+    @property
+    def header_length_with_crc(self):
+        return (self.COMPRESSED_HEADER_LENGTH if self.compression
+                else self.UNCOMPRESSED_HEADER_LENGTH) + CRC24_LENGTH
```
aboudreault, line 117: use self.header_length

PR #1092, opened 2020-09-11
PYTHON-1260: Ensure the driver can connect when invalid peer hosts are in system.peers

aboudreault:


aboudreault:
@dpitera timwaizenegger If you could take a look at this change and test, that would be great.

timwaizenegger:
> @dpitera @timwaizenegger If you could take a look at this change and test, that would be great.

Hi @aboudreault sorry for the delay; I just tested latest master with this change and it works! Tested against one of our DBs that currently has null-entries in the peers tables.
The old client failed to connect, this one works!

Thanks for fixing this :) 

Btw, I just opened a support ticket to have this change included in cqlsh: https://support.datastax.com/hc/en-us/requests/66256

aboudreault:
@timwaizenegger Awesome. Thank you!

PR #1094, opened 2020-09-24
Don't create Cluster with hosts if having cloud config

rzvoncek:
We set out to build a Flask app that talks to Astra.
The CQL plugin we used turned out to be crashing if we didn't give it hosts to connect to. [here](https://github.com/thegeorgeous/flask-cqlalchemy/blob/master/flask_cqlalchemy/__init__.py#L48).
But we also had to use the `cloud` key to pass in Astra credentials (via `setup_kwargs` nearby the link above).
This exploded on us in cqlengine.connection#setup() where the hosts passed into Cluster() made that constructor blow up because:
`ValueError: contact_points, endpoint_factory, ssl_context, and ssl_options cannot be specified with a cloud configuration`

We're not sure this is the right place to patch, but it helped us at least. We figured coming out with a patch directly, instead of an issue, might save some time.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Features
 Bug Fixes
 ---------
 * Asyncore race condition cause logging exception on shutdown (PYTHON-1266)
+* Update list of reserved keywords (PYTHON-1269)
 
 Others
 ------
--- a/cassandra/metadata.py
+++ b/cassandra/metadata.py
@@ -52,18 +52,18 @@ cql_keywords = set((
     'counter', 'create', 'custom', 'date', 'decimal', 'default', 'delete', 'desc', 'describe', 'deterministic', 'distinct', 'double', 'drop',
     'entries', 'execute', 'exists', 'filtering', 'finalfunc', 'float', 'from', 'frozen', 'full', 'function',
     'functions', 'grant', 'if', 'in', 'index', 'inet', 'infinity', 'initcond', 'input', 'insert', 'int', 'into', 'is', 'json',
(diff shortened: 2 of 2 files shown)


aboudreault:
Hey @rzvoncek ! Quick question, can you try to create a normal Cluster+Session with the astra bundle, then just set the session in cqlengine? https://docs.datastax.com/en/developer/python-driver/3.24/api/cassandra/cqlengine/connection/#cassandra.cqlengine.connection.set_session

rzvoncek:
Yeah, I can do this:
```
cluster = Cluster(cloud=cloud_config, auth_provider=auth_provider)
session = cluster.connect()
connection.register_connection('test', session=session)
row = connection.get_connection('test').session.execute("select release_version from system.local").one()
print(row)
```
To get:
```
{'release_version': '4.0.0.684'}
```
But I couldn't do:
```
cluster = Cluster(contact_points='<redacted>.apps.astra.datastax.com', 
                  cloud=cloud_config, 
                  auth_provider=auth_provider)
```

The problem though is that in the context of the Flask and the `falask-cqlalchemy` module we can't change how the connection is set up. 

I understand this is more problem for cqlalchemy and we should perhaps go open PRs to them. I just figured we could try to make the python-driver more resilient too. Perhaps we should abandon this PR then.

aboudreault:
No, I think your PR makes sense. We should probably allow users to initialize cqlengine the same way they were doing on a non-astra cluster. I will create a ticket and see if we can include this quick change in the next release (no date yet).

aboudreault:
Just to mention: for flask-alchemy, you will probably also need to set a fake cassandra host: https://github.com/thegeorgeous/flask-cqlalchemy/blob/master/flask_cqlalchemy/__init__.py#L55

PR #1108, opened 2021-06-18
Add changes to build arm64 wheels

tbbharaj:
This PR addresses the issue to build arm64 wheels. To address this issue, I updated `travis.yml` to include `arm64`. 

Also there were two additional errors (https://travis-ci.com/github/tbbharaj/python-driver/builds/229569631) I encountered when I triggered master build w/o any changes.
1. On python: pypy2.7-6.0:
   Ref: https://travis-ci.com/github/tbbharaj/python-driver/jobs/515470491

```
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-9Ah7r6/lz4/
371The command "if [[ $TRAVIS_PYTHON_VERSION != pypy3.5 ]]; then pip install lz4; fi" failed and exited with 1 during .
```

2. On Python: 3.7
    Ref: https://travis-ci.com/github/tbbharaj/python-driver/jobs/515470490

```
ERROR: InvocationError for command /home/travis/build/tbbharaj/python-driver/.tox/gevent_loop/bin/nosetests --verbosity=2 --no-path-adjustment /home/travis/build/tbbharaj/python-driver/tests/unit/io/test_geventreactor.py (exited with code -11 (SIGSEGV)) (exited with code -11)
939___________________________________ summary ____________________________________
940ERROR:   gevent_loop: commands failed
941The command "tox -e gevent_loop" exited with 1.
```

To fix above two errors, I made following changes:
1. Update setup tools in `travis.yml` to fix build error in `python setup.py egg_info`, 
2. Update greenlet version in `tox.ini` to fix build error in `tox -e gevent_loop`

This morning I checked the builds for Arm are stuck in waiting stage https://www.travis-ci.com/github/tbbharaj/python-driver/builds/229919085
whereas yesterday it was working fine https://travis-ci.com/github/tbbharaj/python-driver/builds/229656900

If there are any questions/comments - I'd be happy fix things as needed.

Thank you
Tanveen

PR #1109, opened 2021-06-18
add test change

tbbharaj:


tbbharaj:
> @tbbharaj I think this one can also be closed for reasons similar to the ones discussed in #1108. Any objection?

Done!

PR #1112, opened 2021-10-14
Refactor deprecated unittest aliases for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268 .

### DIFF
--- a/tests/integration/cqlengine/management/test_compaction_settings.py
+++ b/tests/integration/cqlengine/management/test_compaction_settings.py
@@ -83,7 +83,7 @@ class AlterTableTest(BaseCassEngTestCase):
 
         table_meta = _get_table_metadata(tmp)
 
-        self.assertRegexpMatches(table_meta.export_as_string(), '.*SizeTieredCompactionStrategy.*')
+        six.assertRegex(self, table_meta.export_as_string(), '.*SizeTieredCompactionStrategy.*')
 
     def test_alter_options(self):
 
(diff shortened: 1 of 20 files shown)


absurdfarce:
@tirkarthi I'm inclined to wait on this until 3.11 has had some more time to develop.  What do you think?

tirkarthi:
I am fine with the wait since Python 3.11 is in alpha phase release and final release will be around October 2022.

https://www.python.org/dev/peps/pep-0664/

stefanor:
That cleanup was later reverted, but this patch is still a good idea.

tirkarthi:
The deprecations will be removed in Python 3.12. I will try to fix the merge conflict.

https://docs.python.org/dev/whatsnew/3.12.html#removed

absurdfarce:
Fair point @stefanor, we don't have to wait for a new release to remove these features in order to address them.

According to the underlying Python issue (https://bugs.python.org/issue45162 by way of https://github.com/python/cpython/pull/28268):

```
Most features were deprecated in 3.2, "fail*" methods in 3.1, assertNotRegexpMatches in 3.5. They were kept 
mostly for compatibility with 2.7 (although some of them were new in Python 3 and not compatible with 2.7).
```

The most recent version of the Python driver supports Python 3.5 through 3.8 so clearly we're dealing with deprecated methods for the supported versions.  With that in mind I'm okay with getting this in sooner than originally discussed.

There are still a couple things we need to sort out before we do so however.  We're working on some internal changes which have borked our CI infrastructure temporarily; I'd like to get those sorted out before I commit this.  I also need to make sure we're okay on the CLA front before moving forward.  @tirkarthi can you confirm that you've signed this agreement?

biegelmeyer:
Are there any news about this PR?

absurdfarce:
There were two outstanding issues on this PR when last I looked into it @biegelmeyer.  They were:

* Issue with the internal CI infrastructure used by this project
* Confirmation of a signed CLA from the author (@tirkarthi)

I believe the CI infrastructure is fixed, or at least fixed enough that we can commit this change without too much concern.  We need a CLA signature before we can do anything, though, and unfortunately that's out of my hands.  @tirkarthi have you signed the CLA or will you be able to do so soon?

tirkarthi:
I have just signed the CLA now. Please check.

biegelmeyer:
> I have just signed the CLA now. Please check.

nice

absurdfarce:
Excellent, thank you @tirkarthi !

And with that done I believe we're ready to merge this change.  Thanks everyone!

PR #1114, opened 2021-10-26
Stop reusing stream ids of requests that have timed out due to client-side timeout

haaawk:
Context:
The driver allows to send multiple requests concurrently using a single connection.
Requests are matched with the corresponding responses using stream ids.
Each request gets a unique stream id and the corresponding response contains the same stream id.
The pool of stream ids is limited and in CQL protocol v3 and above has 32768 values.
That means stream ids have to be reused.
Usually when the request gets a response, the stream id used by this request is returned to the pool.

Problem:
The driver has an ability to time out a request on a client side.
The problem is that when such situation happens, the stream id of the timed out request is returned to the pool of available stream ids and another request can pick up the same stream id.

Problematic situation looks as follows:
1. request R1 is sent to the server with stream id S
2. request R1 times out due to a client-side time out
3. stream id S is returned to the pool of available streams
4. request R2 is sent to the server with stream id S
5. Server sends back a response to the request R1. Such response contains stream id S and is now interpreted as a response to the request R2.

This leads to data being returned for one query being interpreted as a response to another query.
For example if user sessions are stored in the DB under a user_id primary key then one user could get session of another.

This PR fixes the problem by making sure that stream ids of requests that time out due to client-side timeouts are not returned to the pool of available stream ids. This can lead to a situation when the pool shrinks significantly which decreases the number of concurrent requests being able to be sent over this connection. To avoid such a bad state the PR also monitors the amount of orphaned stream ids and when it reaches 75% of the initial pool size, the connection is recreated and all stream ids become available again.


cassandra/connection.py
```diff
@@ -764,6 +781,8 @@ def __init__(self, host='127.0.0.1', port=9042, authenticator=None,
         self._io_buffer = _ConnectionIOBuffer(self)
         self._continuous_paging_sessions = {}
         self._socket_writable = True
+        self.orphaned_request_ids = set()
```
absurdfarce, line 784: The drawback for using a straight set here is that req_ids are Python ints which are (usually) 4 bytes in size.  So you're looking at (in the worst case) 32768 msgs * 4 bytes/msg ~= 128k _per connection_.  In practice I don't think this will be a problem but there's no need to rely on that.  If we use a bitset here we'd normalize that to a fixed size of ~4k/conn.  Might be worth thinking about.

absurdfarce (commented):
Thanks for your work on this so far @haaawk!  I agree this is an issue and I like the general direction you've taken to solve it.

I'm publishing a review noting a few things that stood out to me on an initial walk through.  I'm going to try to take another pass at this early next week but I wanted to get some feedback to you as quickly as possible

Thanks again for all your work here!

cassandra/connection.py
absurdfarce: Nevermind, I forgot you had a threshold for connection termination that effectively functions as a cap on this.  As long as that size doesn't become absurdly large there's no issue here.

cassandra/pool.py
```diff
@@ -399,42 +403,59 @@ def __init__(self, host, host_distance, session):
...
             raise NoConnectionsAvailable()
+        return conn
+
+    def borrow_connection(self, timeout):
+        conn = self._get_connection()
+        if conn.orphaned_threshold_reached:
+            with self._lock:
+                if not self._is_replacing:
+                    self._is_replacing = True
+                    self._session.submit(self._replace, conn)
```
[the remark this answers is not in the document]
haaawk: I believe that this code is ok.

The intention is to keep using the old connection until a new connection is ready to operate. Otherwise we would block the client until the new connection is read and we probably don't want to do this. `self._get_connection()` will start to return a new connection after `_replace` assigns `self._connection`.  It's ok to use the old connection for a bit longer as the new connection should be established relatively quickly. Does that make sense, @absurdfarce?

Until the replacing is called `self._get_connection()` will return the same connection that's already assigned to `conn`.

cassandra/connection.py
```diff
@@ -1188,11 +1207,22 @@ def process_msg(self, header, body):
...
                 result_metadata = None
             else:
+                need_notify_of_release = False
+                with self.lock:
+                    if stream_id in self.orphaned_request_ids:
+                        self.in_flight -= 1
+                        self.orphaned_request_ids.remove(stream_id)
+                        need_notify_of_release = True
+                if need_notify_of_release and self._owning_pool:
+                    self._owning_pool.on_orphaned_stream_released()
```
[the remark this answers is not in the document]
haaawk: You're right. Fixed.

haaawk:
Thanks for the review @absurdfarce.
I pushed a new version in which connection takes callback instead of owning pool.
I answered the other comment. I hope it makes things more clear.

haaawk:
@absurdfarce ping

absurdfarce:
@haaawk ACKed.  Thanks for the ping, and (again) my apologies for the delay.  I think we're in a good state here but I wanted to give this one more pass before we called it good.  I'll try to get that wrapped up today.

cassandra/connection.py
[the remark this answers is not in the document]
absurdfarce: Nice, I like the fix you put in here @haaawk !

Side note: I looked into the AppVeyor test failures that sprang up after this change went in.  Turns out this change introduces the possibility of a random test failure in test_cluster.DontPrepareOnIgnoredHostsTest.  There's a race there that can trigger an assertion failure if one specific side wins the race.  I'm planning on writing all that up and filing a separate ticket for it, however; based on my testing the actual failure case is pretty unlikely.  So I don't see any reason why that issue should hold up getting this change in.

cassandra/connection.py
```diff
@@ -1188,11 +1207,22 @@ def process_msg(self, header, body):
...
+                if need_notify_of_release and self._on_orphaned_stream_released:
+                    self._on_orphaned_stream_released()
+
                 try:
                     callback, decoder, result_metadata = self._requests.pop(stream_id)
                 # This can only happen if the stream_id was
                 # removed due to an OperationTimedOut
                 except KeyError:
+                    with self.lock:
+                        self.request_ids.append(stream_id)
```
absurdfarce, line 1225: Just to be very clear... this change isn't explicitly connected to the problem you're trying to address in the PR, correct @haaawk ?  The intent here is that if we get a stream ID that doesn't match up to a request we know about we should be safe to re-use that stream ID regardless of other conditions... do I have that right?

cassandra/pool.py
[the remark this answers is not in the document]
absurdfarce: Agreed, after reading this more carefully it became apparent I wasn't evaluating the full context here @haaawk .  If anything I'd say the requirement is stronger than your argument above.  "conn" is scoped to borrow_connection() and we're using it to validate a good working connection before returning from that fn in the "while True" loop below.  There's no obvious point to setting conn at this point since the loop handles all of that.

I'm good with what you have here.

absurdfarce:
Hey @haaawk , after looking this over again I'm fine with the current state of things.  I'll leave this open through today to give you a chance to reply to any of my more recent comments if you feel there's something to say but otherwise this will likely get merged tonight or tomorrow morning.

cassandra/connection.py
haaawk: In https://github.com/datastax/python-driver/pull/1114/commits/12c9c306d39989142b21d06a315f6657d42d1eef:
1. If a request R has timed out due to a client side timeout, we stopped returning its stream ID to the pool of available stream IDs
2. If the same request R receives a response after R has timed out on a client side, the stream ID of R will be put back to the pool of available stream IDs

This line performs (2).

>if we get a stream ID that doesn't match up to a request we know about we should be safe to re-use that stream ID regardless of other conditions... do I have that right?

Yes. The assumption is that if we got a stream ID we don't know then it must be a stream ID of a request that has already timed out due to a client side timeout.

haaawk:
> Hey @haaawk , after looking this over again I'm fine with the current state of things. I'll leave this open through today to give you a chance to reply to any of my more recent comments if you feel there's something to say but otherwise this will likely get merged tonight or tomorrow morning.

Thanks @absurdfarce. I commented on https://github.com/datastax/python-driver/pull/1114#discussion_r754405011

cassandra/connection.py
absurdfarce: Okay, thanks for the verification @haaawk !

absurdfarce:
FTR The ticket describing the test failures described above is https://datastax-oss.atlassian.net/browse/PYTHON-1287

PR #1115, opened 2021-11-16
fix(test-requirements.txt): Install unitest2 package only for Python2

Orenef11:


absurdfarce:
@Orenef11 I like the idea contained in your PR, but I'm wondering if we can't go even further.

The docs for unittest2 mention that it's known to work for Python 2.6, 2.7 and 3.2 through 3.4.  The only version from this list which isn't EOL'd is 2.7, and unittest2 was created to backport changes to unittest in 2.7 to earlier versions... which in turn suggests that the base unittest in Python 2.7 might be sufficient.  And if that's the case I think we can do away with unittest2 entirely at this point; the versions it backports functionality too just aren't supported by python-driver anymore.

So I created an experiment at https://github.com/datastax/python-driver/pull/1117.  This removes unittest2 entirely and replaces all it's uses with the base unittest lib.  I'm considering going with this idea in favor of this PR in order to remove a now-unnecessary dependency.

What do you think?

Orenef11:
> This removes unittest2 entirely and replaces all it's uses with the base unittest lib. I'm considering going with this idea in favor of this PR in order to remove a now-unnecessary dependency.

Great idea!
I think this is the best and most correct way because Python2.7 is EOL.

PR #1117, opened 2021-11-22
PYTHON-1289: remove unittest2

absurdfarce:
Motivated by recent PR from @Orenef11 (https://github.com/datastax/python-driver/pull/1115).  After digging into it a bit it became clearer that unittest2 likely isn't adding any value at this point.  It was created as a backport for new functionality in the unittest package of 2.7 to earlier versions in the 2.x line.  None of those earlier versions are supported by the current driver, however, and all 3.x Python versions supported by unittest2 are also now EOL.  Seems ideal to remove the dependency all together as long as we can get to a clean test run without it.


absurdfarce:
test_authentication_misconfiguration:MisconfiguredAuthenticationTests passes locally for Python 2.7 so this looks to be another red herring.

Orenef11:
> test_authentication_misconfiguration:MisconfiguredAuthenticationTests passes locally for Python 2.7 so this looks to be another red herring.

LGTM.
Trying to re-start the GitHub actions again.

Jimena904:
> Motivated by recent PR from @Orenef11 (https://github.com/datastax/python-driver/pull/1115).  After digging into it a bit it became clearer that unittest2 likely isn't adding any value at this point.  It was created as a backport for new functionality in the unittest package of 2.7 to earlier versions in the 2.x line.  None of those earlier versions are supported by the current driver, however, and all 3.x Python versions supported by unittest2 are also now EOL.  Seems ideal to remove the dependency all together as long as we can get to a clean test run without it.

PR #1118, opened 2021-11-23
fir(test/__init__.py): Ignoring DeprecationWarning from "cassandra.io.asyncioreactor"

Orenef11:
* Replace all places that use old async node (`@asyncio.coroutine` and `yield from`) and use a new Python syntax (Only for Python v3.5+)
* Save the old code to allow users with Python v2+ to run our tests.
* requirements.txt:
  * Install `futures` package only for Python v2
  * Use the latest `sex` package version

PR #1119, opened 2022-01-25
PYTHON-1290 Convert asyncio reactor away from @asyncio.coroutine

absurdfarce:
Short-term goal is to address the underlying issue at play in https://github.com/datastax/python-driver/pull/1118.  Longer-term this has the nice benefit of making us more compatible with how asyncio appears to be going generally.

### DIFF
--- a/cassandra/io/asyncioreactor.py
+++ b/cassandra/io/asyncioreactor.py
@@ -10,6 +10,13 @@ from threading import Lock, Thread, get_ident
 
 log = logging.getLogger(__name__)
 
+
+# This module uses ``yield from`` and ``@asyncio.coroutine`` over ``await`` and
+# ``async def`` for pre-Python-3.5 compatibility, so keep in mind that the
+# managed coroutines are generator-based, not native coroutines. See PEP 492:
+# https://www.python.org/dev/peps/pep-0492/#coroutine-objects
+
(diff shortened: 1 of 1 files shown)


Orenef11:
Good work.
1. This change is not supported by Python 3.4 and below.
Do we want to abandon the support in 2.7?
In my opinion, I am in favor of not supporting Python2.7 as it is no longer supported but it should be noted in README.
2. The PR contains 4 commits, can you create one commit?

absurdfarce:
Thanks @Orenef11!  With regards to your questions:

1. We should be okay with Python 2.7.  asyncio wasn't introduced until Python ~3.4 so that reactor isn't even an option for 2.7

2. When we get to a PR we're satisfied with I'll do a squash merge; it'll compress all the changes we have here down into a single commit.

Orenef11:
> Thanks @Orenef11! With regards to your questions:
> 
> 1. We should be okay with Python 2.7.  asyncio wasn't introduced until Python ~3.4 so that reactor isn't even an option for 2.7
> 2. When we get to a PR we're satisfied with I'll do a squash merge; it'll compress all the changes we have here down into a single commit.

LGTM
Good work!

Orenef11 (approved):

PR #1120, opened 2022-01-26
Fixing "DeprecationWarning: invalid escape sequence \" warning

Orenef11:
Fixes https://datastax-oss.atlassian.net/browse/PYTHON-1291

### DIFF
--- a/tests/integration/standard/test_cluster.py
+++ b/tests/integration/standard/test_cluster.py
@@ -147,7 +147,8 @@ class ClusterTests(unittest.TestCase):
         get_node(1).pause()
         cluster = TestCluster(contact_points=['127.0.0.1'], connect_timeout=1)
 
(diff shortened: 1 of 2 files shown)


absurdfarce:
Hey @Orenef11, thanks for putting this together!

Based on the docs I'm reading (https://docs.python.org/3/library/unittest.html#deprecated-aliases as an example) it looks like the names assertRaisesRegexp and assertRegexpMatches have been completely deprecated.  Given that I'm wondering if we wouldn't be better off to switch to the supported methods going forward rather than trying to make deprecated methods work correctly.  What do you think?

bschoening:
@absurdfarce @Orenef11  I don't see assertRaisesRegexp in the current trunk.  Can this PR be closed as obsolete?

PR #1121, opened 2022-01-26
fix(tests/integration/__init__.py): Rename "TestCluster" to "IntegrationTestCluster"

Orenef11:
```
PytestCollectionWarning: cannot collect test class 'TestCluster' because it has a __new__ constructor (from: tests/integration/standard/test_client_warnings.py)
  class TestCluster(object):
```

### DIFF
--- a/tests/integration/__init__.py
+++ b/tests/integration/__init__.py
@@ -72,7 +72,7 @@ def get_server_versions():
     if cass_version is not None:
         return (cass_version, cql_version)
 
-    c = TestCluster()
+    c = IntegrationTestCluster()
     s = c.connect()
     row = s.execute('SELECT cql_version, release_version FROM system.local')[0]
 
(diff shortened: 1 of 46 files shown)


Orenef11:
@absurdfarce 
The goal is replace `unittest` to `Pytest`:
1. Can I change all `unittest` marks and use `Pytest`?
2. If not, can I change only the CI to run it with `Pytest`?

absurdfarce:
Hey @Orenef11, apologies for the long delay... I'm trying to juggle way too many plates around here.

We're working on fully converting to pytest as part of the upcoming 3.30.0 release.  The current work around that effort can be found in #1215; there's more detail in [PYTHON-1297](https://datastax-oss.atlassian.net/browse/PYTHON-1297) as well.  @bschoening pointed out that the issue discussed here should be addressed by [this change](https://github.com/datastax/python-driver/pull/1215/files#diff-6ec5224a3ecfa5b56e1dce12620fffa868d0b50190add628982d3596085b7e9d) in that PR.  Once we get that merged feel free to double-check that assumption and let us know if we're wrong.

absurdfarce:
Since this PR was originally opened we've completed the conversion to pytest and are using it regularly.  I also believe the change referenced in my comment above does fix the issue being addressed here.  As a result I'm going to close this out, but as always if anybody believes there's something we missed or an additional part of this PR we should (re)consider please open a new PR and let's talk about it.

PR #1123, opened 2022-03-14
This version of the driver does not support protocol version 21

allen-munsch:
several service providers are deprecating and EOL cassandra 3.0 for cassandra 4

any updates here would be great

i was also unable to submit a ticket to jira ... so figured i'd just leave this here

### DIFF
--- a/cassandra/__init__.py
+++ b/cassandra/__init__.py
@@ -168,6 +168,12 @@ class ProtocolVersion(object):
     """
     v6, in beta from 4.0-beta5
     """
+    
+    V21 = 21
+    """"
+    Support Cassandra 4.0 ... many services are EOL the default of 3.0 in 3-4 months
+    This version of the driver does not support protocol version 21 ????
+    """"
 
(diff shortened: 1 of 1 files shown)


absurdfarce:
Thanks for the PR @allen-munsch !

Could you elaborate a bit on what your concern is here?  This section of the code is specifying names for the various native protocols supported by this driver.  You can find specifications for each version of the protocol in the "doc" directory of the Cassandra repo (see https://github.com/apache/cassandra/tree/trunk/doc as an example).

This protocol is versioned independently of Cassandra itself; Cassandra 4.0 provides the first full implementation of a non-beta version of protocol version 5, for instance.  With regard to your PR specifically there is no version 21 of the native protocol that I am aware of, either in Cassandra itself or DSE.

allen-munsch:
@absurdfarce ah, I was trying to figure out how to do an upgrade. I think I misunderstood what those versions meant.

Thanks for the pointing out that information, and highlighting the versioning.

PR #1125, opened 2022-04-07
add python 3.9 and 3.10 support

themegaphoenix:


absurdfarce:
Thanks for the PR @themegaphoenix!

We have a not-exactly-short-term but also not-exactly-long-term goal to bring platform support for the Python driver (and perhaps other drivers as well) up-to-date with current platforms.  At the moment this effort is most clearly expressed in [PYTHON-1283](https://datastax-oss.atlassian.net/browse/PYTHON-1283) but there is other work going on as well.  Validating a release requires extensive testing (and likely a bug fix or two) but please keep an eye on that ticket for updates.

In the meantime we are working on making wheels for 3.9 and 3.10 available.  This is not to say that we officially support either of these runtimes yet but we are aware that the driver is used on these Python versions so we're trying to make that process easier.  You can track that work at [PYTHON-1295](https://datastax-oss.atlassian.net/browse/PYTHON-1295).

Thanks again!

fabiopedrosa:
@absurdfarce There's a (at least) one bug in Python 3.10 where in `AsyncioConnection` the Queue is getting a outdated argument called `loop`:

```
self._write_queue = asyncio.Queue(loop=self._loop)
```

This raises an exception:

```
TypeError: As of 3.10, the *loop* parameter was removed from Queue() since it is no longer necessary
```

absurdfarce:
Note for future readers: @fabiopedrosa filed [PYTHON-1313](https://datastax-oss.atlassian.net/browse/PYTHON-1313) to describe the issue above.  This issue has subsequently been converted into a subtask of [PYTHON-1283](https://datastax-oss.atlassian.net/browse/PYTHON-1283) which is the overarching ticket for 3.9 and 3.10 compatibility work.

fabiopedrosa:
@absurdfarce there's no news here?

PR #1127, opened 2022-06-08
PYTHON-539 Timeouts in ResponseFuture should notify the conviction policy

absurdfarce:


absurdfarce:
AppVeyor failure is from a few random test failures, which (unfortunately) happens from time to time with those builds.  I believe the Jenkins failures stem from an issue with the Jenkins server in question; this needs some more work.

absurdfarce:
After looking into this today I've determined that the Jenkins failures do represent legit breakage with this PR.  Can't tell yet if it's an unintended consequence of these changes or if the tests are making too many assumptions... still working on sorting that out.

absurdfarce:
Test failures are also occurring on master; we've definitely got an issue to sort out there.  I've filed [PYTHON-1304](https://datastax-oss.atlassian.net/browse/PYTHON-1304) to address those issues (and possibly to get us back to a clean build on Jenkins generally).

PR #1130, opened 2022-11-09
PYTHON-1267 Adding a more useful error message for NaN values into decimal

synedra:


synedra:
Updating the logic.

cassandra/cqltypes.py
```diff
@@ -444,6 +444,8 @@ def deserialize(byts, protocol_version):
 
     @staticmethod
     def serialize(dec, protocol_version):
+        if (Decimal(dec).is_nan()):
+            raise Exception("NaN is not a valid Decimal value")
```
absurdfarce, line 448: So I think this is happening just a bit too early.

Note that immediately after this operation you can see serialize(...) trying to call as_tuple() on the input "dec" value.  If it's a Decimal type this'll work great, but if it's not you'll chuck the AttributeError.  In that case the code will try to convert "dec" to a Decimal (via Decimal(dec)) and try as_tuple() on the result.

I'd suggest that your code should follow a similar pattern.  Call dec.is_nan() and throw as necessary, then call is_tuple() and continue normally.  If any of that throws an AttributeError do something similar in the exception handler; convert "dec" to a Decimal and then do the is_nan() + is_tuple() ops.

absurdfarce:
It would be nice to see a test case here, ideally something that was failing before but passes with this change.

absurdfarce:
Also, while we're here it's worth noting that the [Python docs on Decimal](https://docs.python.org/3/library/decimal.html) indicate that other "special" values are supported for Decimal instances, specifically "Infinity" and "-Infinity".  There's also a distinction between 0 and -0.  Ideally any test case implemented for this work would validate the behaviour for each of these special values; if we can get "NaN' here it's not impossible that we could get any of the others as well.

PR #1131, opened 2022-11-18
Python 3.11 support [bpo-46730]

stefanor:
The AttributeError message for a missing property setter changed in [bpo-46730](https://bugs.python.org/issue46730).

PR #1133, opened 2022-12-28
Handle "log gone" (PYTHON-1266) at the end of _run_loop

tlasica:
If log is somehow gone and file exception due to the race mention in PYTHON-1266 it will also inevitably fail for the same reason after the loop so we need to catch the exception there as well.

I have hit it in some Cassandra dtest:
```
Exception in thread asyncore_cassandra_driver_event_loop (most likely raised during interpreter shutdown):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File "/dse/dse-db/bin/../lib/cassandra-driver-internal-only-3.25.0.zip/cassandra-driver-3.25.0/cassandra/io/asyncorereactor.py", line 262, in _run_loop
<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'debug'
```

[PYTHON-1266]: https://datastax-oss.atlassian.net/browse/PYTHON-1266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cassandra/io/asyncorereactor.py
+++ b/cassandra/io/asyncorereactor.py
@@ -259,7 +259,13 @@ class AsyncoreLoop(object):
                     break
             self._started = False
 
-        log.debug("Asyncore event loop ended")
+        try:
+            log.debug("Asyncore event loop ended")
+        except Exception:
+            # TODO: Remove when Python 2 support is removed
+            # PYTHON-1266. If our logger has disappeared, there's nothing we
+            # can do, so just log nothing.
+            pass            
 
     def add_timer(self, timer):
         self._timers.add_timer(timer)


cassandra/io/asyncorereactor.py
```diff
@@ -259,7 +259,13 @@ def _run_loop(self):
...
             self._started = False
 
-        log.debug("Asyncore event loop ended")
+        try:
+            log.debug("Asyncore event loop ended")
+        except Exception:
+            # TODO: Remove when Python 2 support is removed
+            # PYTHON-1266. If our logger has disappeared, there's nothing we
+            # can do, so just log nothing.
+            pass
```
absurdfarce, line 268: Given that this verbiage is already used for the existing logging try/catch in this method we're starting to get into some ugly duplication here.  Prolly worth pulling this out into a function, something like:

```python
def maybe_log_debug(*args, **kwargs)
    try:
        log.debug(*args, **kwargs)
    except Exception:
        # TODO: Remove when Python 2 support is removed
        # PYTHON-1266. If our logger has disappeared, there's nothing we
        # can do, so just log nothing.
        pass
```

and then call with appropriate params.

This is a relatively minor thing so I can prolly also just clean it up after this gets merged if you don't wanna muck with it @tlasica.  Either way is good with me!

tlasica: I was thinking about it, but I decided not to do it:
1. there are 2 occurences, my rule of thumb is 3
2. all are local in the same function

pepgid:
Perhaps it is prudent to gateway language version sensitive areas of the code using the factory pattern. This would save adding many if and try statements downstream.

absurdfarce:
@pepgid Was your comment perhaps meant to apply to https://github.com/datastax/python-driver/pull/1131 instead of this PR?

cassandra/io/asyncorereactor.py
absurdfarce: No worries, I can take care of it.  Thanks for the fix @tlasica!

PR #1135, opened 2023-02-21
PYTHON-1267: Made error alerting for NaN values when they try to be serialized.

elliottrusso:


cassandra/cqltypes.py
```diff
@@ -450,6 +450,8 @@ def serialize(dec, protocol_version):
             try:
                 sign, digits, exponent = Decimal(dec).as_tuple()
             except Exception:
+                if Decimal(dec).is_nan():
```
absurdfarce, line 453: We're instantiating another instance the same value we have in the try block above here.  Would be nice if there was some way to use a single value in both places, although we'll want to make sure any errors that might be raised by that constructor are handled.

diff.txt
```diff
@@ -0,0 +1,13 @@
...
++++ b/cassandra/cqltypes.py
+@@ -450,6 +450,8 @@ class DecimalType(_CassandraType):
+             try:
+                 sign, digits, exponent = Decimal(dec).as_tuple()
+             except Exception:
++                if Decimal(dec).is_nan():
++                    raise ValueError("NaN values are not accepted values because they can't be serialized.")
+                 raise TypeError("Invalid type for Decimal value: %r", dec)
+         unscaled = int(''.join([str(digit) for digit in digits]))
+         if sign:
```
absurdfarce, line 13: I'm assuming this was added by mistake :)

cassandra/cqltypes.py
```diff
@@ -450,6 +450,8 @@ def serialize(dec, protocol_version):
...
+                    raise ValueError("NaN values are not accepted values because they can't be serialized.")
```
absurdfarce, line 454: A test for this change would be a good idea.  tests:unit:test_types.py would be a good place to look for inspiration.

PR #1136, opened 2023-02-24
Make pipeline step fail when tests fail

joao-r-reis:
Currently the overall build fails when tests fail but the individual step is not marked as failed so it makes it harder to figure out which step actually caused the build failure. This PR fixes that.


joao-r-reis:
The diff looks weird but the only change that I added was the `try catch` and `failure` variable in some `sh` steps.

PR #1138, opened 2023-03-07
Remove different build matrix selection for develop branches

absurdfarce:
Goal here is to use the smoke tests for as many build cases as we reasonably can

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -87,13 +87,9 @@ def getBuildContext() {
   def profile = "${params.PROFILE}"
   def EVENT_LOOP = "${params.EVENT_LOOP.toLowerCase()}"
   matrixType = "SMOKE"
-  developBranchPattern = ~"((dev|long)-)?python-.*"
 
-  if (developBranchPattern.matcher(env.BRANCH_NAME).matches()) {
-    matrixType = "DEVELOP"
-    if (env.BRANCH_NAME.contains("long")) {
-      profile = "FULL"
-    }
+  if (env.BRANCH_NAME.contains("long")) {
+    profile = "FULL"
   }
 
   // Check if parameters were set explicitly


absurdfarce:
Right now the Jenkinsfile for the Python driver uses a different build matrix for feature branches named to match a PYTHON JIRA ticket.  The build matrix in question doesn't really make sense for our plans going forward (it still includes Python 2.7 builds) and we do have the SMOKE build matrix which more naturally corresponds to the versions we'd like to support (and is already used as the default build matrix for most Jenkins builds).

Goal here is basically to remove the special-case handling for branches matching this naming pattern.  Going forward we may want to remove the DEVELOP matrix all together, but honestly I'd expect something like that to be handled in a larger-scale refactoring of the various matrices defined in the Jenkinsfile.

For now, I just want to remove this special behaviour... I _really_ want Jenkins feature branch builds to use the same matrix that will be used when we merge that feature branch into master.

absurdfarce:
Worth mentioning here: the build for "develop branches" actually fails now due to an issue with something on 2.7:

```

[2023-03-07T20:45:29.706Z] + pip install -r test-requirements.txt
[2023-03-07T20:45:30.008Z] DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
[2023-03-07T20:45:30.008Z] Ignoring twisted: markers 'python_version >= "3.5"' don't match your environment
[2023-03-07T20:45:30.008Z] Ignoring futurist: markers 'python_version >= "3.7"' don't match your environment
[2023-03-07T20:45:30.008Z] Ignoring asynctest: markers 'python_version >= "3.5"' don't match your environment
...
[2023-03-07T20:45:31.480Z] Collecting pure-sasl
[2023-03-07T20:45:31.480Z]   Downloading pure_sasl-0.6.2-py2-none-any.whl (11 kB)
[2023-03-07T20:45:31.480Z] Collecting twisted[tls]==19.2.1
[2023-03-07T20:45:31.480Z]   Downloading Twisted-19.2.1.tar.bz2 (3.1 MB)
[2023-03-07T20:45:33.601Z]     ERROR: Command errored out with exit status 1:
[2023-03-07T20:45:33.601Z]      command: /home/jenkins/.pyenv/versions/2.7.18/bin/python2.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-uGalC0/twisted/setup.py'"'"'; __file__='"'"'/tmp/pip-install-uGalC0/twisted/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-JUjBaC
[2023-03-07T20:45:33.601Z]          cwd: /tmp/pip-install-uGalC0/twisted/
[2023-03-07T20:45:33.601Z]     Complete output (25 lines):
[2023-03-07T20:45:33.601Z]     WARNING: The wheel package is not available.
[2023-03-07T20:45:33.601Z]     DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
[2023-03-07T20:45:33.601Z]     Traceback (most recent call last):
[2023-03-07T20:45:33.601Z]       File "<string>", line 1, in <module>
[2023-03-07T20:45:33.601Z]       File "/tmp/pip-install-uGalC0/twisted/setup.py", line 20, in <module>
[2023-03-07T20:45:33.601Z]         setuptools.setup(**_setup["getSetupArgs"]())
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/site-packages/setuptools/__init__.py", line 162, in setup
[2023-03-07T20:45:33.601Z]         return distutils.core.setup(**attrs)
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/distutils/core.py", line 111, in setup
[2023-03-07T20:45:33.601Z]         _setup_distribution = dist = klass(attrs)
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/site-packages/setuptools/dist.py", line 448, in __init__
[2023-03-07T20:45:33.601Z]         k: v for k, v in attrs.items()
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/distutils/dist.py", line 287, in __init__
[2023-03-07T20:45:33.601Z]         self.finalize_options()
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/site-packages/setuptools/dist.py", line 740, in finalize_options
[2023-03-07T20:45:33.601Z]         ep(self)
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/site-packages/setuptools/dist.py", line 747, in _finalize_setup_keywords
[2023-03-07T20:45:33.601Z]         ep.load()(self, ep.name, value)
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2443, in load
[2023-03-07T20:45:33.601Z]         return self.resolve()
[2023-03-07T20:45:33.601Z]       File "/home/jenkins/.pyenv/versions/2.7.18/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2449, in resolve
[2023-03-07T20:45:33.601Z]         module = __import__(self.module_name, fromlist=['__name__'], level=0)
[2023-03-07T20:45:33.601Z]       File "/tmp/pip-install-uGalC0/twisted/.eggs/incremental-22.10.0-py2.7.egg/incremental/__init__.py", line 14, in <module>
[2023-03-07T20:45:33.601Z]         from typing import TYPE_CHECKING, Any, TypeVar, Union, Optional, Dict
[2023-03-07T20:45:33.601Z]     ImportError: No module named typing
[2023-03-07T20:45:33.601Z]     ----------------------------------------
[2023-03-07T20:45:33.601Z] ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

Looks to be a transitive dependency issue with something not supporting 2.7.x anymore (typing is a Python3 module).  I've dug into this enough to identify the above but that's about it; given that 2.7.x support will be going away there's no obvious reason to investigate this any further.  And since that's the case there's no obvious reason to continue to execute anything on Jenkins against 2.7.x.

Jenkinsfile
```diff
@@ -87,13 +87,9 @@ def getBuildContext() {
...
   def EVENT_LOOP = "${params.EVENT_LOOP.toLowerCase()}"
   matrixType = "SMOKE"
-  developBranchPattern = ~"((dev|long)-)?python-.*"
 
-  if (developBranchPattern.matcher(env.BRANCH_NAME).matches()) {
-    matrixType = "DEVELOP"
-    if (env.BRANCH_NAME.contains("long")) {
-      profile = "FULL"
-    }
+  if (env.BRANCH_NAME.contains("long")) {
```
joao-r-reis, line 91: Should we remove this `if` as well?

absurdfarce: That's an interesting question.

There's going to be some changes to all these build matrices to support our new version support policy for the Python driver; from the new release on a given driver release will only support Python versions that aren't out-of-support at the time of release.  So 2.7, 3.5 and 3.6 will be removed from all five build matrices, for instance.  There's also a legit question for me about whether we really need five distinct build matrices... that seems like a lot.

But as I thought about it you're actually asking a different question; if we're already moving away from branch name correlation to a build matrix why not sever that tie completely?  I think this is prolly right; I don't really like the idea that a branch name magically correlates to a specific build matrix.  Seems better to maybe run other matrices on regular schedules and support ad-hoc runs if somebody really wants to run something now.

Whaddya think?

joao-r-reis: > if we're already moving away from branch name correlation to a build matrix why not sever that tie completely

Exactly, if a branch gets named "something-long" and the build starts failing we may take a lot of time to realize it's using a completely different build because of this `if`, I'd just remove it.

absurdfarce: Sold.  I was originally avoiding making this PR too big under the assumption that the overall build matrix would have to be revisited once we change supported versions of the Python runtime but as discussed above that concern isn't explicitly connected to what you're advocating here @joao-r-reis .

I'll make the change.

Jenkinsfile
```diff
@@ -33,30 +33,34 @@ slack = new Slack()
...
+    "SERVER": DEFAULT_DSE,
+    "RUNTIME": DEFAULT_RUNTIME,
+    "CYTHON": DEFAULT_CYTHON
   ],
   "SMOKE": [
-    "SERVER": ['3.11', '4.0', 'dse-6.8.30'],
-    "RUNTIME": ['3.7.7', '3.8.3'],
+    "SERVER": DEFAULT_CASSANDRA.takeRight(2) + DEFAULT_DSE.takeRight(1),
+    "RUNTIME": DEFAULT_RUNTIME.takeRight(2),
     "CYTHON": ["False"]
```
absurdfarce, line 64: Just reducing some redundancy while we're here + making things a bit more declarative

elliottrusso (approved):

Jenkinsfile
```diff
@@ -33,30 +33,34 @@ slack = new Slack()
 //
 // Smoke tests are CI-friendly test configuration.  Currently-supported Python version + modern C*/DSE instances.
 // We also avoid cython since it's tested as part of the nightlies.
+DEFAULT_CASSANDRA = ['2.1', '2.2', '3.0', '3.11', '4.0']
+DEFAULT_DSE = ['dse-5.0.15', 'dse-5.1.35', 'dse-6.0.18', 'dse-6.7.17', 'dse-6.8.30']
+DEFAULT_RUNTIME = ['2.7.18', '3.5.9', '3.6.10', '3.7.7', '3.8.3']
```
absurdfarce, line 38: It's a little strange to keep 2.7.x + 3.5.x + 3.6.x around here.  My rationale is that this PR is really about removing the special handling of "develop" branches + some normalization for how versions are specified.  Actual changes in versions we test against will be done in a PR explicitly focused on that role (and should be accompanied by corresponding changes to the runners).

absurdfarce:
@joao-r-reis Quick ping; are you good with this change as it stands?

Jenkinsfile
joao-r-reis: This is a great way to keep these versions in a single place, TIL!

PR #1139, opened 2023-03-08
PYTHON-1328: Add explicit wait to give cluster time to get initialized

absurdfarce:


tests/integration/standard/test_authentication.py
```diff
@@ -42,12 +43,19 @@ def setup_module():
         log.debug("Starting ccm test cluster with %s", config_options)
         start_cluster_wait_for_up(ccm_cluster)
 
+    # PYTHON-1328
+    #
+    # Give the cluster enough time to startup (and perform necessary initialization)
+    # before executing the test.
+    if CASSANDRA_VERSION > Version('4.0-a'):
```
absurdfarce, line 50: Version string here pulled from the [annotations](https://github.com/datastax/python-driver/blob/master/tests/integration/__init__.py#L333-L335) defined in `tests/integration/__init__.py`

tests/integration/standard/test_authentication.py
```diff
@@ -42,12 +43,19 @@ def setup_module():
...
+        time.sleep(10)
```
absurdfarce, line 51: This wait was happening within a test method but this relies on tests being executed in a specific order which is... not great.  Moving to module initialization logic should resolve that issue.

PR #1144, opened 2023-04-03
PYTHON-1297: Pytest compatibility to move from nosetest

elliottrusso:
Enabled unit tests to work with Pytest
Made initial changes to integration tests for Pytest usage

### DIFF
--- a/tests/integration/cqlengine/conftest.py
+++ b/tests/integration/cqlengine/conftest.py
@@ -1,0 +1,9 @@
+import pytest
+
+from tests.integration.cqlengine import setup_package, teardown_package, setup_connection
+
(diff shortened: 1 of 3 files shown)


elliottrusso:
I'm uncertain if the import of pytest is required but it was when I was was running the tests locally. Also this does work with nosetest too surprisingly.

absurdfarce:
@elliottrusso I know there is some follow-up work being done on this topic in other branches.  Should we just close this particular PR out knowing that bigger and better things are following soon?

absurdfarce:
Closing this out in favor of other work being done elsewhere.  Thanks for all the hard work on this effort @elliottrusso!

PR #1151, opened 2023-04-26
Don't fail when inserting UDTs with prepared queries with some missing fields #224

Mokto:
Let's say you have an object:
```
class Address(object):

    def __init__(self, street, zipcode, **kwargs):
        self.street = street
        self.zipcode = zipcode

cluster.register_user_type('mykeyspace', 'address', Address)
```

And let's say the type actually contains another field, let's call i raw_address

Then inserting data through a prepared statement will actually fail : the driver will complain raw_address is missing.

This change addresses that, as any field should be optional.

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -1026,7 +1026,9 @@ class UserType(TupleType):
             try:
                 item = val[i]
             except TypeError:
-                item = getattr(val, fieldname)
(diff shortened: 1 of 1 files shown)


Mokto:
I just did ;)

Mokto:
Hi,
Any update on this ?

Thanks.

absurdfarce (approved):

PR #1152, opened 2023-04-27
PYTHON-1341 Documentation for CLE functionality 

absurdfarce:


johnsmartco (commented):
Hi Bret. Looks really good. 

Typo on L1242 of policies.py: interally should be internally.

Other minor things in policies.py from prior edits:

falsey (ok? 3 occurrences)
L525: falsy 
truthy (ok?  3 occurrences)
L627, 637, 643, 666:  inbetween
L726: acheive s/b achieve

johnsmartco (commented):
In docs/column_encryption.rst, a very minor edit:

L15, "it's work"  s/b  "its work"

( it's  is appropriate only as a contraction of two words  "it is" , as done correctly on L21.  There's no possessive form with an apostrophe of the word "its".)

absurdfarce:
Thanks for the review @johnsmartco!  I addressed most of your comments in my most recent commit.  Apologies, my system clock was off so it looks like that commit actually came in before your review comments were posted... apparently time zones aren't the only time-related thing I have a problem with.

That "it's" thing has been the bane of my existence for years.  Somewhere along the line the incorrect usage you cite got stuck in my noggin' and I haven't been able to completely remove it even yet.  Someday...

Anyways, on the question of "truthy" and "falsy".  Python doesn't explicitly use these terms but they are used in many similar programming languages and Python _does_ have a similar concept (even if it doesn't call it "truthy" or "falsy").  I'd argue the terms are broadly used enough that their use here is probably okay.  On the question of "falsy" vs. "falsey"... I don't have any particular authority to cite but in general I usually see it as "falsy"... and that seems to be the more common usage in the resources I can find.  I guess I don't have a strong opinion either way, but I agree that whichever way we go with we should at least be consistent in our own documentation.

johnsmartco:
Thanks Bret for the explanations.  "falsy" has a nice ring to it.  Thanks for the edits.  Looking good for Monday 5/1.

johnsmartco (approved):
LGTM

johnsmartco:
Hi  Bret @absurdfarce.  My follow up on Monday 01-May will be to:

1. Check with you on code / docs build timing.
2. If we're a "go," I'll edit [ docs.yaml ](https://github.com/datastax/python-driver/blob/master/docs.yaml#L24) on a branch, adding an entry for 3.27 and the parent commit ID.
3. I'll then follow the revised and corrected docs build instructions that Jamie edited so well on [this wiki page](https://datastax.jira.com/wiki/spaces/DOCS/pages/431226938/Building+and+publishing+driver+docs+and+driver+API+docs).

PR #1159, opened 2023-05-18
Fix for rendering of code blocks in CLE documentation

absurdfarce:
Looks like I wasn't following rst as closely as I should've been.

Relevant notes on code block formatting can be found [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives).

### DIFF
--- a/docs/column_encryption.rst
+++ b/docs/column_encryption.rst
@@ -21,6 +21,7 @@ and adding information about columns to be encrypted to it.  This policy is then
 when it's created.
 
 .. code-block:: python
+
     import os
 
     from cassandra.policies import ColDesc, AES256ColumnEncryptionPolicy, AES256_KEY_SIZE_BYTES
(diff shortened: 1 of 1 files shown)


docs/column_encryption.rst
```diff
@@ -57,6 +58,7 @@ supplied parameters.  For example, we can create a prepared statement to insert
 by executing the following code after creating a :class:`~.Cluster` in the manner described above:
 
 .. code-block:: python
+
```
jgillenwater, line 61:
```suggestion
by executing the following code after creating a :class:`~.Cluster` in the manner described above::

```

docs/column_encryption.rst
```diff
@@ -70,6 +72,7 @@ standard serialization methods employed by the driver.  The result is then encry
 of the policy.  Using this approach the example above could be implemented along the lines of the following:
 
 .. code-block:: python
+
```
jgillenwater, line 75:
```suggestion
of the policy.  Using this approach the example above could be implemented along the lines of the following::

```

jgillenwater (changes requested):
I left two comments that would make this structured like some of the codeblocks in other topics. Can we try this instead?

absurdfarce:
Met with @jgillenwater to discuss and we settled on the Sphinx-specific code-block usage for now.  Also filed PYTHON-1349 to come up with a more robust long-term answer for this across the entire set of Python driver docs.

PR #1162, opened 2023-05-23
PYTHON-1343 Use Cython for smoke builds

absurdfarce:


absurdfarce:
As expected we're now seeing the test failures documented in PYTHON-1347 on Jenkins runs with this change.  The benefit of testing against the default case (i.e. Cython) as our default for smoke builds is too large, though.

PR #1163, opened 2023-05-30
Jenkinsfile cleanup

absurdfarce:
Updating Jenkinsfile to reflect recent supported Python versions.  Also trying to get nightlies going again against master.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -386,8 +386,9 @@ def describeBuild(buildContext) {
   }
 }
 
-def scheduleTriggerJobName() {
-  "drivers/python/oss/master/disabled"
+// branch pattern for cron
+def branchPatternCron() {
+  ~"(master)"
 }
 
 pipeline {
@@ -617,7 +618,7 @@ pipeline {
   }
 
   triggers {
-    parameterizedCron((scheduleTriggerJobName() == env.JOB_NAME) ? """
+    parameterizedCron(branchPatternCron().matcher(env.BRANCH_NAME).matches() ? """
       # Every weeknight (Monday - Friday) around 4:00 AM
       # These schedules will run with and without Cython enabled for Python 3.7.7 and 3.8.3
       H 4 * * 1-5 %CI_SCHEDULE=WEEKNIGHTS;EVENT_LOOP=LIBEV;CI_SCHEDULE_PYTHON_VERSION=3.7.7 3.8.3;CI_SCHEDULE_SERVER_VERSION=2.2 3.11 dse-5.1.35 dse-6.0.18 dse-6.7.17
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,9 +13,10 @@ Test Profiles:
 Matrix Types:
 
   Full: All server versions, python runtimes tested with and without Cython.
-  Develop: Smaller matrix for dev purpose.
   Cassandra: All cassandra server versions.
   Dse: All dse server versions.
+  Smoke: CI-friendly configurations. Currently-supported Python version + modern Cassandra/DSE instances.
+         We also avoid cython since it's tested as part of the nightlies
 
 Parameters: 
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -629,10 +618,10 @@ pipeline {
   }
 
   triggers {
-    parameterizedCron((scheduleTriggerJobName() == env.JOB_NAME) ? """
+    parameterizedCron(branchPatternCron().matcher(env.BRANCH_NAME).matches() ? """
```
absurdfarce, line 621: Trying to make this more consistent with the other drivers.  This usage also seems to make more sense logically.

Jenkinsfile
```diff
@@ -29,25 +30,16 @@ import com.datastax.jenkins.drivers.python.Slack
...
 slack = new Slack()
 
-// Define our predefined matrices
-//
-// Smoke tests are CI-friendly test configuration.  Currently-supported Python version + modern C*/DSE instances.
-// We also avoid cython since it's tested as part of the nightlies.
 DEFAULT_CASSANDRA = ['2.1', '2.2', '3.0', '3.11', '4.0']
 DEFAULT_DSE = ['dse-5.0.15', 'dse-5.1.35', 'dse-6.0.18', 'dse-6.7.17', 'dse-6.8.30']
-DEFAULT_RUNTIME = ['2.7.18', '3.5.9', '3.6.10', '3.7.7', '3.8.3']
+DEFAULT_RUNTIME = ['3.7.7', '3.8.3']
```
absurdfarce, line 35: Didn't seem to make much sense to include Python versions we no longer explicitly support

Jenkinsfile
```diff
@@ -29,25 +30,16 @@ import com.datastax.jenkins.drivers.python.Slack
...
   "FULL": [
     "SERVER": DEFAULT_CASSANDRA + DEFAULT_DSE,
     "RUNTIME": DEFAULT_RUNTIME,
     "CYTHON": DEFAULT_CYTHON
   ],
-  "DEVELOP": [
-    "SERVER": ['2.1', '3.11', 'dse-6.8.30'],
-    "RUNTIME": ['2.7.18', '3.6.10'],
-    "CYTHON": DEFAULT_CYTHON
-  ],
```
absurdfarce, line 50: This test matrix doesn't really make sense, especially after we remove these Python versions from the supported list above

Jenkinsfile
```diff
@@ -629,10 +618,10 @@ pipeline {
   }
 
   triggers {
-    parameterizedCron((scheduleTriggerJobName() == env.JOB_NAME) ? """
+    parameterizedCron(branchPatternCron().matcher(env.BRANCH_NAME).matches() ? """
       # Every weeknight (Monday - Friday) around 4:00 AM
-      # These schedules will run with and without Cython enabled for Python v2.7.18 and v3.5.9
-      H 4 * * 1-5 %CI_SCHEDULE=WEEKNIGHTS;EVENT_LOOP=LIBEV;CI_SCHEDULE_PYTHON_VERSION=2.7.18 3.5.9;CI_SCHEDULE_SERVER_VERSION=2.2 3.11 dse-5.1.35 dse-6.0.18 dse-6.7.17
+      # These schedules will run with and without Cython enabled for Python 3.7.7 and 3.8.3
+      H 4 * * 1-5 %CI_SCHEDULE=WEEKNIGHTS;EVENT_LOOP=LIBEV;CI_SCHEDULE_PYTHON_VERSION=3.7.7 3.8.3;CI_SCHEDULE_SERVER_VERSION=2.2 3.11 dse-5.1.35 dse-6.0.18 dse-6.7.17
```
absurdfarce, line 624: Part of my goal here is to get nightlies working again.  PYTHON-1351 will need some environment customizations to make sure we can test the cryptography and non-cryptography cases.  It probably makes sense to include one as part of the default build and the other in the nightlies.

absurdfarce: This probably won't be as bad as it looked initially.  PYTHON-1351 is following the example of similar functionality in the Python driver (for a Gremlin dependency for DS Graph) and importing functionality specific to cryptography only if the lib is present.  Since cryptography will be present in the Jenkins environment [1]  those tests will be run.

[1] cryptography has now been moved to test-datastax-requirements.txt, the same spot where the Gremlin requirement lives.  Requirements in that file are imported when the Jenkins environment is setup.

Jenkinsfile
```diff
@@ -394,8 +386,9 @@ def describeBuild(buildContext) {
   }
 }
 
-def scheduleTriggerJobName() {
-  "drivers/python/oss/master/disabled"
+// branch pattern for cron
+def branchPatternCron() {
+  ~"(master)"
```
joao-r-reis, line 391: you could change this to `jenkinsfile_cleanup` to test things before merging this PR if you want to

PR #1164, opened 2023-05-30
PYTHON-1351 Convert cryptography to an optional dependency

absurdfarce:


absurdfarce:
I've (manually) confirmed that after this PR the tests in this package fail due to an inability to find imports in cassandra.column_encryption.policies if cryptography is NOT present.  This result is entirely expected since the absence of cryptography results in cassandra.column_encryption._policies never being imported.  Tests pass cleanly if cryptography is present.

This isn't an issue for the Jenkins builds since cryptography is included in the default environment there due to the inclusion of test-datastax-requirements.txt in that environment.  It would've been included anyway as a transitive dependency of Twisted but that's a different story.

absurdfarce:
@jgillenwater I included you here for review of the documentation changes associated with this change.  I was debating adding more detail to the column_encryption.rst page about how the cryptography module was no longer required... but the more I thought about it the more it seemed like it duplicated what's added in installation.rst... and that definitely seems like the right place for it.

If you need to re-assign this please feel free to do so.

JeremiahDJordan (approved):
LGTM

docs/installation.rst
```diff
@@ -67,6 +67,24 @@ support this::
...
+versions treat this module as an optional requirement.
+
+If you wish to make use of the CLE feature you will need to install the cryptography
+module.  You can do so by installing the driver with the `cle` requirements::
+
+    pip install cassandra-driver[cle]
+
+You can also install the package directly via pip::
+
+    pip install cryptography
```
jgillenwater, line 87:
```suggestion
*Optional:* Column-Level Encryption (CLE) Support
--------------------------------------------------
The driver has built-in support for client-side encryption and
decryption of data. Please see :doc:`column_encryption` for more detail
about this feature, including how to configure and use it.  This
feature relies on the `cryptography <https://cryptography.io/en/latest/>`_ module
for its implementation. The initial release of CLE (Python driver 3.27.0) explicitly 
lists cryptography as a required dependency; later
versions allow this module as an optional requirement.

If you want to use the CLE feature, install the cryptography
module.

You can installing the driver with the `cle` requirements::

    pip install cassandra-driver[cle]

Alternatively, you can also install the package directly via `pip`::

    pip install cryptography
```

jgillenwater (approved):
Minor docs edits.

docs/installation.rst
```diff
@@ -67,6 +67,29 @@ support this::
...
+The driver has built-in support for client-side encryption and
+decryption of data. Please see :doc:`column_encryption` for more detail
+about this feature, including how to configure and use it.  This
+feature relies on the `cryptography <https://cryptography.io/en/latest/>`_ module
+for its implementation. The initial release of the CLE feature in Python driver 3.27.0
+explicitly listed cryptography as a required dependency; as a result the module will be
+automatically downloaded and installed when that version of the driver is installed.
+Later versions of the driver do not specify this module as a requirement, so if you
+want to use the CLE feature with a version of the driver after 3.27.0 you will need
+to take additional steps to install the cryptography module.
```
absurdfarce, line 81: I could see this change being a source of confusion so I wanted to very explicitly spell out the differences between the versions.  It's a bit more verbose but I'd argue this is a case where we want to provide a bit more information to make the constraints clear to our users.

docs/installation.rst
```diff
@@ -67,6 +67,29 @@ support this::
...
+You can install this module along with the driver by specifying the `cle` extra::
+
+    pip install cassandra-driver[cle]
+
+Alternatively, you can also install the module directly via `pip`::
+
+    pip install cryptography
+
+Any version of cryptography >= 35.0 will work for the CLE feature.  You can find additional
+details at `PYTHON-1351 <https://datastax-oss.atlassian.net/browse/PYTHON-1351>`_
```
absurdfarce, line 92: I meant to include the link to the relevant JIRA ticket in my initial version and just forgot it.  We don't have any other links to PYTHON tickets in this doc but it seemed like a good idea for this case which (as discussed above) is a bit more complicated than a normal doc change.

absurdfarce:
@jgillenwater My doc changes are in.  Take a look when you have time and let me know what you think!

docs/installation.rst
jgillenwater, line 81:
```suggestion
*Optional:* Column-Level Encryption (CLE) Support
--------------------------------------------------
The driver has built-in support for client-side encryption and
decryption of data. For more, see :doc:`column_encryption`.  

CLE depends on the Python `cryptography <https://cryptography.io/en/latest/>`_ module. 
When installing Python driver 3.27.0. the `cryptography` module is 
also downloaded and installed.
If you are using Python driver 3.28.0 or later and want to use CLE, you must
install the `cryptography <https://cryptography.io/en/latest/>`_ module.
```

jgillenwater (changes requested):

jgillenwater (changes requested):
Docs edits

absurdfarce:
Latest doc changes reflect the outcome of a conversation between @jgillenwater and myself.  Calling it good with what we have now.

bschoening:
Note: appveyor reports a warning due to missing init file:

package init file 'cassandra\column_encryption\__init__.py' not found (or not a regular file)

PR #1165, opened 2023-06-01
PYTHON-1356 Create session-specific protocol handlers to contain session-specific CLE policies

absurdfarce:


tests/integration/standard/test_policies.py
```diff
@@ -161,3 +161,37 @@ def test_end_to_end_simple(self):
         (encrypted,unencrypted) = session.execute(prepared, [expected]).one()
         self.assertEquals(expected, encrypted)
         self.assertEquals(expected, unencrypted)
+
+    def test_end_to_end_python1356(self):
```
absurdfarce, line 165: Confirmed that this test fails with what's on master now but passes after this change

PR #1171, opened 2023-07-08
Python 1297 add support for pytest for unit tests

bschoening:
Updated test_host_connection_pool.py to use the magic attribute __test__ = False (abstract class) and __test__ = True on derived classes.  Nose uses a leading underscore in the class name for similar functionality.

NOTE: the appveyor tests fail on the use of Python 3.6 f-string syntax. PR 1168 resolves this by changing appveyor.yaml to switch from Python 3.5 to 3.7.


bschoening:
@absurdfarce any update on this? It's only a couple of lines of code.

absurdfarce:
Apologies @bschoening, I've been driving hard to try and get 3.29.0 out the door and this one slipped past me.

My original hope was to get [PYTHON-1297](https://datastax-oss.atlassian.net/browse/PYTHON-1297) (which replaces nose with pytest) into 3.29.0 but we had to abandon that due to a pressing need to get 3.29.0 out.  We have done [quite a bit of work internally](https://github.com/datastax/python-driver/compare/master...jenkins-pytest) on PYTHON-1297 but that work isn't complete yet; at the moment it's the integration tests that are giving us the most problem.  And I'm reluctant to take steps (such as documentation changes) which suggest a change from nose to pytest until we're able to reliably run the entire test suite without issue.

We're trying to figure out if there's an easy way we can collaborate with you on this work without opening our Jenkins instance up to the world.  I don't know if we'll be able to do so but I'll post updates here if anything comes of that.

bschoening:
@absurdfarce Ok, I wasn't aware there was a problem with converting Nose integration tests.

Pytest is largely compatible with Nose test, but contributors such as myself can't run pytest because of the classnames in test_host_connection_pool.py. I've also never gotten nose installed and working.

So, a middle ground might be to commit the changes in test_host_connection_pool.py.  That would let anyone with a PR run the pytests and still work with nose.

bschoening:
@absurdfarce updated for compatibility with both nodestests and pytest.

absurdfarce:
Closing this in favor of #1215 which will represent our complete notion of pytest compat going forward.  @bschoening that branch _should_ have all of this work in it already so if you see something missing please let me know.  Obviously I very much welcome any additional feedback you may have as well.

PR #1173, opened 2023-07-25
cqltypes: Serialize None values in collections as NULLs

Lorak-mmk:
When using parepared statements, None values in collections were serialized as empty values (values with length == 0). This is unexpected and inconsistent - None values are serialized as NULLs (vlaues with length == -1) in other cases:
 - Statement arguments, both for simple and prepared statements
 - Collection elements in simple statement

This PR fixes this weird behavior - now None values should be serialized as NULLs in all cases. It also adds an integration test that checks new behavior.

Fixes https://datastax-oss.atlassian.net/jira/software/c/projects/PYTHON/issues/PYTHON-1355

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -835,9 +835,12 @@ class _SimpleParameterizedType(_ParameterizedType):
         buf.write(pack(len(items)))
         inner_proto = max(3, protocol_version)
         for item in items:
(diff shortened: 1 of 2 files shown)


Lorak-mmk:
I assume there is no interest in this PR, so closing.

Lorak-mmk:
Ok, thanks for the info. Reopening.

PR #1175, opened 2023-09-03
PYTHON-1358 Add a TimeUUID class

SiyaoIsHiding:
Closes [PYTHON-1358](https://datastax-oss.atlassian.net/browse/PYTHON-1358). Based on [Cassandra's implementation](https://github.com/apache/cassandra/blob/ae537abc6494564d7254a2126465522d86b44c1e/src/java/org/apache/cassandra/utils/TimeUUID.java) and the issue reporter's [code](https://gist.github.com/cvybhu/ed5b64d8b62eff51dc46258157a92e41).

I am considering where to put the class `TimeUUID` and the relevant tests. I currently put them in `util.py` and `test_types.py`, which is weird. Any suggestions?

Any requests for changes are welcomed.

### DIFF
--- a/cassandra/cqltypes.py
+++ b/cassandra/cqltypes.py
@@ -677,7 +677,7 @@ class TimeUUIDType(DateType):
 
     @staticmethod
     def deserialize(byts, protocol_version):
-        return UUID(bytes=byts)
+        return util.TimeUUID(bytes=byts)
 
     @staticmethod
     def serialize(timeuuid, protocol_version):
--- a/cassandra/util.py
+++ b/cassandra/util.py
@@ -116,6 +116,35 @@ def max_uuid_from_time(timestamp):
     """
     return uuid_from_time(timestamp, 0x7f7f7f7f7f7f, 0x3f7f)  # Max signed bytes (0x7f = 127)
 
+class TimeUUID(uuid.UUID):
+
+    @property
+    def lsb(self) -> int:
+        return int.from_bytes(self.bytes[8:16], 'big', signed=True)
+
+    def __eq__(self, other):
+        if isinstance(other, TimeUUID):
+            return self.time == other.time and self.lsb == other.lsb 
+        if isinstance(other, uuid.UUID):
+            return self.time == other.time and self.bytes[8:16] == other.bytes[8:16] 
+        return NotImplemented
+    
+    # def __lt__(self, other):
+    #     if isinstance(other, TimeUUID):
+    #         if self.time != other.time:
+    #             return self.time < other.time
+    #         else:
+    #             return self.lsb < other.lsb
+    #     return NotImplemented
+
+    def __gt__(self, other):
+        return not self <= other
+
+    def __le__(self, other):
+        return self < other or self == other
+
+    def __ge__(self, other):
+        return not self < other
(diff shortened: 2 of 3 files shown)


SiyaoIsHiding:
Signed the cla. And do I need to do anything about the Python 2 thing?

absurdfarce:
Hey @SiyaoIsHiding , nice to (virtually) see you again!  And hey, thanks for the PR!  Apologies for the delay in getting back to you; we've been trying to get multiple releases out at the same time and (as usual) we're running behind a bit.

Re: the Python 2 thing... you mean the AppVeyor build + tests, right?  Yeah, you can completely ignore those for now.  The AppVeyor builds for several of the drivers need to be updated so for now that's a no-op... focus on the Travis builds instead.

It looks like this PR is still in draft... are you at a spot where I should be reviewing anything here?  Or do you still have some work you want to do on it?

SiyaoIsHiding:
Hi @absurdfarce , excited to join your team!
No rush on this PR. I was just trying to familiarize myself with the codebase and I understand this issue is a minor priority. 

This is still a draft PR because I am considering where to put the class TimeUUID and the relevant tests. I currently put them in `util.py` and `test_types.py`, which is kinda weird. Do you have any suggestions?

Again, no rush at all. And looking forward to working with you in the team soon!

bschoening:
@SiyaoIsHiding there exists several PyPI libraries for type-1 uuid's, including time-uuid.

But Cassandra's util.py already has utility functions for comparing type-1 UUIDs. I'd question the need to introduce a new type here.  

>>> from uuid import UUID
>>> u1 = UUID('00000257-0efc-11ee-9547-00006490e9a6')
>>> u2 = UUID('fed35080-0efb-11ee-a1ca-00006490e9a4')
>>> u1 > u2
False
>>> from cassandra.util import datetime_from_uuid1
>>> dt1 = datetime_from_uuid1(u1)
>>> dt2 = datetime_from_uuid1(u2)
>>> dt1
datetime.datetime(2023, 6, 19, 23, 49, 58, 961212)
>>> dt2
datetime.datetime(2023, 6, 19, 23, 49, 56, 990579)
>>> dt1 > dt2
True

absurdfarce:
I'm inclined to agree with @bschoening on this one.  Changing the Python type we return when deserializing timeuuid types is a fairly significant change and definitely not something we'd want to undertake in a patch version update (or even a minor version update for that matter).

More generally it seems correct to me that this behaviour can be handled via a library function (or something similar) rather than something built into the type.  My rationale is that these are Python types representing the same data contained in Cassandra; at no point do we guarantee that they'll behave the same as types within Cassandra itself in all cases.  We do not guarantee that the string representations of all our types match up to those in Cassandra, for example.  Nor do we make any kind of attempt to ensure that iteration of (non-ordered) container types return data in the same order shown by the server.

For me, trying to match server behaviour here opens us up to a standard we have never claimed to follow (and which I'd argue we should not follow).  If there was no other obvious way to fix this issue I might feel differently but, again, utility functions already exist which enable the needed behaviour here.

SiyaoIsHiding:
Closing as it's not necassary and would introduce breaking chagnes

PR #1176, opened 2023-09-07
PYTHON-1362 updated travis config to use current python versions

weideng1:


absurdfarce:
A version of these fixes went in with an [earlier PR](https://github.com/datastax/python-driver/pull/1194) for 3.29.0.

PR #1156, opened 2023-05-09
remove unnecessary import __future__

bschoening:
unit it tests ran successfully (with pytest) except for those related to PYTHON-1300.
============================================================= short test summary info ==============================================================
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_borrow_and_return - TypeError: 'NoneType' object is not callable
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_failed_wait_for_connection - TypeError: 'NoneType' object is not callable
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_return_closed_connection - TypeError: 'NoneType' object is not callable
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_return_defunct_connection - TypeError: 'NoneType' object is not callable
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_return_defunct_connection_on_down_host - TypeError: 'NoneType' object is not callable
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_spawn_when_at_max - TypeError: 'NoneType' object is not callable
FAILED tests/unit/test_host_connection_pool.py::_PoolTests::test_successful_wait_for_connection - TypeError: 'NoneType' object is not callable
============================================= 7 failed, 559 passed, 68 skipped, 852 warnings in 20.11s =============================================

PR #1184, opened 2023-11-04
Remove backup(.bak) files and PY2 __future__ imports

bschoening:
These appear to be accidentally checked in and should be removed.


absurdfarce:
Hey, thanks for the PR @bschoening!  I'm quite good with the removal of the backup files but unless I'm missing something the __future__ imports removed here were the same ones we had a problem with in a [previous PR](https://github.com/datastax/python-driver/pull/1156).  There was some follow-up conversation around the (very unexpected) issues that were introduced by removing this code... but I can't find it right away.

absurdfarce:
[PYTHON-1348](https://datastax-oss.atlassian.net/browse/PYTHON-1348) is what I was thinking of

bschoening:
@absurdfarce ah, ok. It's ok to cherry pick just the .bak file deletes then, and we can deal with __future__ separately.

PR #1186, opened 2023-11-06
Remove outdated Python pre-3.7 references

bschoening:


README-dev.rst
```diff
@@ -176,7 +176,7 @@ Use tee to capture logs and see them on your terminal::
 
 Testing Multiple Python Versions
 --------------------------------
-If you want to test all of python 2.7, 3.5, 3.6, 3.7, and pypy, use tox (this is what
+If you want to test all of python 3.7, 3.8, and pypy, use tox (this is what
```
absurdfarce, line 179: Nice find!  I must've missed this one when I put out the last release 🤦

tests/integration/long/test_ssl.py
```diff
@@ -28,7 +28,7 @@
 
 if not hasattr(ssl, 'match_hostname'):
     try:
-        from backports.ssl_match_hostname import match_hostname
+        from ssl import match_hostname
```
absurdfarce, line 31: This seems like exactly the right idea.

I'll also note that ssl.match_hostname has been deprecated since Python 3.7 and is actually removed in 3.12 (more on that in [PYTHON-1331](https://datastax-oss.atlassian.net/browse/PYTHON-1331)) so these changes will likely be subsequently changed based on whatever happens with that ticket (which is next in my queue).  I very much like the other changes in this PR, though, so my inclination is to merge this now and just lay any fixes for PYTHON-1331 on top of this in their own PR.

absurdfarce (approved):
Jenkins is running against these changes now.  Assuming there aren't any issues there (and I very much doubt there will be) I'll merge this in right away.

absurdfarce:
Jenkins build looks good (test failures were all expected given current conditions)

PR #1187, opened 2023-11-10
PYTHON-1366 Handle removal of asyncore in Python 3.12

absurdfarce:
Goal is to maintain existing behaviour for Python 3.8 through 3.11.  For Python 3.12 return a specific exception type with a message indicating that asyncore won't work on Python 3.12 due to it's removal.

We also change the load process for the libs to use a reduce op.  We'll try to load working event loops in a specific sequence; if we're able to load one we short-circuit the rest, otherwise we keep trying.  Return a single exception which contains the full set of failures for loading each event loop implementation.

### DIFF
--- a/tests/integration/standard/test_connection.py
+++ b/tests/integration/standard/test_connection.py
@@ -443,6 +443,8 @@ class AsyncoreConnectionTests(ConnectionTests, unittest.TestCase):
     def setUp(self):
         if is_monkey_patched():
             raise unittest.SkipTest("Can't test asyncore with monkey patching")
+        if AsyncoreConnection is None:
+            raise unittest.SkipTest('Unable to import asyncore module')
         ConnectionTests.setUp(self)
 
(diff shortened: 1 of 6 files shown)


tests/integration/standard/test_connection.py
```diff
@@ -440,6 +443,8 @@ class AsyncoreConnectionTests(ConnectionTests, unittest.TestCase):
     def setUp(self):
         if is_monkey_patched():
             raise unittest.SkipTest("Can't test asyncore with monkey patching")
+        if AsyncoreConnection is None:
+            raise unittest.SkipTest('Unable to import asyncore module')
```
absurdfarce, line 447: Necessary to avoid executing this test on Python 3.12

absurdfarce:
Jenkins run with these changes confirms that with this change we're only seeing failures for other known issues on Python 3.12.

PR #1188, opened 2023-11-13
PYTHON-1368 Avoid installing DSE deps + executing DSE tests for Python 3.12

absurdfarce:


absurdfarce:
Test runs against DSE now complete normally after this change

PR #1190, opened 2023-11-14
Removed deprecated ez_setup file (use setup.py instead)

bschoening:
Easy_install and ez_setup.py were deprecated a while ago. Ez_setup.py hasn't been modified since 2014.

History for ez_setup.py
Commits on Apr 16, 2014
    Fix ez_setup
    Michael Figuiere committed on Apr 16, 2014
    Add License headers
    Michael Figuiere committed on Apr 16, 2014

Commits on Oct 4, 2013
    Switch back to plain setuptools, packaging tweaks
    thobbs committed on Oct 4, 2013
End of commit history for this file


bschoening:
It seems setup imports ez_setup, so this is more complex to remove/refactor

PR #1191, opened 2023-11-20
PYTHON-1331 ssl.match_hostname() is deprecated in 3.7

absurdfarce:


Jenkinsfile
```diff
@@ -196,7 +196,7 @@ def initializeEnvironment() {
   }
 
   sh label: 'Install unit test modules', script: '''#!/bin/bash -lex
-    pip install pynose nose-ignore-docstring nose-exclude service_identity
+    pip install nose-ignore-docstring nose-exclude service_identity
```
absurdfarce, line 199: Moved into test-requirements (replacing nose) rather than keeping it in the Jenkinsfile

cassandra/cluster.py
```diff
@@ -165,10 +165,12 @@ def _connection_reduce_fn(val,import_fn):
...
     return (rv or import_result, excs)
 
+log = logging.getLogger(__name__)
+
 conn_fns = (_try_gevent_import, _try_eventlet_import, _try_libev_import, _try_asyncore_import)
 (conn_class, excs) = reduce(_connection_reduce_fn, conn_fns, (None,[]))
-if excs:
-    raise DependencyException("Exception loading connection class dependencies", excs)
+if not conn_class:
+    raise DependencyException("Unable to load a default connection class", excs)
```
absurdfarce, line 173: A fix for an issue discovered in testing.  Goal of the changes to support asyncore as an optional dependency was to raise an error if we were unable to find a workable event loop at all.  Unfortunately my original impl just looked for the presence of any exception in any event loop impl which is... not at all the same thing.

cassandra/connection.py
```diff
@@ -852,6 +858,28 @@ def factory(cls, endpoint, timeout, *args, **kwargs):
...
+        if certfile:
+            rv.load_cert_chain(certfile, keyfile)
+        ca_certs = opts.pop('ca_certs', None)
+        if ca_certs:
+            rv.load_verify_locations(ca_certs)
+        ciphers = opts.pop('ciphers', None)
+        if ciphers:
+            rv.set_ciphers(ciphers)
+
+        return rv
```
absurdfarce, line 881: Follows the idea roughly sketched out in [Python docs for ssl.wrap_socket()](https://docs.python.org/3.11/library/ssl.html#ssl.wrap_socket)

absurdfarce (commented):
Local testing confirms that this change gets identical behaviour for a simple client using Python 3.8, 3.10 and 3.12 (see notes on the ticket for a lot more detail about this testing).  Next step is probably to try to bring test_ssl.py back from the dead, confirm that it's good with the existing impl and then try it out with these changes.

absurdfarce:
Tested by running the following ops using Python 3.8, 3.10 and 3.12 against a C* instance with client_encryption_options enabled:

```
client_encryption_options:
  enabled: true
  keystore: /work/git/working/python/client/client.jks
  keystore_password: password
  require_client_auth: true
  truststore: /work/git/working/python/rootca/ca_truststore.jks
  truststore_password: password
```

The following scenarios were tested:

* Basic SSL connectivity (using a CA-signed cert)
* Same with a self-signed cert, confirm that connection fails
* Specify invalid server_hostname in ssl_options, confirm that connection fails

All tests behaved as expected

PR #1195, opened 2024-01-10
Replace deprecated method assertRaisesRegexp for assertRaisesRegex

eamanu:
The legacy unittest class method assertRaisesRegexp was deprecated in favour of assertRaisesRegex


absurdfarce:
Great catch @eamanu, thanks for sending this one in!

absurdfarce (approved):

PR #1197, opened 2024-02-24
clean up last "import six"

a-detiste:


absurdfarce:
Thanks for the PR @a-detiste!  I agree it would be nice to tighten up these last few areas in order to get six _completely_ out of the code base.

a-detiste:
Hi, I ve signed the CLA.

appveyor/appveyor.ps1
```diff
@@ -50,7 +50,7 @@ if(!(Test-Path $simulacron_jar)) {
 }
 
 # Install Python Dependencies for CCM.
-Start-Process python -ArgumentList "-m pip install psutil pyYaml six numpy" -Wait -NoNewWindow
+Start-Process python -ArgumentList "-m pip install psutil pyYaml numpy" -Wait -NoNewWindow
```
absurdfarce, line 53: This one probably should stay.  The usage here is to add the dependencies that ccm needs (in order to support starting and stopping clusters for unit tests) and as of this writing it [does still require six](https://github.com/riptano/ccm/blob/master/requirements.txt).  It's not a _huge_ thing since we don't really rely on AppVeyor much at the moment but I also don't want to delete it as it might cause some level of confusion later.

absurdfarce (changes requested):
Good changes overall, but I think we do want to keep the usage in the AppVeyor PS code.

a-detiste:
I force pushed to remove the change

absurdfarce:
Excellent, thanks for the quick turnaround @a-detiste !  I think we're ready to go here!

absurdfarce (approved):

PR #1198, opened 2024-02-28
PYTHON-1378 Expand search directories for includes

absurdfarce:
Need to make sure we can find include directories for packages installed on the Docker images used by cibuildwheel

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -145,13 +145,15 @@ murmur3_ext = Extension('cassandra.cmurmur3',
                         sources=['cassandra/cmurmur3.c'])
 
 libev_includes = ['/usr/include/libev', '/usr/local/include', '/opt/local/include', '/usr/include']
+libev_libdirs = ['/usr/local/lib', '/opt/local/lib', '/usr/lib64']
 if is_macos:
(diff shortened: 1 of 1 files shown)


absurdfarce:
Note that this will also require the merge of [a fix](https://github.com/datastax/python-driver-wheels/pull/9) on the datastax/python-driver-wheels side as well in order to leverage these changes.

setup.py
```diff
@@ -138,15 +138,20 @@ class BuildFailed(Exception):
...
 
+is_windows = sys.platform.startswith('win32')
+is_macos = sys.platform.startswith('darwin')
 
 murmur3_ext = Extension('cassandra.cmurmur3',
                         sources=['cassandra/cmurmur3.c'])
 
+libev_includes = ['/usr/include/libev', '/usr/local/include', '/opt/local/include', '/usr/include']
+if is_macos:
+    libev_includes.extend(['/opt/homebrew/include', os.path.expanduser('~/homebrew/include')])
```
absurdfarce, line 149: Just making the [recent changes](https://github.com/datastax/python-driver/commit/2c61ab22792475445d21b77bd165cf330ee9b87b) to support builds using Homebrew constrained to OSX builds only

PR #1201, opened 2024-03-27
remove dependency on old mock external module

a-detiste:
Hi again,

After the `six` removal, here's more Python2-era removal.

https://github.com/testing-cabal/mock 

**mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.**

### DIFF
--- a/cassandra/cqlengine/functions.py
+++ b/cassandra/cqlengine/functions.py
@@ -17,16 +17,8 @@ from datetime import datetime
 
 from cassandra.cqlengine import UnicodeMixin, ValidationError
 
-import sys
-
-if sys.version_info >= (2, 7):
-    def get_total_seconds(td):
-        return td.total_seconds()
-else:
-    def get_total_seconds(td):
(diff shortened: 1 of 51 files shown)


a-detiste:
ping

absurdfarce:
Greetings @a-detiste, nice to hear from you again!

I think I'm good with the mock changes but we've definitely had some problems with the removal of the absolute_import imports from future.  Those issues are referenced on #1184 and discussed in more detail in [PYTHON-1348](https://datastax-oss.atlassian.net/browse/PYTHON-1348).

So for the moment would you mind removing the absolute_import removals and just leaving the rest of the changes in this PR?  I think that change could go in pretty quickly.

a-detiste:
Hi, let's try again this trimmed down PR. The ```__future__``` statements are harmless and can be removed later.

absurdfarce (approved):

absurdfarce:
Thanks for this one @a-detiste!

PR #1202, opened 2024-04-19
Explicitly select Java8 via Jabba before running integration tests

absurdfarce:
Fix for unexpected test issues after recent Jenkins upgrade forced us to move to Java11 as the default for all runners

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -272,6 +272,9 @@ def executeStandardTests() {
     . ${HOME}/environment.txt
     set +o allexport
 
+    . ${JABBA_SHELL}
+    jabba use 1.8
+
     SIMULACRON_JAR="${HOME}/simulacron.jar"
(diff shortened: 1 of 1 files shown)


absurdfarce:
Confirmed that after this change we no longer see failures in DSE integration tests around an inability to support Java11

Jenkinsfile
```diff
@@ -272,6 +272,9 @@ def executeStandardTests() {
     . ${HOME}/environment.txt
     set +o allexport
 
+    . ${JABBA_SHELL}
```
msmygit, line 275: nit: should we extract this as a method and reuse it where needed?

msmygit (approved):
lgtm

Jenkinsfile
absurdfarce: Perhaps, but there are some downsides to that approach.  Each of these shell blocks start up with zero context so you have to configure everything each time.  There are some config options in the openstack-jenkins-drivers repo which provides tools to help a bit with this; that's where environment.txt is built, for instance.  So we could add the function there and just call it here; that would be the easiest way to make it fairly common.

Problem with that is that your Jenkinsfile is now more explicitly tied to the configuration of the runner.  We already have numerous other cases of that in the build right now so it's not like you're creating a _new_ dependency there.  But if we were wanting to change that I'd want it to go in the other way; I'd rather have this build be more self-contained so that we can get as close as possible to the idea of taking this Jenkinsfile, dropping it in a new Jenkins install and largely having what we need.  We'll never have that completely; there's too much config built into openstack-jenkins-drivers as-is.  But I'd like to at least keep moving in that direction.

SiyaoIsHiding (approved):

joao-r-reis (approved):
Idk enough about the CI situation for this driver to know if we're now back to the state we were in before the jenkins upgrade irt the test failures but the changes lgtm.

PR #1204, opened 2024-04-30
Update installation.rst with Python3 print() syntax

bschoening:


absurdfarce:
Ah, good catch (as usual) @bschoening .  Works for 3.8 as expected.

absurdfarce (approved):

PR #1207, opened 2024-05-07
Update geomet to latest in requirements.txt

bschoening:
Geomet 0.3 only officially supports EOL Python versions 2.7 ... 3.7 and has an unnecessary dependency on the python 2 compatibility library 'six'.  Upgrading to 1.1.0 resolves this.

This a blocking issue for a Homebrew [PR](https://github.com/Homebrew/homebrew-core/commit/7bad50e808ee856a8d8a432a302215e81f0bef9a#commitcomment-141738577) to remove six from the Cassandra package


absurdfarce:
Triggered a PR build for this, just to make sure there aren't any obvious regressions with the unit/integration tests

bschoening:
@absurdfarce I'm unable to see the pr-merge output which is on dsinternal.org, but can you share what failed?

absurdfarce:
Hey @bschoening, thanks for following up!  The failures for that run are all known intermittent test failures that creep up from time to time.  I've tried to fix a few of the more obvious ones but my expectation is that most (maybe all?) will go away once we finish the move to pytest.

The one thing I do still need to check here is how this behaves with tests.unit.advanced.test_geometry; that test wasn't run for the test configuration powering that Jenkins run.  If that test is good I think I'm satisfied there's no API change in this version of geomet.  Worth noting that the test_geometry unit test does employ a skipUnless annotation to skip several test cases if geomet isn't installed.

absurdfarce:
A quick update: tests.unit.advanced.test_geometry passes locally for me with geomet 1.1 in play.

absurdfarce:
I also re-ran tests.integration.advanced.test_geometry against DSE 6.8.30 locally and observed results similar to those reported by Jenkins (i.e. all passed except for two methods on BasicGeometricPointTypeTest which both have unittest.skip annotations in place).  At this point I don't have any reason to believe there's a problem with the upgrade here so I'm calling this good.

bschoening:
There is an open PR to fix this [#1236](https://github.com/datastax/python-driver/pull/1236).

PR #1209, opened 2024-05-09
PYTHON-1385: Remove nosetests references from AppVeyor build

absurdfarce:


bschoening:
@absurdfarce one more that can be close as it appears already fixed.

absurdfarce:
Yup, agreed @bschoening .... looks like this was fixed up with the general conversion to pytest

PR #1210, opened 2024-05-21
Update README.rst with badges for version and license

bschoening:


absurdfarce:
Nice, I'm good with that.

absurdfarce (approved):

PR #1213, opened 2024-06-23
cassandra/query: use timezone specific API to avoid deprecated warning

tchaikov:
before this change, when testing with cqlsh, we have warnings like:

```
  <frozen importlib._bootstrap>:488: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
````

in this change, we replace the deprecated API with timezone-aware API, to avoid this warning.

### DIFF
--- a/tests/unit/cython/types_testhelper.pyx
+++ b/tests/unit/cython/types_testhelper.pyx
@@ -38,7 +38,7 @@ def test_datetype(assert_equal):
         cdef BytesIOReader reader
         cdef Buffer buf
 
-        dt = datetime.datetime.utcfromtimestamp(timestamp)
+        dt = datetime.datetime.fromtimestamp(timestamp, tz=datetime.timezone.utc)
 
         bytes = io.BytesIO()
         write_value(bytes, DateType.serialize(dt, 0))
@@ -52,7 +52,7 @@ def test_datetype(assert_equal):
     # deserialize
     # epoc
     expected = 0
-    assert_equal(deserialize(expected), datetime.datetime.utcfromtimestamp(expected))
+    assert_equal(deserialize(expected), datetime.datetime.fromtimestamp(expected, tz=datetime.timezone.utc).replace(tzinfo=None))
 
     # beyond 32b
     expected = 2 ** 33
--- a/cassandra/cqlengine/columns.py
+++ b/cassandra/cqlengine/columns.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from copy import deepcopy, copy
-from datetime import date, datetime, timedelta
+from datetime import date, datetime, timedelta, timezone
 import logging
 from uuid import UUID as _UUID
 
@@ -551,7 +551,7 @@ class DateTime(Column):
         elif isinstance(value, date):
             return datetime(*(value.timetuple()[:6]))
 
-        return datetime.utcfromtimestamp(value)
+        return datetime.fromtimestamp(value, tz=timezone.utc).replace(tzinfo=None)
 
     def to_database(self, value):
         value = super(DateTime, self).to_database(value)
--- a/cassandra/query.py
+++ b/cassandra/query.py
@@ -19,7 +19,7 @@ queries.
 """
 
 from collections import namedtuple
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 import re
 import struct
 import time
@@ -1085,7 +1085,7 @@ class TraceEvent(object):
 
     def __init__(self, description, timeuuid, source, source_elapsed, thread_name):
         self.description = description
-        self.datetime = datetime.utcfromtimestamp(unix_time_from_uuid1(timeuuid))
+        self.datetime = datetime.fromtimestamp(unix_time_from_uuid1(timeuuid), tz=timezone.utc)
         self.source = source
         if source_elapsed is not None:
             self.source_elapsed = timedelta(microseconds=source_elapsed)
--- a/cassandra/util.py
+++ b/cassandra/util.py
@@ -38,8 +38,8 @@ except:
 
 from cassandra import DriverException
 
-DATETIME_EPOC = datetime.datetime(1970, 1, 1)
-UTC_DATETIME_EPOC = datetime.datetime.utcfromtimestamp(0)
+DATETIME_EPOC = datetime.datetime(1970, 1, 1).replace(tzinfo=None)
+UTC_DATETIME_EPOC = datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc).replace(tzinfo=None)
 
 _nan = float('nan')
 
--- a/tests/integration/cqlengine/columns/test_validation.py
+++ b/tests/integration/cqlengine/columns/test_validation.py
@@ -15,7 +15,7 @@
 import unittest
 
 import sys
-from datetime import datetime, timedelta, date, tzinfo, time
+from datetime import datetime, timedelta, date, tzinfo, time, timezone
 from decimal import Decimal as D
 from uuid import uuid4, uuid1
 from packaging.version import Version
@@ -97,7 +97,7 @@ class TestDatetime(BaseCassEngTestCase):
         dt_value = 1454520554
         self.DatetimeTest.objects.create(test_id=5, created_at=dt_value)
         dt2 = self.DatetimeTest.objects(test_id=5).first()
-        self.assertEqual(dt2.created_at, datetime.utcfromtimestamp(dt_value))
+        self.assertEqual(dt2.created_at, datetime.fromtimestamp(dt_value, tz=timezone.utc).replace(tzinfo=None))
 
     def test_datetime_large(self):
         dt_value = datetime(2038, 12, 31, 10, 10, 10, 123000)
@@ -809,7 +809,7 @@ class TestTimeUUIDFromDatetime(BaseCassEngTestCase):
         assert isinstance(uuid, UUID)
 
         ts = (uuid.time - 0x01b21dd213814000) / 1e7 # back to a timestamp
-        new_dt = datetime.utcfromtimestamp(ts)
+        new_dt = datetime.fromtimestamp(ts, tz=timezone.utc).replace(tzinfo=None)
 
         # checks that we created a UUID1 with the proper timestamp
         assert new_dt == dt
--- a/tests/integration/cqlengine/model/test_model_io.py
+++ b/tests/integration/cqlengine/model/test_model_io.py
@@ -15,7 +15,7 @@ import unittest
 
 from uuid import uuid4, UUID
 import random
-from datetime import datetime, date, time
+from datetime import datetime, date, time, timezone
 from decimal import Decimal
 from operator import itemgetter
 
@@ -197,13 +197,13 @@ class TestModelIO(BaseCassEngTestCase):
 
         sync_table(AllDatatypesModel)
 
-        input = ['ascii', 2 ** 63 - 1, bytearray(b'hello world'), True, datetime.utcfromtimestamp(872835240),
+        input = ['ascii', 2 ** 63 - 1, bytearray(b'hello world'), True, datetime.fromtimestamp(872835240, tz=timezone.utc).replace(tzinfo=None),
                  Decimal('12.3E+7'), 2.39, 3.4028234663852886e+38, '123.123.123.123', 2147483647, 'text',
                  UUID('FE2B4360-28C6-11E2-81C1-0800200C9A66'), UUID('067e6162-3b6f-4ae2-a171-2470b63dff00'),
                  int(str(2147483647) + '000')]
 
         AllDatatypesModel.create(id=0, a='ascii', b=2 ** 63 - 1, c=bytearray(b'hello world'), d=True,
-                                 e=datetime.utcfromtimestamp(872835240), f=Decimal('12.3E+7'), g=2.39,
+                                 e=datetime.fromtimestamp(872835240, tz=timezone.utc), f=Decimal('12.3E+7'), g=2.39,
                                  h=3.4028234663852886e+38, i='123.123.123.123', j=2147483647, k='text',
                                  l=UUID('FE2B4360-28C6-11E2-81C1-0800200C9A66'),
                                  m=UUID('067e6162-3b6f-4ae2-a171-2470b63dff00'), n=int(str(2147483647) + '000'),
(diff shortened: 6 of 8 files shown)

### CODE REVIEW

**Comment by absurdfarce** [comment]
Thanks for the PR @tchaikov !

**Comment by tchaikov** [comment]
@absurdfarce yeah, i did.

**Comment by absurdfarce** [comment]
Excellent, thank you @tchaikov!

I like this change; it looks like the deprecation came in with Python 3.12 and we're doing a lot of work now to try and stabilize our story on that Python version.  If we're going to make this change, though, I think we should make it everywhere datetime.datetime.utcfromtimestamp() is used throughout the code base.  That means we're looking at the following instances:

```
$ rg utcfromtimestamp
cassandra/util.py
42:UTC_DATETIME_EPOC = datetime.datetime.utcfromtimestamp(0)

cassandra/cqlengine/columns.py
554:        return datetime.utcfromtimestamp(value)

cassandra/query.py
1088:        self.datetime = datetime.utcfromtimestamp(unix_time_from_uuid1(timeuuid))

tests/unit/test_types.py
204:        now_datetime = datetime.datetime.utcfromtimestamp(now_time_seconds)
215:        self.assertEqual(DateType.deserialize(int64_pack(1000 * expected), 0), datetime.datetime.utcfromtimestamp(expected))

tests/unit/cython/types_testhelper.pyx
41:        dt = datetime.datetime.utcfromtimestamp(timestamp)
55:    assert_equal(deserialize(expected), datetime.datetime.utcfromtimestamp(expected))

tests/integration/cqlengine/columns/test_validation.py
100:        self.assertEqual(dt2.created_at, datetime.utcfromtimestamp(dt_value))
812:        new_dt = datetime.utcfromtimestamp(ts)

tests/integration/cqlengine/model/test_udts.py
275:                             e=datetime.utcfromtimestamp(872835240), f=Decimal('12.3E+7'), g=2.39,

tests/integration/cqlengine/model/test_model_io.py
200:        input = ['ascii', 2 ** 63 - 1, bytearray(b'hello world'), True, datetime.utcfromtimestamp(872835240),
206:                                 e=datetime.utcfromtimestamp(872835240), f=Decimal('12.3E+7'), g=2.39,
```

Are you able to update this PR to include fixes for all these occurrences?  If not I can certainly create a PR to do so.  Just let me know!

Thanks again!

**Comment by tchaikov** [comment]
@absurdfarce hi Bret, thanks for your review.

```console
$ rg utcfromtimestamp || echo done
done
```

absurdfarce (changes requested):
Looking pretty good @tchaikov!  It looks like there are a few places where the presence or absence of timezone information in the datetime object needs to be a bit more consistent but overall I think we're pretty close here.

tchaikov:
v3:

- use local tz for `DATETIME_EPOC`
- remove timezone in the change in [tests/integration/cqlengine/model/test_model_io.py](https://github.com/datastax/python-driver/pull/1213/files#diff-d9d6dead22089138e2ec5736b2cee9b2b1b42b0d2e9ce13d998c6e1ec3746fba)
- remove timezone in the change in [test_validation.py](https://github.com/datastax/python-driver/pull/1213/files#diff-548fd1901ab8b3b00269ac1573017f9b5c4879300d07f0ebb63b6f96b351ca7a)
- remove timezone in the change in [tests/integration/cqlengine/model/test_udts.py](https://github.com/datastax/python-driver/pull/1213/files#diff-7db0273d3a797b7589818a4d1a495e3761c24c83bebc1ce6f238df4bf7908a21)

tchaikov:
@absurdfarce hi Bret, thank you for the detailed and constructive review. it's very helpful. i rely on the CI to identify the test failures. am i able to access the jenins job you referenced in the comment as an external contributor?

absurdfarce:
Hey @tchaikov, thanks for all the work to address the various test issues!  Unfortunately I can't make the Jenkins results public but I'm happy to try another run with the various changes you've made here.  That's been kicked off; I'll report back here when it completes and we can see whether all the outstanding test issues have been addressed!

absurdfarce:
Hey @tchaikov, Jenkins run just finished and it looks like we still have a problem.  I observed the following on all Python + Cassandra combinations we test against:

```
Failed
Build and Test / 3.11, py3.8.16, Cython / tests.integration.cqlengine.model.test_model_io.TestModelIO.test_can_insert_model_with_all_column_types (from nosetests)

Error Message
datetime.datetime(1997, 8, 29, 6, 14, tzinfo=datetime.timezone.utc) != datetime.datetime(1997, 8, 29, 6, 14)

Stacktrace
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/jenkins/workspace/drivers_python_oss_PR-1213/tests/integration/cqlengine/model/test_model_io.py", line 216, in test_can_insert_model_with_all_column_types
    self.assertEqual(input[i], output[chr(i_char)])
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: datetime.datetime(1997, 8, 29, 6, 14, tzinfo=datetime.timezone.utc) != datetime.datetime(1997, 8, 29, 6, 14)
```

I can repro this error by running the test locally, so that's a good thing.

Looks like that test explicitly compare the "input" sequence to whatever it gets back from the cqlengine model.  And since [elsewhere in this PR](https://github.com/datastax/python-driver/pull/1213/files#r1659093797) we explicitly change the model to return a naive datetime we need to compare against one here as well.  So "input" needs to contain a naive datetime which means this change will fix the issue:

```diff
diff --git a/tests/integration/cqlengine/model/test_model_io.py b/tests/integration/cqlengine/model/test_model_io.py
index c4880d4f..9414d83d 100644
--- a/tests/integration/cqlengine/model/test_model_io.py
+++ b/tests/integration/cqlengine/model/test_model_io.py
@@ -197,7 +197,7 @@ class TestModelIO(BaseCassEngTestCase):
 
     	sync_table(AllDatatypesModel)
 
-    	input = ['ascii', 2 ** 63 - 1, bytearray(b'hello world'), True, datetime.fromtimestamp(872835240, tz=timezone.utc),
+    	input = ['ascii', 2 ** 63 - 1, bytearray(b'hello world'), True, datetime.fromtimestamp(872835240, tz=timezone.utc).replace(tzinfo=None),
              	Decimal('12.3E+7'), 2.39, 3.4028234663852886e+38, '123.123.123.123', 2147483647, 'text',
              	UUID('FE2B4360-28C6-11E2-81C1-0800200C9A66'), UUID('067e6162-3b6f-4ae2-a171-2470b63dff00'),
              	int(str(2147483647) + '000')]
```

It's also worth noting that the referenced change to cassandra.cqlengine.columns will always return a naive datetime regardless of input.  That means we can supply a datetime with or without time zone info to our model creation call.  That in turn means that this change also passes the test without issue:

```diff
diff --git a/tests/integration/cqlengine/model/test_model_io.py b/tests/integration/cqlengine/model/test_model_io.py
index 9414d83d..ce43422e 100644
--- a/tests/integration/cqlengine/model/test_model_io.py
+++ b/tests/integration/cqlengine/model/test_model_io.py
@@ -203,7 +203,7 @@ class TestModelIO(BaseCassEngTestCase):
              	int(str(2147483647) + '000')]
 
     	AllDatatypesModel.create(id=0, a='ascii', b=2 ** 63 - 1, c=bytearray(b'hello world'), d=True,
-                             	e=datetime.fromtimestamp(872835240, tz=timezone.utc).replace(tzinfo=None), f=Decimal('12.3E+7'), g=2.39,
+                             	e=datetime.fromtimestamp(872835240, tz=timezone.utc), f=Decimal('12.3E+7'), g=2.39,
                              	h=3.4028234663852886e+38, i='123.123.123.123', j=2147483647, k='text',
                              	l=UUID('FE2B4360-28C6-11E2-81C1-0800200C9A66'),
                              	m=UUID('067e6162-3b6f-4ae2-a171-2470b63dff00'), n=int(str(2147483647) + '000'),
```

Note that the first change above is required to make the test pass.  The second change is entirely optional.

tchaikov:
v2:

- `tests/integration/cqlengine/model/test_model_io.py`: use a offset-naive timestamp in `AllDatatypesModel.input` 
- `tests/integration/cqlengine/model/test_model_io.py`: use a offset-aware timestamp for `e` `AllDatatypesModel.e`, simpler this way.

tests/unit/cython/types_testhelper.pyx
```diff
@@ -38,7 +38,7 @@ def test_datetype(assert_equal):
         cdef BytesIOReader reader
         cdef Buffer buf
 
-        dt = datetime.datetime.utcfromtimestamp(timestamp)
+        dt = datetime.datetime.fromtimestamp(timestamp, tz=datetime.timezone.utc)
```
absurdfarce, line 41: The original impl (using utcfromtimestamp()) returns a naive datetime and this version doesn't.  But after looking into this  a bit it doesn't actually matter.  This datetime is immediately serialized and then deserialized in order to test the end-to-end process.  All that really matters here are the time values; any timezone info is lost in that process.

tests/unit/cython/types_testhelper.pyx
```diff
@@ -52,11 +52,11 @@ def test_datetype(assert_equal):
     # deserialize
     # epoc
     expected = 0
-    assert_equal(deserialize(expected), datetime.datetime.utcfromtimestamp(expected))
+    assert_equal(deserialize(expected), datetime.datetime.fromtimestamp(expected, tz=datetime.timezone.utc))
 
     # beyond 32b
     expected = 2 ** 33
-    assert_equal(deserialize(expected), datetime.datetime(2242, 3, 16, 12, 56, 32))
+    assert_equal(deserialize(expected), datetime.datetime(2242, 3, 16, 12, 56, 32, tz=datetime.timezone.utc))
```
absurdfarce, line 59: Note that we're dealing with a datetime constructor here rather than a utcfromtimestamp() invocation.  We don't _need_ to replace this in order to handle any deprecation concerns.

More to the point: the datetime coming out of deserialize() is naive so this actually introduces a test failure.  I believe this should stay as `assert_equal(deserialize(expected), datetime.datetime(2242, 3, 16, 12, 56, 32))`

tests/unit/cython/types_testhelper.pyx
```diff
@@ -52,11 +52,11 @@ def test_datetype(assert_equal):
     # deserialize
     # epoc
     expected = 0
-    assert_equal(deserialize(expected), datetime.datetime.utcfromtimestamp(expected))
+    assert_equal(deserialize(expected), datetime.datetime.fromtimestamp(expected, tz=datetime.timezone.utc))
```
absurdfarce, line 55: This one took some digging to sort out but I believe I have it now.  deserialize() returns a naive datetime based on the UTC values so you're correct to continue to resolve the datetime against UTC... but we shouldn't have the timezone in place for what we return.  As a result this needs to become `assert_equal(deserialize(expected), datetime.datetime.fromtimestamp(expected, tz=datetime.timezone.utc).replace(tzinfo=None))`

absurdfarce (changes requested):
Good news is that with your last round of changes @tchaikov we did get rid of some of the test failures in the Jenkins run.  Bad news is that when I looked originally I missed a few test failures that were hiding in amongst the failures we already discussed.  The new failures look like this:

```
Build and Test / 3.11, py3.8.16, Cython / tests.unit.cython.test_types.TypesTest.test_datetype (from nosetests)

Error Message
datetime.datetime(1970, 1, 1, 0, 0) != datetime.datetime(1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
Stacktrace
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/jenkins/workspace/drivers_python_oss_PR-1213/tests/unit/cython/test_types.py", line 25, in test_datetype
    types_testhelper.test_datetype(self.assertEqual)
  File "tests/unit/cython/types_testhelper.pyx", line 55, in tests.unit.cython.types_testhelper.test_datetype
    assert_equal(deserialize(expected), datetime.datetime.fromtimestamp(expected, tz=datetime.timezone.utc))
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/home/jenkins/.pyenv/versions/3.8.16/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: datetime.datetime(1970, 1, 1, 0, 0) != datetime.datetime(1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
```

I tracked down the root cause of these failures and added some comments to types_testhelper.pyx which should address them.

tests/unit/cython/types_testhelper.pyx
tchaikov: yeah, silly me! reverted this change.

tests/unit/cython/types_testhelper.pyx
tchaikov: yeah, i missed it in previous fixes.

tchaikov:
v3:

- reverted the change to `datetime.datetime()` call
- add `replace(tzinfo=None)` to yet another `datetime.datetime.fromtimestamp()` call

tchaikov:
@absurdfarce hi Bret, thank you very much! was tied up yesterday. just sent another revision. hopefully it's better now.

absurdfarce:
Last CI run looked great, so I think we're ready to go here!  Thanks for the PR (and your perseverance) @tchaikov!

tchaikov:
Thank you, Bret!

PR #1214, opened 2024-07-11
PYTHON-1297 Convert from nose to pytest

absurdfarce:

PR #1215, opened 2024-07-11
PYTHON-1297 Convert to pytest for running unit tests

absurdfarce:
This branch represents a unification of separate work that has been done by @bschoening in #1171 and @weideng1 in #1214.  Additional work will be added on as necessary to get to a working version which will support pytest as the default for running all tests.  We'll make a best-efforts push to maintain compatibility with pynose but there are no guarantees on that front; in the end the only fully supported test runner will be pytest.


bschoening:
There are nosetest references in .gitignore and setup.py also.

absurdfarce:
Good call @bschoening .... I'll yank those too.

absurdfarce:
Test build at this point is only showing a few new errors, most of which appear to be PYTHON-1389.  I'm going to fix those quickly and take a stab at the setup.py changes @bschoening mentioned above.  After that we should be good for a retest.

absurdfarce:
A quick update on this.  We're down to a reasonable number of test failures (38 as of this writing, which is way better than the several thousand we were seeing).  We were seeing some failures due to [PYTHON-1389](https://datastax-oss.atlassian.net/browse/PYTHON-1389) but I put a fix in for that which seems to have addressed the issue.  Nearly all of the failures I'm seeing at the moment are covered by one of two tickets:

* [PYTHON-1390](https://datastax-oss.atlassian.net/browse/PYTHON-1390) (just filed)
* A reappearance of something that looks very much like [PYTHON-1347](https://datastax-oss.atlassian.net/browse/PYTHON-1347).  We definitely had this fixed when the original PR for that ticket went but it's now showing up for Python 3.12 builds only.

Both of these issues need to be addressed, but as of this writing I don't have any reason to believe they're related to the pytest conversion... so I'm inclined to not hold on merging this functionality due only to these issues.

Opinions welcome.  @bschoening do you think I'm off-base here?

bschoening:
@absurdfarce so, pytests are pretty clean except for when using Python 3.12?  If so, yes, I wouldn't hold of on merging.

absurdfarce:
Yup, you've got the gist of it @bschoening.  There are a few other failures from flaky tests and the asyncore tests need to have a version bound for Python 3.12 but outside of that and the issues mentioned above the tests look pretty good.

I'm inclined to agree with you.  It seems to make more sense to get this in sooner rather than later; that way we can use pytest to help address the other issues mentioned above.

tests/unit/io/test_twistedreactor.py
```diff
@@ -66,7 +66,9 @@ def setUp(self):
         self.tr.protocol = self.obj_ut
 
     def tearDown(self):
-        pass
+        loop = twistedreactor.TwistedConnection._loop
+        if not loop._reactor_stopped():
+            loop._cleanup()
```
absurdfarce, line 71: This test started failing after moving to pytest, apparently because the order of operations changed.  It looks like the old infrastructure included implicit assumptions about when the reactor was started... and now that pytest executed tests in a different order those assumptions led to unexpected test failures.  The goal of this change is to make the test more robust by forcing each test case to do a better job cleaning up after itself.

cassandra/io/twistedreactor.py
```diff
@@ -102,6 +102,9 @@ def maybe_start(self):
                 self._thread.start()
                 atexit.register(partial(_cleanup, weakref.ref(self)))
 
+    def _reactor_stopped(self):
+        return reactor._stopped
+
```
absurdfarce, line 107: Used in order-of-operations test fix discussed below

tests/unit/io/test_asyncorereactor.py
```diff
@@ -54,7 +62,8 @@ def tearDownClass(cls):
             except:
                 pass
 
-
+has_asyncore = Version(platform.python_version()) < Version("3.12.0")
+@unittest.skipUnless(has_asyncore, "asyncore has been removed in Python 3.12")
```
absurdfarce, line 66: Not part of the pytest conversion... just a drive-by cleanup while we're here

setup.py
```diff
@@ -16,16 +16,6 @@
...
-if __name__ == '__main__' and sys.argv[1] == "gevent_nosetests":
-    print("Running gevent tests")
-    from gevent.monkey import patch_all
-    patch_all()
-
-if __name__ == '__main__' and sys.argv[1] == "eventlet_nosetests":
-    print("Running eventlet tests")
-    from eventlet import monkey_patch
-    monkey_patch()
-
```
absurdfarce, line 28: This monkey patch in the gevent/eventlet case is now handled [in test infrastructure](https://github.com/datastax/python-driver/blob/3.29.1/tests/__init__.py#L67-L84) so we shouldn't need to support it as a CLI arg as well.

PR #1216, opened 2024-07-12
PYTHON-1389 assertEquals is deprecated in Python 3.12

absurdfarce:


absurdfarce:
Local runs of tests.integration.standard.column_encryption.test_policies.ColumnEncryptionPolicyTest confirm that this test now passes on both Python 3.8 and 3.12.

PR #1217, opened 2024-07-23
PYTHON-1369 Extend driver vector support to arbitrary subtypes and fix handling of variable length types (OSS C* 5.0)

absurdfarce:


cassandra/marshal.py
```diff
@@ -143,3 +142,48 @@ def vints_pack(values):
...
+        if num_extra_bytes > 8:
+            raise ValueError('Value %d is too big and cannot be encoded as vint' % val)
+
+        # We can now store the last bits in the first byte
+        n = 8 - num_extra_bytes
+        v |= (0xff >> n << n)
+        rv.append(abs(v))
+
+    rv.reverse()
+    return bytes(rv)
```
absurdfarce, line 189: Adapted from vints_pack and vints_unpack above.  Those functions are different enough (built-in zig-zag encoding + tuples for incoming/outgoing data) that it seemed worthwhile to live with the code duplication here.

cassandra/__init__.py
```diff
@@ -744,9 +744,3 @@ def __init__(self, msg, excs=[]):
         if excs:
             complete_msg += ("\nThe following exceptions were observed: \n - " + '\n - '.join(str(e) for e in excs))
         Exception.__init__(self, complete_msg)
-
-class VectorDeserializationFailure(DriverException):
-    """
-    The driver was unable to deserialize a given vector
-    """
-    pass
```
absurdfarce, line 752: Added in PYTHON-1371 to indicate an attempt to decode a vector of an unsupported subtype.  With the other changes in this PR this exception is now completely unnecessary.

cassandra/cqltypes.py
```diff
@@ -1410,6 +1433,11 @@ class VectorType(_CassandraType):
     vector_size = 0
     subtype = None
 
+    @classmethod
+    def serial_size(cls):
+        serialized_size = cls.subtype.serial_size()
+        return cls.vector_size * serialized_size if serialized_size is not None else None
```
absurdfarce, line 1439: Here (as in many other points in this PR) we're following the [example of what's done](https://github.com/apache/cassandra/commit/ae537abc6494564d7254a2126465522d86b44c1e#diff-8d59be6f2cd342929e99b6da7f889239e85605cbb11c5b1044a5d04421099486R91-R93) on the server side.

tests/unit/test_types.py
```diff
@@ -309,8 +319,31 @@ def test_cql_quote(self):
         self.assertEqual(cql_quote('test'), "'test'")
         self.assertEqual(cql_quote(0), '0')
 
-    def test_vector_round_trip_types_with_serialized_size(self):
-        # Test all the types which specify a serialized size... see PYTHON-1371 for details
+    def _normalize_set(self, val):
+        if isinstance(val, set) or isinstance(val, util.SortedSet):
+            return frozenset([self._normalize_set(v) for v in val])
+        return val
```
absurdfarce, line 325: Really just about making the types line up.  The set codec returns a SortedSet by default and that class doesn't have all the niceties of frozenset so we convert here to simplify the equality comparison later.

absurdfarce:
Jenkins build at this point looks good: all failures appear to be known issues, most notably PYTHON-1390 and friends.

absurdfarce:
@SiyaoIsHiding the integration test is now complete and covers both tuples and UDTs.  At this point I think I've covered everything we intended to cover so this PR should be done (plus or minus changes coming out of review).  Would you mind taking another pass at this?

cassandra/encoder.py
```diff
@@ -217,3 +219,6 @@ def cql_encode_ipaddress(self, val):
         is suitable for ``inet`` type columns.
         """
         return "'%s'" % val.compressed
+
+    def cql_encode_decimal(self, val):
+        return self.cql_encode_float(float(val))
```
absurdfarce, line 224: Fixing a bug discovered during testing.  When using the drivers support for positional parameters in the execute() statement we weren't correctly expanding decimal types.

cassandra/cqltypes.py
```diff
@@ -1418,19 +1448,31 @@ def apply_parameters(cls, params, names):
...
+            return [cls.subtype.deserialize(byts[idx:idx + serialized_size], protocol_version) for idx in indexes]
+
+        idx = 0
+        rv = []
+        while (idx < len(byts)):
+            size, bytes_read = uvint_unpack(byts[idx:])
+            idx += bytes_read
+            rv.append(cls.subtype.deserialize(byts[idx:idx + size], protocol_version))
+            idx += size
+        return rv
```
SiyaoIsHiding, line 1463: Do we want to throw an error when the length of elements in the `byts` does not match the vector dimension definition? Same for `serialize`.

tests/integration/__init__.py
```diff
@@ -330,9 +330,10 @@ def _id_and_mark(f):
 greaterthanorequalcass36 = unittest.skipUnless(CASSANDRA_VERSION >= Version('3.6'), 'Cassandra version 3.6 or greater required')
 greaterthanorequalcass3_10 = unittest.skipUnless(CASSANDRA_VERSION >= Version('3.10'), 'Cassandra version 3.10 or greater required')
 greaterthanorequalcass3_11 = unittest.skipUnless(CASSANDRA_VERSION >= Version('3.11'), 'Cassandra version 3.11 or greater required')
-greaterthanorequalcass40 = unittest.skipUnless(CASSANDRA_VERSION >= Version('4.0-a'), 'Cassandra version 4.0 or greater required')
-lessthanorequalcass40 = unittest.skipUnless(CASSANDRA_VERSION <= Version('4.0-a'), 'Cassandra version less or equal to 4.0 required')
-lessthancass40 = unittest.skipUnless(CASSANDRA_VERSION < Version('4.0-a'), 'Cassandra version less than 4.0 required')
+greaterthanorequalcass40 = unittest.skipUnless(CASSANDRA_VERSION >= Version('4.0'), 'Cassandra version 4.0 or greater required')
+greaterthanorequalcass50 = unittest.skipUnless(CASSANDRA_VERSION >= Version('5.0-beta'), 'Cassandra version 5.0 or greater required')
```
SiyaoIsHiding, line 334: nit: as far as I know there isn't a branch called '5.0-beta' anymore. It's now called '5.0.0'.

SiyaoIsHiding (changes requested):
Apart from the type mismatch error handling, looks good to me!

cassandra/cqltypes.py
absurdfarce: This is actually happening (at least for deserialize) with the changes above.  I've also recently added support for similar tests in serliaze() (along with tests for all the cases).

tests/integration/__init__.py
absurdfarce: The intent here was to handle the various betas and release candidate releases as we run up to Cassandra 5.0.  Once we get to an actual 5.0 release this can be easily changed.

absurdfarce:
I believe everything from the last go-round has been addressed.  @SiyaoIsHiding would you mind taking another look?

SiyaoIsHiding:
Sorry that previously I missed `Counter` subtype. `Counter` is considered a fixed length type, even if `BigInt` is var-sized. We may want to change the `CounterColumnType` as follows:
```java
class CounterColumnType(LongType):
    typename = 'counter'

    @classmethod
    def serial_size(cls):
        return None
```
I don't quite understand how `vector<counter, 2>` works though. I don't know how we can test it.

SiyaoIsHiding (approved):

PR #1218, opened 2024-07-24
Add 5.0-beta1 to matrix

joao-r-reis:
Currently tests are failing and I think most (if not all) are related to cassandra.yml setting renaming that happened in C* 4.1. This is happening with other drivers as well. The Java driver already handled this in https://github.com/apache/cassandra-java-driver/pull/1924

C* 4.1 renamed some config settings but C* is still looking at the old names because of backwards compability reasons. However, C* fails to launch if it finds both the NEW name and the OLD name for a specific setting. The driver tests call ccm to add the OLD config setting but because the name is different from the one in cassadra.yml, the setting is added to cassandra.yml instead of replacing it.

Another issue that I found is that Jenkins is not  marking a step as "red"/failed if the tests fail. I don't know if this is related to the recent changes to the test infrastructure that happened on this driver.

I've already updated the AMI so it's already running the updated ccm that supports 5.0.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ import com.datastax.jenkins.drivers.python.Slack
 
 slack = new Slack()
 
-DEFAULT_CASSANDRA = ['3.0', '3.11', '4.0']
+DEFAULT_CASSANDRA = ['3.0', '3.11', '4.0', '5.0-beta1']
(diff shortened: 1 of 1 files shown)


absurdfarce:
The changes in this PR have been folded into the work being done for #1220

PR #1221, opened 2024-08-07
Upgraded cython to 3.0.x

bschoening:
Upgrade cython to 3.x which builds for Python 3.x by default.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -368,7 +368,7 @@ def run_setup(extensions):
         # 1.) build_ext eats errors at compile time, letting the install complete while producing useful feedback
         # 2.) there could be a case where the python environment has cython installed but the system doesn't have build tools
         if pre_build_check():
-            cython_dep = 'Cython>=0.20,!=0.25,<0.30'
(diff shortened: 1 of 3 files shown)


absurdfarce:
Cross-reference: assuming this works out this should address PYTHON-1390 as well.

absurdfarce (approved):

absurdfarce:
I'm going to merge this now in order to get this ball rolling.  I think we'll have some work to do based on some of the [changes that come in](https://cython.readthedocs.io/en/stable/src/userguide/migrating_to_cy30.html) with Cython 3.0.x, specifically around the NumPy integration.  Jenkins runs for this PR are showing something that looks very much like the problems described by PYTHON-1347 but in this case I _believe_ they're caused by the changes in the interaction between Cython 3.0.x and NumPy.  This needs to be investigated a bit more but it's not an argument for holding off on this upgrade; it seems pretty obvious that supporting Cython 3.0.x is the way forward for the Python driver.

PR #1161, opened 2023-05-23
PYTHON-1352 Add vector type, codec + support for parsing CQL type

absurdfarce:

PR #1222, opened 2024-08-15
Remove a leftover code

dkropachev:
It was part of https://github.com/datastax/python-driver/pull/1161.
Looks like some excessive debug code that was overlooked by @absurdfarce


absurdfarce:
Nice catch @dkropachev !  If I recall correctly this was added originally because the repr() op used in the conversion string for the ValueError in the next line was failing... and I was trying to get at the actual error string.  But I might be wrong about all of that. :)

Regardless, you're absolutely correct, this code doesn't need to be here now that repr() for vector types should work correctly.

PR #1223, opened 2024-08-22
PYTHON-1386 Make libev extension consistent with other extensions and allow specifying of libev install via env vars

absurdfarce:


absurdfarce:
Tests for this branch show no change (as expected), calling this good.

PR #1228, opened 2024-09-14
add convenience method for authenticating with astra via db_id and token

jbellis:
example usage:
```

cloud_config = {
    'db_id': '04ab96df-...',
    'token': 'AstraCS:...'
    }
cluster = Cluster(cloud=cloud_config)
```

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -2240,9 +2240,9 @@ class Cluster(object):
             str: The secure connect bundle URL for the given inputs.
 
         Raises:
-            URLError: If the request to the Astra DB API fails.
+            ValueError: If the Astra DB API response is missing the download url or the specified db_region is not found.
+            Exception: If there's an error connecting to the Astra API or processing the response.
         """
-        # set up the request
         url = f"https://api.astra.datastax.com/v2/databases/{db_id}/datacenters"
         headers = {
             "Authorization": f"Bearer {token}",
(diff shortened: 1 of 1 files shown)


jbellis:
If you have a good solution, I can give it a try. Otherwise I'm happy for
this to be a single region option.

absurdfarce:
We've had discussions about adding automatic download of the SCB into the drivers as an option in the past and in the end the decision has always been not to do so.  That API has changed underneath us in the past IIRC but far more importantly it makes the DevOps API outages production-level issues for _lots_ of clients... and DBPE has expressed some reservation about elevating that API to that status.

@joao-r-reis I know you and I have had these discussions in the past.  Did I misrepresent/get anything wrong in the description above?

jbellis:
We need some kind of solution here, the number of manual clicks required to
download the scb is way beyond "I have better things to do with my time"
for most devs. Especially now that the UI prioritizes data API so hard.

jbellis:
> That API has changed underneath us in the past IIRC

It's a public, documented API.  I think this is a non issue now. https://docs.datastax.com/en/astra-api-docs/_attachments/devops-api/index.html#tag/Database-Operations/operation/listDatacenters

phact:
> but far more importantly it makes the DevOps API outages production-level issues for lots of clients... 
would caching it on disk alleviate your concern @absurdfarce

absurdfarce:
It doesn't @phact.  The API stability concerns are definitely a second-tier issue.  The much larger issue for me is that we have spoken with ProdEng about this specific change and been told that we don't know that the relevant APIs/downloads were something that should be exposed for broad public use at the rates we'd expect from wide deployment of the drivers.

Even if _that_ were no longer the case I wouldn't want to address this as a one-off for a single driver.  I'd prefer to see a proposal to fix this handled as part of our product planning process so that we can deploy it uniformly to _all_ drivers in a way that keeps the interface reasonably consistent.  I'll raise this with our product folks so that they're aware of it.

msmygit:
> I also just realized we have this https://github.com/mieslep/cassandra-connector/blob/main/cassandra_connector/cassandra_connector.py#L38 which could be improvized outside of the python driver itself to just accept the DB ID (as opposed to the endpoint as is being currently available here).

absurdfarce:
We've discussed this internally amongst the driver developers and the decision was to close this one out.  I would note, however, that this functionality can be easily added already via cassio or some other small front-end into Astra (something in the vein of [gocql-astra](https://github.com/datastax/gocql-astra)); it doesn't need to be in the driver proper.  So something like that may be worth exploring as a story for users who want to explore this functionality.

Thanks all!

PR #1229, opened 2024-09-17
add execute_concurrent_async and expose execute_concurrent_* in Session

jbellis:
This makes these functions infinitely more discoverable for someone (correctly) looking at the Session class for ways to throw data into Cassandra


absurdfarce:
Apologies @jbellis, I know this has taken forever... I've been distracted by a lot of other things in the interim.  That said, I agree that elevating the visibility of this functionality (both the existing stuff plus the async impl here) by hanging it directly off of Session is probably a good idea.

Many thanks for the PR!

PR #1230, opened 2024-09-25
Test with DSE version 6.9.0

lukasz-antoniak:
`self._check_hostname` was never set in `connection.py` and `_validate_hostname()` never executed. I am also not sure about default implementation of `_validate_hostname()` but it looks to me that test `test_cannot_connect_ssl_context_with_invalid_hostname` succeeds only because of expired TLS certificates.


cassandra/connection.py
```diff
@@ -779,6 +779,7 @@ def __init__(self, host='127.0.0.1', port=9042, authenticator=None,
             self.ssl_options.update(self.endpoint.ssl_options or {})
         elif self.endpoint.ssl_options:
             self.ssl_options = self.endpoint.ssl_options
+        self._check_hostname = self.ssl_options.get('check_hostname', False)
```
absurdfarce, line 782: _check_hostname is no longer used for anything; see PYTHON-1331 for more details.  I kept the [logic](https://github.com/datastax/python-driver/blob/master/cassandra/connection.py#L935-L941) to call _validate_hostname() if this flag is set in order to allow custom connection subclasses to do additional validation if they wish but it's completely unnecessary.

cassandra/connection.py
```diff
@@ -906,7 +907,7 @@ def _initiate_connection(self, sockaddr):
     #
     # Allow implementations specific to an event loop to add additional behaviours
     def _validate_hostname(self):
-        pass
+        ssl.match_hostname(self._socket.getpeercert(), self.endpoint.address)
```
absurdfarce, line 910: We definitely do NOT want to do this.  ssl.match_hostname() is a deprecated function (and is actually removed in 3.12 IIRC but I don't remember that part for certain) that has been replaced by other functionality.  We leverage that other functionality by setting check_hostname on the ssl_context [here](https://github.com/datastax/python-driver/blob/master/cassandra/connection.py#L870).

PYTHON-1331 has A LOT more detail on this point.

PR #1231, opened 2024-09-25
Fix spelling issues in python code

bschoening:
This PR addresses spelling issues in code (an earlier PR addressed test code). PyCharm's spell checker was used to flag issues needed correction.  Given the number of misspellings identified, a conservative approach was taken to fix misspellings in comments and log messages.


absurdfarce:
Good point @bschoening ... I'll update the other ticket and close it out as well!

PR #1232, opened 2024-10-01
PYTHON-1331 Recreate expired SSL certificates for integration tests

lukasz-antoniak:


absurdfarce:
Note to self: would be good to address [this comment](https://github.com/apache/cassandra-python-driver/pull/1293#discussion_r3574517450) as well when we revisit this PR.

PR #1233, opened 2024-10-01
PYTHON-1402 Run integration tests with DSE 6.9.0

lukasz-antoniak:
SASI indexes have been removed in DSE 6.9.0.
Error message: `Column 'text' has a SASI index but SASI is not supported anymore. This index should removed and probably replaced by an equivalent Storage Attached Index (SAI)`

### DIFF
--- a/tests/integration/cqlengine/statements/test_base_statement.py
+++ b/tests/integration/cqlengine/statements/test_base_statement.py
@@ -25,7 +25,7 @@ from cassandra.cqlengine.columns import Column
 
 from tests.integration.cqlengine.base import BaseCassEngTestCase, TestQueryUpdateModel
 from tests.integration.cqlengine import DEFAULT_KEYSPACE
-from tests.integration import greaterthanorequalcass3_10, TestCluster
+from tests.integration import greaterthanorequalcass3_10, lessthandse69, TestCluster
 
 from cassandra.cqlengine.connection import execute
 
@@ -101,6 +101,7 @@ class ExecuteStatementTest(BaseCassEngTestCase):
         self.assertEqual(TestQueryUpdateModel.objects.count(), 0)
 
     @greaterthanorequalcass3_10
+    @lessthandse69
     def test_like_operator(self):
         """
         Test to verify the like operator works appropriately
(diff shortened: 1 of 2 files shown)


absurdfarce:
For sake of clarity: this PR makes changes necessary to get the test suite to pass on DSE 6.9.  The actual Jenkinsfile changes to enable _running_ tests against 6.9.x are in #1234.

tests/integration/cqlengine/statements/test_base_statement.py
```diff
@@ -101,6 +101,7 @@ def test_insert_statement_execute(self):
         self.assertEqual(TestQueryUpdateModel.objects.count(), 0)
 
     @greaterthanorequalcass3_10
+    @lessthandse69
```
absurdfarce, line 104: Wait, why does this test... fail on DSE 6.9?  Did we remove support for LIKE or something?

absurdfarce: Oh, it's because of the removal of SASI.. I see.

Can we replace this with a new SAI test then?

lukasz-antoniak: Unfortunately not the `LIKE` operator.

> SAI does not supports the following query operators for tables with SAI indexes:
> - Strings or Numerics: LIKE, OR, IN

Documentation [link](https://docs.datastax.com/en/cql/dse-6.9/develop/indexing/sai/sai-query.html).

absurdfarce: The [CEP](https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-7%3A+Storage+Attached+Index) has it as a "Phase 2" thing which puts it in the same category as OR queries which were discussed at some length for the Java driver (see [CASSANDRA-19931](https://issues.apache.org/jira/browse/CASSANDRA-19931)).  Excluding for 6.9.0 and up seems like the right answer until we get a better answer from SAI.

absurdfarce (approved):

Dianne087:
Area()

PR #1234, opened 2024-10-01
PYTHON-1402 Support running test suite with HCD 1.0.0

lukasz-antoniak:


absurdfarce:
Kicked off a Jenkins build of this to confirm we're okay, will post updates here when that completes.

Jenkinsfile
```diff
@@ -50,6 +52,11 @@ matrices = [
     "RUNTIME": DEFAULT_RUNTIME,
     "CYTHON": DEFAULT_CYTHON
   ],
+  "HCD": [
+    "SERVER": DEFAULT_HCD,
+    "RUNTIME": DEFAULT_RUNTIME,
+    "CYTHON": DEFAULT_CYTHON
+  ],
```
absurdfarce, line 59: Discussed last week but the [default matrix](https://github.com/datastax/python-driver/blob/master/Jenkinsfile#L86) used for most test builds is the "SMOKE" matrix... and it seems to me we'd prefer to bring HCD into that matrix rather than introducing a new matrix for HCD (which would have to be explicitly triggered).  To that end this PR should:

* Remove the "HCD" matrix above
* Either add HCD versions to the DEFAULT_DSE array or keep them in the DEFAULT_HCD array
* Update the server selection logic below (or [here](https://github.com/datastax/python-driver/blob/master/Jenkinsfile#L54)) to include HCD versions in some way.

There is some subtlety here.  We're already pulling the most recent DSE version for smoke tests via the `DEFAULT_DSE.takeRight(1)` call.  What we probably want to wind up with is smoke tests running against the current DSE minor (6.9.x), the previous DSE minor (6.8.x) and HCD.  You can make this happen whether you keep a separate array for HCD or fold i into the DEFAULT_DSE array.

Keep in mind that every new server backend adds 2 x (number of Python versions) builds to the Jenkins runs... so we want to be judicious about what we add here.

Jenkinsfile
```diff
@@ -228,6 +235,21 @@ ENVIRONMENT_EOF
       '''
   }
 
+  if (env.CASSANDRA_VERSION.split('-')[0] == 'hcd') {
```
absurdfarce, line 238: Presumably this should be an "else if" in order to avoid some weird case in which both of these if blocks execute.  There's no obvious way for that to happen but in the worst case it still clearly communicates intent.

tests/integration/standard/test_query.py
```diff
@@ -942,7 +942,7 @@ def test_no_connection_refused_on_timeout(self):
                 exception_type = type(result).__name__
                 if exception_type == "NoHostAvailable":
                     self.fail("PYTHON-91: Disconnected from Cassandra: %s" % result.message)
-                if exception_type in ["WriteTimeout", "WriteFailure", "ReadTimeout", "ReadFailure", "ErrorMessageSub"]:
+                if exception_type in ["WriteTimeout", "WriteFailure", "ReadTimeout", "ReadFailure", "ErrorMessageSub", "ErrorMessage"]:
```
absurdfarce, line 945: Do we know why we're suddenly seeing the parent class `ErrorMessage` here when we weren't before?

absurdfarce (changes requested):
Looks quite good!  Think we still need a bit more work re: how we include HCD versions into the existing set of matrices but that's the only thing we _need_ to do at this point.

Jenkinsfile
lukasz-antoniak: Very good catch. Fixed.

tests/integration/standard/test_query.py
lukasz-antoniak: Driver receives exceptions:
```
WriteTimeout('Error from server: code=1100 [Coordinator node timed out waiting for replica nodes\' responses] message="CAS operation timed out - encountered contentions: 18" info={\'consistency\': \'SERIAL\', \'required_responses\': 2, \'received_responses\': 0, \'write_type\': \'CAS\'}')

ErrorMessage: <Error from server: code=1700 [Unknown] message="CAS operation result is unknown - proposal accepted by 1 but not a quorum.">
```

It looks clear [here](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/transport/messages/ErrorMessage.java#L192) why `CAS_WRITE_UNKNOWN` is wrapped in `ErrorMessage `. Cannot answer why this was not the case before.

I have asked HCD engineers around, but they are not familiar with LWT changes introduced in C* 5.0.

absurdfarce: Thanks for checking into it.  For the record I don't think this difference should hold up getting this PR in.  It seemed like an odd change when looking this over but I can envision perfectly good reasons why it might be the case.

Jenkinsfile
```diff
@@ -51,7 +53,7 @@ matrices = [
     "CYTHON": DEFAULT_CYTHON
   ],
   "SMOKE": [
-    "SERVER": DEFAULT_CASSANDRA.takeRight(2) + DEFAULT_DSE.takeRight(1),
+    "SERVER": DEFAULT_CASSANDRA.takeRight(2) + DEFAULT_DSE.takeRight(2) + DEFAULT_HCD.takeRight(1),
```
absurdfarce, line 56: For the record: I had this wrong in my earlier comment.  We should actually be adding only four new builds to our smoke test build matrix with this change: 2 new backends * 2 Python versions (oldest and newest).  That should be reasonably manageable.  More to the point it's hard to imagine how we'd remove either 6.8.x, 6.9.x or HCD from our test matrix.

absurdfarce (approved):

Dianne087:
python return

< AnnotationElenments> Interdace, functionality

PR #1236, opened 2024-11-02
Update geomet to align with requirements.txt

bschoening:
also fixes some pycodestyle issues

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ class DocCommand(Command):
 
             try:
                 os.makedirs(path)
-            except:
+            except OSError:
                 pass
 
         if has_subprocess:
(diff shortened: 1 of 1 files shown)


bschoening:
@absurdfarce just a small tweak, but needed to update the Homebrew formula for Cassandra's dependencies.

absurdfarce:
Thanks for the ping @bschoening.  I was looking at this some last night but didn't actually get it finished.

Looking over this ticket also reminded me that we need to handle the removal of distutils in 3.12.  My guess is that [PYTHON-1334](https://datastax-oss.atlassian.net/browse/PYTHON-1334) (the migration to pyproject.toml) will handle that as a side-effect but if setup.py is around in some capacity after that we need to fix it.  I've filed [PYTHON-1421](https://datastax-oss.atlassian.net/browse/PYTHON-1421) to handle that.

absurdfarce:
Thanks (once again) @bschoening!  That was a nice catch on the disconnect between requirements.txt and setup.py... yuck.

absurdfarce:
Hey @eladkal, we don't have a scheduled date yet.  We're trying to get several other driver releases out this quarter so it certainly won't be this month.  That said, I'd like to see if we can get 3.30.0 out in the next quarter.. so my guess is sometime late summer.

That's very much contingent on being able to get the asyncio reactor to production quality, though, so that's certainly not a guarantee.

absurdfarce:
@eladkal A quick update: 3.29.3 was just [released today](https://pypi.org/project/cassandra-driver/3.29.3/) and it does include this change

PR #1237, opened 2024-11-07
Update README.rst to add python version badge

bschoening:


absurdfarce:
Thanks @bschoening!  I'm always super-terrible at this kind of thing so I'm glad you're around to make this work in a useful way! :)

absurdfarce (approved):

PR #1239, opened 2024-12-02
Add method to asynchronously prepare CQL statements

lukasz-antoniak:


lukasz-antoniak:
I have implemented a small fix to preserve synchronous behaviour of prepare logic in the `prepare()` call.

cassandra/cluster.py
```diff
@@ -3118,36 +3152,27 @@ def prepare(self, query, custom_payload=None, keyspace=None):
...
         if self.cluster.prepare_on_all_hosts:
+            # prepare on all hosts in a synchronous way, not asynchronously
+            # as internally in prepare_async() (PrepareFuture)
             host = future._current_host
             try:
-                self.prepare_on_all_hosts(prepared_statement.query_string, host, prepared_keyspace)
+                self.prepare_on_all_nodes(response.query_string, host, response.keyspace)
             except Exception:
                 log.exception("Error preparing query on all hosts:")
+        return response
```
absurdfarce, line 3173: This seems exactly right: prepare() should be implemented as prepare_async().get() once we have a good working prepare_async().  But why do we need to do the prepare_on_all_nodes() operation here?  We should've already done that when future.result() completed (since it's done in PrepareFuture._set_final_result()).

cassandra/cluster.py
```diff
@@ -5105,6 +5130,49 @@ def __str__(self):
...
+        prepared_statement.custom_payload = response.custom_payload
+        cluster.add_prepared(response.query_id, prepared_statement)
+        self._final_prepare_result = prepared_statement
+
+        if self._prepare_on_all_hosts:
+            # trigger asynchronous preparation of query on other C* nodes,
+            # we are on event loop thread, so do not execute those synchronously
+            session.submit(
+                session.prepare_on_all_nodes,
+                self.query_string, self._current_host, self._keyspace)
```
absurdfarce, line 5159: It seems strange to me to have this logic embedded within the future impl like this.  Seems like this should be done when the future is created, something like:

```
# _create_prepare_response_future() in this impl returns a regular ResponseFuture
future = self._create_prepare_response_future(query, keyspace, custom_payload, prepare_on_all_hosts)
if prepare_on_all_hosts:
   # Hand waving about partial application here in order to pass in parameters; the point is we get to a future that
   # calls prepare_on_all_hosts() after we've received a response to our prepare here here
   future = future.then(prepare_on_all_hosts)
future._protocol_handler = self.client_protocol_handler
```

Problem of course is the ResponseFuture doesn't support this then() syntax.  It does have native support for callbacks but that isn't the same; that's just a function that gets invoked when the operation completes.  We don't return a new future that returns the result of the function defined in the then() call like you do in most future APIs.

I'm wondering if we can either (a) add something like that or (b) find a way to wrap this functionality in another future lib in order to simplify an impl like this.

absurdfarce (changes requested):
I think it's worth our while trying to figure out a way to implement this using something like a Future.then() syntax to return a future which does the prepare and then calls prepare_on_all_hosts() when that's successful.  This will take some more plumbing to support this use case but it's a _much_ simpler implementation _and_ such an impl maps naturally onto this use case.

lukasz-antoniak:
I totally agree with your comments. Previously I could not find another way to reuse `ResponseFuture` and return `PreparedStatement` object. Can you check current state of PR?

PR #1240, opened 2025-02-14
Remove duplicated condition in primary key check

daniloarodrigues:
Motivation:
We had a redundant check (`col.primary_key or col.primary_key`) that caused confusion and had no logical effect.

Modifications:
Removed the duplicated check, leaving only a single condition for `col.primary_key`.

Result:
The code is now clearer and avoids unnecessary duplication.

### DIFF
--- a/cassandra/cqlengine/management.py
+++ b/cassandra/cqlengine/management.py
@@ -256,7 +256,7 @@ def _sync_table(model, connection=None):
 
                 continue
 
-            if col.primary_key or col.primary_key:
+            if col.primary_key:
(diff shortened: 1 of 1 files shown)


daniloarodrigues:
> Thanks for the PR @daniloarodrigues! That is indeed a pretty awesome find... 🤦
> 

I've just signed up. Thanks for letting me know.

absurdfarce:
Just kicked off a full CI run on this guy just to make absolutely sure there's no regression.  I can't imagine how there could be but... I've been surprised before.

I've also looked at the referenced code like ten times to make sure those things in fact say the same thing.

Assuming there's nothing unexpected with the build (and I'm not expecting anything) this will go straight in.

PR #1242, opened 2025-03-27
Maintain compatibility with CPython 3.13

daskol:
See changes in Python C API in [CPython 3.13 changelog][1].

[1]: https://docs.python.org/3/whatsnew/3.13.html

### DIFF
--- a/cassandra/io/libevwrapper.c
+++ b/cassandra/io/libevwrapper.c
@@ -665,9 +665,14 @@ initlibevwrapper(void)
     if (PyModule_AddObject(module, "Timer", (PyObject *)&libevwrapper_TimerType) == -1)
         INITERROR;
 
+#if PY_MAJOR_VERSION < 3 && PY_MINOR_VERSION < 7
+    // Since CPython 3.7, `Py_Initialize()` routing always initializes GIL.
+    // Routine `PyEval_ThreadsInitialized()` has been deprecated in CPython 3.7
+    // and completely removed in CPython 3.13.
     if (!PyEval_ThreadsInitialized()) {
         PyEval_InitThreads();
     }
+#endif
 
 #if PY_MAJOR_VERSION >= 3
     return module;


medaminezghal:
@daskol You need to edit the files that use asyncore (Remove in python 3.13) to make it use asyncio
This what I did in the prepare function in PKGBUILD file:
```
sed -i 's/_try_asyncore_import/_try_asyncio_import/g' cassandra/cluster.py
    sed -i 's/from cassandra.io.asyncorereactor import AsyncoreConnection/from cassandra.io.asyncioreactor import AsyncioConnection/g' cassandra/cluster.py
    sed -i 's/from cassandra.io.asyncorereactor import AsyncoreConnection/from cassandra.io.asyncioreactor import AsyncioConnection/g' tests/integration/long/test_ipv6.py
    sed -i 's/(AsyncoreConnection,None)/(AsyncioConnection,None)/g' cassandra/cluster.py
    sed -i '/import asynctest/d' tests/unit/io/test_asyncioreactor.py
    sed -i 's/asynctest.TestSelector()/AsyncMock()/g' tests/unit/io/test_asyncioreactor.py
    sed -i '/from unittest.mock import patch/a from unittest.mock import AsyncMock' tests/unit/io/test_asyncioreactor.py
```
Maybe be you could edit the files to use asyncore or asyncio by case (python>=3.13 or not)

medaminezghal:
@daskol and even after I edit for some reason I can't import the cassandra.io.libevwrapper

daskol:
@medaminezghal What about `cqlsh` utility? It does not work without this fix because of missing `cp313` wheels.

medaminezghal:
@daskol if you build the package without the C extensions it should work (of course after editing the files that use asyncore) and then you can build cqlsh in python 3.13

medaminezghal:
@daskol https://docs.python.org/3/library/asyncore.html

daskol:
Do `pacman -S python-pyasyncore` or `pip install pyasyncore` solve your issue? Initially, I faced with the issue of missing cp313 wheels and build failure of this dreiver from scratch.

medaminezghal:
I think it’s not useful to use unmaintained package as provided in the description. So it’s better to use asyncio

medaminezghal:
@daskol I think you need to fix the problem related to [this issue from other project](https://github.com/openwallet-foundation/acapy/issues/1717).
There is only [1 file](https://github.com/datastax/python-driver/blob/master/tests/unit/io/test_asyncioreactor.py) that use asynctest package.
Also the distutils (used to build the C extensions) is deprecated and will be removed soon.

absurdfarce:
Thanks for the PR @daskol!

I mentioned this in another recently submitted PR but it's worth restating here: because the Python driver aims to support all Python runtimes that aren't end-of-life we still have to support Python 3.9.x and up.  And since the asyncore reactor is more stable than the current asyncio version I have no plans to remove it anytime soon.  There's a lot more detail in [my comment](https://github.com/datastax/python-driver/pull/1244#issuecomment-2917674322).

For the record: if you want to use the libev reactor you'll need libev installed and accessible via the lib path before the Python driver will even attempt to use it.  You can find some additional info in our [docs](https://github.com/datastax/python-driver/blob/master/docs/installation.rst#libev-support) on installing the Python driver.

bschoening:
@daskol since the driver only supports 3.8+, is there any reason the deprecated initializer code cannot be simply removed entirely?

absurdfarce:
Hey @andy-slac, thanks for the clarification!  Yeah, you're seeing those errors because libev itself isn't available on your box.  The libevwrapper code referenced in this PR provides interop between the driver and libev but we still require the library itself.  This is mentioned in the installation docs, specifically the [section on libev](https://docs.datastax.com/en/developer/python-driver/3.29/installation/index.html#libev-support).

For what it's worth part of the motivation for my earlier comments was the mess around asyncio support in Python 3.12 and up.  There's been a lot of confusion around that point and we've indicated to a number of folks that libev is the recommended reactor for Python 3.12+ (until we can get to a prod-ready asyncio reactor).  So I didn't want users to see a suggestion that libev wasn't functionality on Python 3.13 without providing the additional context.

absurdfarce:
@andy-slac I'm not 💯 sure what the root cause might be in your case but a couple things do occur to me.

First, you need to have both libev and it's corresponding dev package installed; without the dev package you won't get the headers necessary to build everything.

Second, the directories where libev is installed on your system might not be on the default search path.  setup.py will search a set of default paths to try and find libev (details [here](https://github.com/datastax/python-driver/blob/3.29.2/setup.py#L127-L138)) and at least the paths you mention in your comment are different from those defaults.  You can always try defining CASS_DRIVER_LIBEV_INCLUDES and CASS_DRIVER_LIBEV_LIBS to point to the correct directories on your system to see if that helps.

absurdfarce:
Jenkins run has finished and looks good, saw improvement in the expected area around the libev extension build on Python 3.13.  No other errors were introduced.

andy-slac:
@absurdfarce, in my case everything was installed properly, but wrapper extension failed to build because gcc generated an error for an implicit declaration of PyEval_ThreadsInitialized. I think this is a correct behavior, implicit declarations should not be allowed in modern compiler. Here is the log from pip install:
```
2025-10-01T22:04:21,136   gcc -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O3 -Wall -fPIC -I/usr/include/libev -I/usr/local/include -I/opt/local/include -I/usr/include -I/usr/local/include/python3.13 -c cassandra/io/libevwrapper.c -o build/temp.linux-x86_64-cpython-313/cassandra/io/libevwrapper.o
2025-10-01T22:04:21,186   cassandra/io/libevwrapper.c: In function <E2><80><98>PyInit_libevwrapper<E2><80><99>:
2025-10-01T22:04:21,186   cassandra/io/libevwrapper.c:668:10: error: implicit declaration of function <E2><80><98>PyEval_ThreadsInitialized<E2><80><99> [-Wimplicit-function-declaration]
2025-10-01T22:04:21,186     668 |     if (!PyEval_ThreadsInitialized()) {
2025-10-01T22:04:21,186         |          ^~~~~~~~~~~~~~~~~~~~~~~~~
2025-10-01T22:04:21,186   cassandra/io/libevwrapper.c:669:9: warning: <E2><80><98>PyEval_InitThreads<E2><80><99> is deprecated [-Wdeprecated-declarations]
2025-10-01T22:04:21,186     669 |         PyEval_InitThreads();
2025-10-01T22:04:21,186         |         ^~~~~~~~~~~~~~~~~~
2025-10-01T22:04:21,186   In file included from /usr/local/include/python3.13/Python.h:125,
2025-10-01T22:04:21,186                    from cassandra/io/libevwrapper.c:1:
2025-10-01T22:04:21,186   /usr/local/include/python3.13/ceval.h:114:37: note: declared here
2025-10-01T22:04:21,186     114 | Py_DEPRECATED(3.9) PyAPI_FUNC(void) PyEval_InitThreads(void);
2025-10-01T22:04:21,186         |                                     ^~~~~~~~~~~~~~~~~~
2025-10-01T22:04:21,188   command '/usr/bin/gcc' failed with exit code 1
```

Thanks for merging the patch, when should we expect a new release?

bschoening:
`#if PY_MAJOR_VERSION < 3 && PY_MINOR_VERSION < 7`

Should this be **PY_MAJOR_VERSION == 3**? Isn't this checking for **version < 2.7**?

absurdfarce:
Great catch @bschoening... you're absolutely right.  I'm angry at myself for missing that!

Changing that to:

```
#if PY_MAJOR_VERSION < 3 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION < 7)
```

in a ninja fix.  I'll then run the full test suite on it just to make sure it's good.

absurdfarce:
Jenkins run is clean, calling this good

andy-slac:
@absurdfarce, great, thanks for the news!

PR #1244, opened 2025-05-23
CPython 3.13 Compat and Asyncore Removal

kheina:
Per [Python3.13 Release Notes](https://docs.python.org/3/whatsnew/3.13.html):
> Remove the `PyEval_ThreadsInitialized()` function, deprecated in Python 3.9. Since Python 3.7, `Py_Initialize()` always creates the GIL: calling `PyEval_InitThreads()` does nothing and `PyEval_ThreadsInitialized()` always returns non-zero. (Contributed by Victor Stinner in [gh-105182](https://github.com/python/cpython/issues/105182).)

Removes the use of `PyEval_ThreadsInitialized()` and `PyEval_InitThreads()` if python version is >= 3.7 (written by @daskol in https://github.com/datastax/python-driver/pull/1242/commits/27969fc0c6e5407602c776270bcdcce1ec127b30)

Additionally, removes existing references to `asyncore` in cassandra/cluster.py tests/integration/long/test_ipv6.py tests/unit/io/test_asyncioreactor.py (per @medaminezghal in https://github.com/datastax/python-driver/pull/1242#issuecomment-2760397717

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -148,10 +148,10 @@ def _try_libev_import():
     except DependencyException as e:
         return (None, e)
 
-def _try_asyncore_import():
+def _try_asyncio_import():
     try:
-        from cassandra.io.asyncorereactor import AsyncoreConnection
-        return (AsyncoreConnection,None)
+        from cassandra.io.asyncioreactor import AsyncioConnection
+        return (AsyncioConnection,None)
     except DependencyException as e:
         return (None, e)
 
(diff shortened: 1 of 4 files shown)


absurdfarce:
Hey @kheina, it looks like the thread initialization changes in this PR are replicated in #1242 and the asyncore-to-asyncio changes aren't something I'd like to do (for reasons discussed above).  As a result I'm going to close this PR out; we'll continue the discussion around the thread initialization logic in #1242.  Please join in there!

And thanks again for putting this PR together!

PR #1248, opened 2025-06-23
Remove obsolete urllib2 from ez_setup.py

bschoening:
In Python 2, you would import using import urllib2. In Python 3, you use import urllib.request.


absurdfarce:
Good catch @bschoening.  There's no obvious reason to continue with urllib2 so I'm 💯 in agreement here.

PR #1268, opened 2026-02-09
fix: Remove ez_setup for compatibility with setuptools v82

Dev-iL:
## Summary

Remove the legacy `ez_setup.py` bootstrap script to fix build failures with setuptools >= 82.

- **Delete `ez_setup.py`**: This 249-line script was a setuptools bootstrapper from circa 2013 that attempted to download setuptools 0.9.6 if not present. It imported `pkg_resources`, which was [removed in setuptools v82](https://setuptools.pypa.io/en/latest/history.html#v82-0-0) (2026-02-08), causing `ModuleNotFoundError` during builds.
- **Remove `ez_setup.py` from `MANIFEST.in`**: Stop including the deleted file in source distributions.

The script was already dead code — `setup.py` does not import or call it. The project's `pyproject.toml` (added in CASSPYTHON-3) declares `setuptools` as a build dependency, so any PEP 517-compliant build frontend (pip, uv, build) installs setuptools automatically, making `ez_setup.py` unnecessary.

### Build failure before this change

```
$ uv pip install cassandra-driver
  × Failed to build `cassandra-driver==3.29.3`

        File ".../ez_setup.py", line 115, in use_setuptools
          import pkg_resources
      ModuleNotFoundError: No module named 'pkg_resources'
```

## Test plan

- [x] `uv pip install -e .` succeeds and compiles all C/Cython extensions
- [x] `python -c 'import cassandra; print(cassandra.__version__)'` prints `3.30.0`

Generated-by: Claude Opus 4.6


potiuk (approved):

gopidesupavan (approved):

xrmx (approved):
This is breaking opentelemetry-python-contrib CI, a release with this would be appreciated.

tommystendahl (approved):

tommystendahl:
I also noticed this problem in our CI fro the ericsson/ecchronos project, if a quick release of this is possible it would be very much appreciated.

absurdfarce:
Oh Claude, once again your directions confuse me.  The build failure snippet above references 3.29.3 suggesting that the issue applies to installs of a version already deployed on pypi while your test plan references 3.30.0 suggesting that it only applies to what's in git. 🤦 

I've confirmed that package installs are only failing on Python 3.14.  The following sequence works without issue on everything exception that version:

```
$ uv venv --python 3.14 test-venv
$ . ./test-venv/bin/activate
$ uv pip install --upgrade setuptools # confirm this gives you 82.0.0
$ uv pip install cassandra-driver
```

Presumably 3.14 fails because we don't have wheels for 3.14 so you're forced to actually do a source build on package install.  It's relevant here that 3.29.3 doesn't include 3.14 as a supported platform.

sdist builds look like they might be a different story.  Still doing some digging on that.

potiuk:
> I've confirmed that package installs are only failing on Python 3.14. The following sequence works without issue on everything exception that version:

Yeah. It's not a **huge** problem - in our case it also failed for 3.13 when we installed 3.29.1  - see for example https://github.com/apache/airflow/actions/runs/21804090800/job/62904899230#step:8:1365  - mostly because there were no 3.13 wheels in this version (and it did not support it "officially" - yet we we have >=3.29.1 as minimum specified as airlfow dependencies. 

The test where it failed was the "lowest dependency" test - so the test when we attempt to bring all dependencies to their minimum specified versions and we want to check that our tests are still passing with them. So even if 3.29.1 did not support 3.13 officially, we got it working there because we built it from sdist and ... it worked.

I guess other people might have similar issues when they are installing older versions.

Sadly - indeed - there is not much you can do for already released versions.  They are immutable and they are what they are -  and if people  will want to use sdist to build "unsupported" version for 3.13 with sdist - they are able to do it and it will work as long as they apply workarounds.

So, this is true that there is no urgency here, some CI failures will happen in edge cases here, most installations will work with binary wheels and there is already a workaround - especially if you use modern tool versions - like 25.3+ of `pip` that supports --build-constraints, people can specify constraints to not install newer setuptools version.

It would however be great to get it fixed for the next release :)

potiuk:
I updated Airflow provider spec to better support version/whl availability https://github.com/apache/airflow/pull/61685

xrmx:
On top of the pkg_resources requirement this script is broken anyway:

```
      "/home/runner/.cache/uv/sdists-v9/pypi/cassandra-driver/3.29.3/8B-7iB8YnvcAwKBrY0GDo/src/ez_setup.py",
      line 203, in _extractall
          self.chown(tarinfo, dirpath)
      TypeError: chown() missing 1 required positional argument:
      'numeric_owner'
```

aarena-domotz:
Hello, we see this failing when installing cassandra driver with poetry (python 3.11, poetry 2.3.0, poetry-core 2.3.0)

absurdfarce:
To update progress here: we have a [PR](https://github.com/apache/cassandra-python-driver/pull/1271) to fix Jenkins to work with the pyproject.toml changes in CASSPYTHON-3.  Once that's in I'd like to get that merged into this PR so that we can do a CI run with this change before we merge it in.

From my perspective this fix _has_ to go in to 3.30.0 so @Dev-iL I'll let ya know when the other PR is in and you can merge trunk into your feature branch.

absurdfarce:
Okay, PR to fix Jenkins builds to work with pyproject.toml has been merged. @Dev-iL you should be good to merge trunk into this PR whenever you're ready.  Once that's done I'll run a test Jenkins build on this branch to ensure we don't see any regressions; I certainly don't expect any but, you know, better safe than sorry and all that. :)

Dev-iL:
@absurdfarce Done

absurdfarce:
Thanks @Dev-iL!

As expected the Jenkins build showed no regressions so I think we're good to go here.  Thanks to @Dev-iL for driving this one home!

absurdfarce:
@bschoening can you give me an additional 👍 on this one?  I thought about just doing it as a 🥷 but we are actually deleting content here... so I kinda want a second committer to confirm this one.  I wouldn't expect any problems on modern systems from this change and the Jenkins run was clean so we _should_ be good here... I just wanna make extra-special-sure.

Please and thank you. :)

bschoening (approved):

PR #1269, opened 2026-02-10
CASSPYTHON-4 Bump Python versions to match what's being deployed on the runner

absurdfarce:
Update Python versions to match what will be deployed on the Jenkins runner.  We're also bumping the underlying platform from Ubuntu Focal to Jammy so let's update the corresponding label while we're here.


absurdfarce:
I'm going to close this one out; there are _a lot_ of changes in-flight to the Jenkinsfile in order to support builds after the pyproject.toml migration so I want to wrap those up before we do anything with these changes.

absurdfarce:
Replaced by #1272

PR #1270, opened 2026-02-18
Replace usage of `with await lock` (`asyncio` reactor broken in Python 3.9+)

ramikg:
Trying to connect to Cassandra using the `asyncio` reactor times out in Python 3.9+.
This is because pushing messages to the queue silently fails due to the usage of the `with await asyncio.Lock` syntax, removed in Python 3.9.

I've replaced it with `async with lock` (as suggested by the [docs](https://docs.python.org/3/library/asyncio-sync.html#:~:text=Use%20async%20with%20lock%20instead)), and now it works.

### DIFF
--- a/cassandra/io/asyncioreactor.py
+++ b/cassandra/io/asyncioreactor.py
@@ -173,7 +173,7 @@ class AsyncioConnection(Connection):
 
     async def _push_msg(self, chunks):
         # This lock ensures all chunks of a message are sequential in the Queue
-        with await self._write_queue_lock:
+        async with self._write_queue_lock:
             for chunk in chunks:
                 self._write_queue.put_nowait(chunk)
 


absurdfarce:
Yup, this looks right to me.  Syntax was changed in Python 3.9 so all the versions we're planning on covering for 3.30.0 should be fine with this change.  I'm going to kick off a CI run to confirm there aren't any other unexpected regressions here but I think this should work just as it sits.

Nice find @ramikg!

absurdfarce:
Jenkins test run looks clean.  @bschoening if I can get a 👍 from you I believe we're good to merge this guy.

absurdfarce (approved):

bschoening (approved):

bschoening:
@absurdfarce @ramikg looks good and copilot approves!

ramikg:
Thank you both!

PR #1271, opened 2026-02-19
Update Jenkins build to work with pyproject.toml

absurdfarce:


fix-jenkinsfile-libev.py
```diff
@@ -0,0 +1,9 @@
+import sys
+import toml
+
+pyproject = toml.load(sys.argv[1])
+base = pyproject["tool"]["cassandra-driver"]
+base["libev-includes"] = [sys.argv[2]]
+base["libev-libs"] = [sys.argv[3]]
+
+print(toml.dumps(pyproject))
```
absurdfarce, line 9: Helper script to update pyproject.toml to point at the correct location for libev includes/libs.  Called in Jenkinsfile below.

Jenkinsfile
```diff
@@ -34,7 +34,7 @@ slack = new Slack()
 DEFAULT_CASSANDRA = ['3.11', '4.0', '4.1', '5.0']
 DEFAULT_DSE = ['dse-5.1.35', 'dse-6.8.30', 'dse-6.9.0']
 DEFAULT_HCD = ['hcd-1.0.0']
-DEFAULT_RUNTIME = ['3.9.23', '3.10.18', '3.11.13', '3.12.11', '3.13.5']
+DEFAULT_RUNTIME = ['3.10.18', '3.11.13', '3.12.11', '3.13.5']
```
absurdfarce, line 37: Bumped minimum version to 3.10 since we won't be officially supporting 3.9 with Python driver 3.30.0.  We'll need to add 3.14 separately but I need to get that version installed on the Jenkins runners first.

Jenkinsfile
```diff
@@ -165,15 +165,18 @@ def getMatrixBuilds(buildContext) {
...
+
+    pyenv shell ${PYTHON_VERSION}
+    python -m venv jenkins-venv
+    . ./jenkins-venv/bin/activate
+    pip install --upgrade pip setuptools wheel
+
+    # Install a version of pyyaml<6.0 compatible with ccm-3.1.5 as of Aug 2023
     # this works around the python-3.10+ compatibility problem as described in DSP-23524
-    pip install wheel
-    pip install "Cython<3.0" "pyyaml<6.0" --no-build-isolation
```
absurdfarce, line 176: Cython is now handled by the build backend so there's no need to install it in the venv here

Jenkinsfile
```diff
@@ -244,39 +240,65 @@ ENVIRONMENT_EOF
...
-    '''
-  }
+def installDriver() {
+  sh label: 'Install the driver and compile with C extensions with Cython', script: '''#!/bin/bash -lex
+    # Update libev includes and libs to point to the right spot for this install
+    pyenv shell ${PYTHON_VERSION}
+    python -m venv libev-venv
+    . ./libev-venv/bin/activate
+    pip install toml
+    python fix-jenkinsfile-libev.py ./pyproject.toml "/usr/include" "/usr/lib/x86_64-linux-gnu" | sponge ./pyproject.toml
```
absurdfarce, line 263: The heart of the matter.  These paths match the install location for the package installs used on the Jenkins runners.

Note that this is done in a separate venv to avoid contaminating the actual build venv with unrelated packages.

Jenkinsfile
```diff
@@ -244,39 +240,65 @@ ENVIRONMENT_EOF
...
+    # Now that we've made relevant mods to our local pyproject.toml we're ready to build the driver
+    . ./jenkins-venv/bin/activate
+
+    # Load CCM environment variables
+    set -o allexport
+    . ${HOME}/environment.txt
+    set +o allexport
+
+    cat ./pyproject.toml
+    pip install --verbose --editable .
```
absurdfarce, line 278: We need a local install of the built driver; without this pytest will discover (and apparently prefer) the local versions of the corresponding Python packages so unless the built native libs are also available in the same directory structure we'll get lots of complaints about libev being unavailable.

Jenkinsfile
```diff
@@ -244,39 +240,65 @@ ENVIRONMENT_EOF
...
       # Load CCM environment variables
       set -o allexport
       . ${HOME}/environment.txt
       set +o allexport
 
+      LIBEV_MOD_PATH=`find . -name "libevwrapper.cpython*"`
+      ldd $LIBEV_MOD_PATH
+
+      python -c 'import cassandra.io.libevwrapper'
+
```
absurdfarce: Some leftover debug code... this can probably be removed.

pyproject.toml
```diff
@@ -51,6 +51,6 @@ include = ['cassandra', 'cassandra.io', 'cassandra.cqlengine', 'cassandra.graph'
 build-murmur3-extension = true
 build-libev-extension = true
 build-cython-extensions = true
-libev-includes = []
-libev-libs = []
+libev-includes = ["/usr/include/libev", "/usr/local/include", "/opt/local/include", "/usr/include"]
+libev-libs = ["/usr/local/lib", "/opt/local/lib", "/usr/lib64"]
```
absurdfarce, line 55: We used to have these defaults embedded in setup.py which we would apply when the appropriate conditions were met.  In the spirit of making the build more explicit it seemed better to just put these values directly in pyproject.toml as defaults.

absurdfarce (commented):
Mainly explanation

absurdfarce:
Hey @bschoening I'm guessing you're not super-concerned about the Jenkinsfile changes but I would like your feedback on the pyproject.toml/setup.py changes embedded here.

absurdfarce:
Ping @millerjp for visibility

Jenkinsfile
```diff
@@ -165,15 +165,18 @@ def getMatrixBuilds(buildContext) {
...
-    pip install --upgrade pip
-    pip install -U setuptools
 
-    # install a version of pyyaml<6.0 compatible with ccm-3.1.5 as of Aug 2023
+    # One of the integration tests relies on socat so let's install that here
+    sudo apt-get install -y socat moreutils
+
+    pyenv shell ${PYTHON_VERSION}
+    python -m venv jenkins-venv
+    . ./jenkins-venv/bin/activate
```
absurdfarce, line 174: I moved everything in to a contained venv because we were seeing weird behaviours in which pyenv wasn't necessarily updating the local Python instance which in turn led to packages being applied inconsistently.  To avoid all of that it seemed worthwhile to be very explicit about the idea of a container for all necessary Python dependencies... which is pretty much a venv.

Jenkinsfile
Dev-iL: Why not install uv then use that to install 3.14 + dependencies, all as part of the jenkins flow?

setup.py
```diff
@@ -66,62 +67,59 @@
...
+    libev_libs = driver_project_data["libev-libs"]
+    if is_macos:
+        libev_includes.extend(['/opt/homebrew/include', os.path.expanduser('~/homebrew/include')])
+        libev_libs.extend(['/opt/homebrew/lib'])
+    libev_ext = Extension('cassandra.io.libevwrapper',
+                          sources=['cassandra/io/libevwrapper.c'],
+                          include_dirs=libev_includes,
+                          libraries=['ev'],
+                          library_dirs=libev_libs,
+                          runtime_library_dirs=libev_libs)
```
absurdfarce: I discovered this option in the setuptools docs.  At the time I thought the underlying problem was that native shared objects were getting compiled without proper references to their dependencies, a situation which this config addresses exactly.  That wasn't actually the underlying problem, however, and since this config creates a problem on Windows builds (since Windows compilers apparently don't support anything like this) I'm gonna try a test build without it.  Assuming there isn't any regression I'm gonna argue to just keep it out.

Jenkinsfile
absurdfarce: A fair question @Dev-iL.  The Python versions listed here correspond to pre-installed Python versions on the Jenkins runner.  We don't want to pay the cost to download and install a runtime on every CI run so we pre-bake Python versions into the images Jenkins uses to run tests.

bschoening (approved):

setup.py
absurdfarce: Confirmed that build still looks good after removal of this option so we're going to go without it.

PR #1272, opened 2026-02-23
CASSPYTHON-4 Bump Python versions to match recent upgrades to the Jenkins runners

absurdfarce:


absurdfarce:
Jenkins runner has been updated so this should now be safe to merge.

PR #1274, opened 2026-02-26
Please don't merge this - but I wanted to let you know the bug tracker seems to be broken

ell1e:
I deeply apologize for this not being a reasonable pull request. It seems to be the only easy way to reach the developers, see pull request description. Please DO NOT MERGE this.

However, it seems like the bug tracker isn't working correctly, and won't let users report any issues:

Screenshot_20260226_215113

It also doesn't seem to be possible to reach the team via `info@datastix.com` to point anything out:

Screenshot_20260226_215547-1

My apologies if all of the above is actually intentional.

The reason I wanted to reach the bug tracker is to point out that setuptools versions that break [due to ez_setup](https://github.com/apache/cassandra-python-driver/commit/7d8015e3c1cff543a5f64c70cff3e14216e58037) seem to be out in the wild. Therefore, I simply wanted suggest that [a new release is made on pip](https://pypi.org/project/cassandra-driver/) so that the package can be installed again.

### DIFF
--- a/NONSENSENOTICE
+++ b/NONSENSENOTICE
@@ -1,0 +1,3 @@
+I deeply apologize for this not being a reasonable pull request.
+It seems to be the only easy way to reach the developers,
+see pull request description. Please DO NOT MERGE this.


absurdfarce:
Thanks for the PR @ell1e!

This project has recently been donated to the Apache Software Foundation (ASF).  You can find updated information for contacting the project on our [README](https://github.com/apache/cassandra-python-driver/blob/trunk/README.rst).  Among the information provided there is the location of our new [CASSPYTHON](https://issues.apache.org/jira/issues/?jql=project%20%3D%20CASSPYTHON%20ORDER%20BY%20key%20DESC) JIRA; that's the best place to file bugs going forward.

With respect to the ez_setup issue referenced above: the change to address that issue has been merged and will be included in the 3.30.0 release we're working on now.  I don't have a precise timeline for that release but I am hoping to have it out in the very near future.

Thanks again!

PR #1275, opened 2026-02-26
Remove DSE and HCD from the build matrix.

absurdfarce:
This is a temporary measure until ccm support for both servers is restored.


absurdfarce:
This change is no longer necessary; the issues around downloading HCD servers for testing have been addressed in another way.

PR #1283, opened 2026-03-23
support the tuple ip and port

wuzhongdehua:


cassandra/connection.py
```diff
@@ -237,18 +236,27 @@ def create(self, row):
 
         # create the endpoint with the translated address
         # TODO next major, create a TranslatedEndPoint type
+        ## Overriden by Pandu. using ip + port translation
+        translated_address = self.cluster.address_translator.translate(addr)
+        log.debug("PANDU: translated address {}".format(translated_address))
+        if type(translated_address) is tuple and len(translated_address)>1:
+            ip = translated_address[0]
+            port = translated_address[1]
+            log.debug( "PANDU: CREATING Default endpoint {} {}".format(ip, port))
```
absurdfarce, line 245: This debug line should also be removed

cassandra/connection.py
```diff
@@ -237,18 +236,27 @@ def create(self, row):
 
         # create the endpoint with the translated address
         # TODO next major, create a TranslatedEndPoint type
+        ## Overriden by Pandu. using ip + port translation
+        translated_address = self.cluster.address_translator.translate(addr)
+        log.debug("PANDU: translated address {}".format(translated_address))
```
absurdfarce, line 241: This debug line should be removed

cassandra/connection.py
```diff
@@ -237,18 +236,27 @@ def create(self, row):
...
+        if type(translated_address) is tuple and len(translated_address)>1:
```
absurdfarce, line 242: I think I'd prefer a namedtuple here rather than a straight tuple... just to avoid any possible confusion with positions in a base tuple.

cassandra/connection.py
```diff
@@ -237,18 +236,27 @@ def create(self, row):
 
         # create the endpoint with the translated address
         # TODO next major, create a TranslatedEndPoint type
+        ## Overriden by Pandu. using ip + port translation
+        translated_address = self.cluster.address_translator.translate(addr)
```
absurdfarce, line 240: It seems really odd to me that this changes the API from address -> translated address to address -> (translated address, translated port).  It's not immediately obvious to me how many use cases would be able to compute a translation for port numbers based only on an input address.  Seems like the more complete option would be to provide both address and port if we're going to return a combination of translated address and port.

This also probably helps out with the API.  We have something like this right now:

```
def translate(self, addr) -> str:
```

The proposal here is to move to something like this:

```
def translate(self, addr) -> (str | [tuple[str,int])
```

This is an API change but it's an addition rather than a change of existing functionality; we won't break any address translators out there.  But if we add a new method which clearly returns a host/string tuple (of some form) then we can clearly distinguish between these cases:

```
def translate(self, addr) -> str:
...
def translate_all(self, addr, port) -> tuple[str,int]
```

We'd still have the question of how we want DefaultEndPointFactory to use these methods... but it _might_ be clearer if we did so?

absurdfarce (changes requested):
Thanks for the PR @wuzhongdehua!

The main things for me on an initial pass would be a preference for namedtuples instead of conventional tuples and the comments around the API changes.

absurdfarce:
@bschoening I'm curious what you think about this one.

bschoening:
@wuzhongdehua can you add a description to this PR?  I don't understand why a tuple format is preferable and turns one line into ten.

absurdfarce:
I'm going to close this one out for now.  If you would like to re-open @wuzhongdehua we'd be happy to continue the conversation above... just let us know!

PR #1285, opened 2026-04-05
Add support for GRACEFUL_DISCONNECT events

AshenScribe:
[JIRA](https://issues.apache.org/jira/browse/CASSPYTHON-16)

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -3718,7 +3718,8 @@ class ControlConnection(object):
             connection.register_watchers({
                 "TOPOLOGY_CHANGE": partial(_watch_callback, self_weakref, '_handle_topology_change'),
                 "STATUS_CHANGE": partial(_watch_callback, self_weakref, '_handle_status_change'),
-                "SCHEMA_CHANGE": partial(_watch_callback, self_weakref, '_handle_schema_change')
+                "SCHEMA_CHANGE": partial(_watch_callback, self_weakref, '_handle_schema_change'),
+                "GRACEFUL_DISCONNECT": partial(_watch_callback, self_weakref, '_handle_graceful_disconnect')
             }, register_timeout=self._timeout)
 
             sel_peers = self._get_peers_query(self.PeersQueryType.PEERS, connection)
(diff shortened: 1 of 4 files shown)


omniCoder77:
@bschoening @absurdfarce requesting your review on this PR.

SiyaoIsHiding:
Thank you for the PR! 
Please add tests. Such a feature needs to be thoroughly tested

cassandra/cluster.py
```diff
@@ -3748,6 +3749,15 @@ def _try_connect(self, host):
 
         return connection
 
+    def _handle_graceful_disconnect(self, event):
+        if self._connection is None:
+            return
+        host = self._cluster.metadata.get_host(self._connection.endpoint)
+        print("host found:", host)
```
SiyaoIsHiding, line 3756: We need to remove those print statements before we merge the PR for sure

cassandra/connection.py
```diff
@@ -1061,16 +1064,29 @@ def get_request_id(self):
...
             except Exception:
                 log.exception("Pushed event handler errored, ignoring:")
 
+    def _handle_graceful_disconnect(self):
+        log.info("Received GRACEFUL_DISCONNECT from %s. Draining connection...", self.endpoint)
+        self.is_draining = True
+        self._socket_writable = False
+
+        with self.lock:
+            if self.in_flight == 0:
```
SiyaoIsHiding, line 1082: For the several occurrences of `if self.in_flight == 0:` in this PR,
If `self.in_flight == 0` is evaluated as false at that time, will the code _wait_ and evaluate `self.in_filght == 0` again?

cassandra/connection.py
```diff
@@ -1397,6 +1413,7 @@ def _handle_options_response(self, options_response):
                         locally_supported_compressions[compression_type]
 
         self._send_startup_message(compression_type, no_compact=self.no_compact)
+        self.supports_graceful_disconnect = options_response.options.get('GRACEFUL_DISCONNECT') == ['true']
```
SiyaoIsHiding, line 1416: Wondering where is the code that regsiters to listen on the GRACEFUL_DISCONNECT event?

SiyaoIsHiding (commented):
Thank you for the PR!
I'm not exactly knowledgeable about the python driver, but I do have some questions in my mind that I left in the comments.
We also need more testing. If the testing can not be automated as integration tests, then we should attach manual testing results in the PR to show that those timeout errors that we want to get rid of are indeed gone.
Also note that we should not merge the python driver PR before the server side CEP-59 implementation gets merged (and maybe more strictly, released).
So we will have plenty of time to polish this PR :)

PR #1292, opened 2026-05-04
Add warning about backslash corruption when building CQL with string formatting in the doc

howiezhao:
Update Getting Started documentation.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -215,6 +215,29 @@ Althought it is not recommended, you can also pass parameters to non-prepared
 statements. The driver supports two forms of parameter place-holders: positional
 and named.
 
+.. warning::
+
+    **Avoid "Round-Tripping" data using string formatting.**
+
+    Never use f-strings or ``%`` to insert data — especially driver-returned collections (maps, sets, lists) — back into a CQL query.
+
+    **The Gotcha:** The driver's collection objects use Python's ``repr()`` formatting for nested values.
+    This automatically doubles backslashes (e.g., ``\`` becomes ``\\``).
+    Because CQL does not use backslashes as escape characters,
+    Cassandra will store those extra backslashes literally, **corrupting your data**.
+
+    **The Fix:** Always use **prepared statements**.
+    They transmit data in a binary format that bypasses Python's string serialization entirely.
+
+    .. code-block:: python
+
+        # BAD: f-strings cause double-escaping/corruption
+        session.execute(f"UPDATE t SET my_map={row.my_map} WHERE id=1")
+
+        # GOOD: Prepared statements preserve data exactly
+        stmt = session.prepare("UPDATE t SET my_map=? WHERE id=1")
+        session.execute(stmt, [row.my_map])
+
 Positional parameters are used with a ``%s`` placeholder.  For example,
 when you execute:
 


bschoening:
@howiezhao Does the problem occur when you round-trip data?  Read from the database, and then insert it back into the database?  For the getting started docs, that might be a bit of a corner case.

howiezhao:
> @howiezhao Does the problem occur when you round-trip data? Read from the database, and then insert it back into the database? For the getting started docs, that might be a bit of a corner case.

Hi @bschoening yes, if the map data contains backslashes, reading it in and writing it using Python string formatting will cause backslashes to be added, as I mentioned in the example, so I recommend using a prepared statement.

bschoening:
Would prefer something more concise, maybe something like...

`.. warning::

**Avoid "Round-Tripping" data using string formatting.**

Never use f-strings or ``%`` to insert data—especially driver-returned collections (maps, sets, lists)—back into a CQL query. 

**The Gotcha:** The driver's collection objects use Python's ``repr()`` formatting for nested values. This automatically doubles backslashes (e.g., ``\`` becomes ``\\``). Because CQL does not use backslashes as escape characters, Cassandra will store those extra backslashes literally, **corrupting your data.**

**The Fix:** Always use **prepared statements**. They transmit data in a binary format that bypasses Python's string serialization entirely.

.. code-block:: python

    # BAD: f-strings cause double-escaping/corruption
    session.execute(f"UPDATE t SET my_map={row.my_map} WHERE id=1")

    # GOOD: Prepared statements preserve data exactly
    stmt = session.prepare("UPDATE t SET my_map=? WHERE id=1")
    session.execute(stmt, [row.my_map])`

howiezhao:
> Would prefer something more concise, maybe something like...
> 
> `.. warning::
> 
> **Avoid "Round-Tripping" data using string formatting.**
> 
> Never use f-strings or `%` to insert data—especially driver-returned collections (maps, sets, lists)—back into a CQL query.
> 
> **The Gotcha:** The driver's collection objects use Python's `repr()` formatting for nested values. This automatically doubles backslashes (e.g., `\` becomes `\\`). Because CQL does not use backslashes as escape characters, Cassandra will store those extra backslashes literally, **corrupting your data.**
> 
> **The Fix:** Always use **prepared statements**. They transmit data in a binary format that bypasses Python's string serialization entirely.
> 
> .. code-block:: python
> 
> ```
> # BAD: f-strings cause double-escaping/corruption
> session.execute(f"UPDATE t SET my_map={row.my_map} WHERE id=1")
> 
> # GOOD: Prepared statements preserve data exactly
> stmt = session.prepare("UPDATE t SET my_map=? WHERE id=1")
> session.execute(stmt, [row.my_map])`
> ```

Hi @bschoening good suggestion, updated

bschoening (approved):

howiezhao:
Hi @bschoening what else do I need to do to move this PR to be merged?

bschoening:
@howiezhao it just needs a committer with the time available to commit it.

docs/getting_started.rst
```diff
@@ -215,6 +215,29 @@ Althought it is not recommended, you can also pass parameters to non-prepared
 statements. The driver supports two forms of parameter place-holders: positional
 and named.
 
+.. warning::
+
+    **Avoid "Round-Tripping" data using string formatting.**
```
absurdfarce, line 220: This seems far too aggressive to me.  There are plenty of situations in which session.execute() is used with f-strings, format() or % substitution.  I note, for example, that the following works just fine:

```python

cluster = Cluster()
session = cluster.connect()

alist = ["foo", "bar"]
amap = {"foo": "bar", "baz": "qux"}
aset = set(["foo", "bar"])
session.execute(‘delete from backslash.foo where i = 1’)
session.execute(f"insert into backslash.foo (i, alist, amap, aset) values (1,{alist}, {amap}, {aset})")
for row in session.execute("select i, alist, amap, aset from backslash.foo where i = 1"):
	print(row[0],row[1],row[2],row[3])
```

```
1 ['foo', 'bar'] {'baz': 'qux', 'foo': 'bar'} SortedSet(['bar', 'foo'])
```

I don't think we want to tell people not to use a fairly useful element of functionality because of potential escaping issues.  Seems like we'd be better off describing what can happen and letting users form their own conclusions.

docs/getting_started.rst
```diff
@@ -215,6 +215,29 @@ Althought it is not recommended, you can also pass parameters to non-prepared
...
+
+    Never use f-strings or ``%`` to insert data — especially driver-returned collections (maps, sets, lists) — back into a CQL query.
+
+    **The Gotcha:** The driver's collection objects use Python's ``repr()`` formatting for nested values.
```
absurdfarce, line 224: This is actually incorrect.  The driver converts strings to bytes (and back again) by [encoding and decoding them using UTF-8](https://github.com/apache/cassandra-python-driver/blob/3.30.0/cassandra/cqltypes.py#L773-L787).  We don't explicitly call repr() at any point in that process.  repr() is indeed the problem... but it's not coming from the driver.

Consider the following simple variant on my code sample above:

```python

cluster = Cluster()
session = cluster.connect()

alist = ["f\oo", "ba\z"]
amap = {"f\oo": "ba\z", "ba\z": "q\ox"}
aset = set(["fo\o", "b\oz"])
session.execute('delete from backslash.foo where i = 1')
session.execute(f"insert into backslash.foo (i, alist, amap, aset) values (1,{alist}, {amap}, {aset})")
for row in session.execute("select i, alist, amap, aset from backslash.foo where i = 1"):
	print(row[0],row[1],row[2],row[3])
```

```
1 ['f\\\\oo', 'ba\\\\z'] {'ba\\\\z': 'q\\\\ox', 'f\\\\oo': 'ba\\\\z'} SortedSet(['b\\\\oz', 'fo\\\\o'])
```

This seems to manifest the problematic behaviour, right?  Well, kinda.  When we go to print the list (row[1] in our example) in our result set Python has to call repr() in order to get to a string it can print... which is part of our problem.  Just avoiding that step gets us haflway there:

```python

cluster = Cluster()
session = cluster.connect()

alist = ["f\oo", "ba\z"]
amap = {"f\oo": "ba\z", "ba\z": "q\ox"}
aset = set(["fo\o", "b\oz"])
session.execute('delete from backslash.foo where i = 1')
session.execute(f"insert into backslash.foo (i, alist, amap, aset) values (1,{alist}, {amap}, {aset})")
for row in session.execute("select i, alist, amap, aset from backslash.foo where i = 1"):
	f,b = row[1]
	print(row[0],f,b,row[2],row[3])
```

```
1 f\\oo ba\\z {'ba\\\\z': 'q\\\\ox', 'f\\\\oo': 'ba\\\\z'} SortedSet(['b\\\\oz', 'fo\\\\o'])
```

Okay, so what about the extra escaping that's still being added there?  That's also coming from the use of f-strings in the client, this time when we create the query string we're sending to the server.  When the f-string containing our insert statement is evaluated the same thing occurs; we're being passed a list with the "alist" param so Python has to call repr() on it to generate a string, which leads to the remaining escaping.  We can avoid that by manually generating the list syntax ourselves:

```python

cluster = Cluster()
session = cluster.connect()

alist = ["f\oo", "ba\z"]
amap = {"f\oo": "ba\z", "ba\z": "q\ox"}
aset = set(["fo\o", "b\oz"])
alist_cql = "['{}']".format("','".join(alist))
session.execute('delete from backslash.foo where i = 1')
session.execute(f"insert into backslash.foo (i, alist, amap, aset) values (1,{alist_cql}, {amap}, {aset})")
for row in session.execute("select i, alist, amap, aset from backslash.foo where i = 1"):
	f,b = row[1]
	print(row[0],f,b,row[2],row[3])
```

```
1 f\oo ba\z {'ba\\\\z': 'q\\\\ox', 'f\\\\oo': 'ba\\\\z'} SortedSet(['b\\\\oz', 'fo\\\\o'])
```

absurdfarce: To be very clear: I'm not suggesting doing nothing.  I'm fine with adding some language here warning users to be careful about using string escapes when working with collections (or any container type really, anything that Python has to call repr() on) when using parameters with non-prepared statements.  We can even provide some variant of the above as a concrete example to make it clearer to users.

docs/getting_started.rst
```diff
@@ -215,6 +215,29 @@ Althought it is not recommended, you can also pass parameters to non-prepared
...
+    This automatically doubles backslashes (e.g., ``\`` becomes ``\\``).
+    Because CQL does not use backslashes as escape characters,
+    Cassandra will store those extra backslashes literally, **corrupting your data**.
```
absurdfarce, line 227: This also seems much too strong.  I suppose it does wind up "corrupting" your data here but the suggestion seems to be that data will always be corrupted if these interfaces are used... and that's just not the case.

docs/getting_started.rst
```diff
@@ -215,6 +215,29 @@ Althought it is not recommended, you can also pass parameters to non-prepared
...
+
+    **The Fix:** Always use **prepared statements**.
+    They transmit data in a binary format that bypasses Python's string serialization entirely.
```
absurdfarce, line 230: I'm fine with the recommendation to use prepared statements but the explanation isn't really correct (for reasons outlined above).

More importantly this section is being added directly after [a section which made the same recommendation](https://github.com/apache/cassandra-python-driver/blob/3.30.0/docs/getting_started.rst#L178-L180).  I'd rather see the addition of some text to add warnings about potential escaping issues to the existing docs for users who wish to make use of parameters of non-prepared statements rather than telling them to use something else.

absurdfarce (changes requested):
Good find @howiezhao!  I think we need to clarify our messaging here to provide our users clear context about when they should be aware of this behaviour.  I've added some details in my review... let me know what you think!

PR #1295, opened 2026-05-26
CASSPYTHON-18: removed obsolete checks for python version > 3.7+

bschoening:
A recreated version of #1290 (which was closed accidentally)


cassandra/cluster.py
```diff
@@ -1424,26 +1424,26 @@ def _create_thread_pool_executor(self, **kwargs):
...
+                from futurist import GreenThreadPoolExecutor
+                tpe_class = GreenThreadPoolExecutor
+            except ImportError:
+                # futurist is not available
+                raise ImportError(
+                    ("Python 3.7+ and Eventlet cause the `concurrent.futures.ThreadPoolExecutor` "
+                     "to hang indefinitely. If you want to use the Eventlet reactor, you "
+                     "need to install the `futurist` package to allow the driver to use "
+                     "the GreenThreadPoolExecutor. See https://github.com/eventlet/eventlet/issues/508 "
+                     "for more details."))
```
absurdfarce: So I think this was the only point left from our conversation on the prior PR @bschoening.  Based on [my earlier comment](https://github.com/apache/cassandra-python-driver/pull/1290/changes#r3157196338) I'm inclined to say we should just remove this check (and executor class set operation) completely.

Whaddya think?

tests/integration/standard/test_metadata.py
```diff
@@ -1639,12 +1631,12 @@ def test_function_cql_called_on_null(self):
...
         with self.VerifiedFunction(self, **kwargs) as vf:
             fn_meta = self.keyspace_function_meta[vf.signature]
-            self.assertRegex(fn_meta.as_cql_query(), "CREATE FUNCTION.*\) CALLED ON NULL INPUT RETURNS .*")
+            self.assertRegex(fn_meta.as_cql_query(), "CREATE FUNCTION.*) CALLED ON NULL INPUT RETURNS .*")
 
         kwargs['called_on_null_input'] = False
         with self.VerifiedFunction(self, **kwargs) as vf:
             fn_meta = self.keyspace_function_meta[vf.signature]
-            self.assertRegex(fn_meta.as_cql_query(), "CREATE FUNCTION.*\) RETURNS NULL ON NULL INPUT RETURNS .*")
+            self.assertRegex(fn_meta.as_cql_query(), "CREATE FUNCTION.*) RETURNS NULL ON NULL INPUT RETURNS .*")
```
absurdfarce: I'm not trying to remember where this change (and the similar change above) came from.  These tests are failing on Jenkins due to these changes... and they're failing when I run the test locally as well.

I think I made these changes and I think I convinced myself that they were okay via local runs... but something went wrong somewhere along the way.

absurdfarce: Oh, I think I get it... they just need to be raw strings. Working on fixing this now.

absurdfarce:
Jenkins run for this appears clean (or at least clean of any regressions).  @bschoening are you good with this one as it stands?  I think this is the last thing we wanted to get in for 3.30.1.

tests/integration/standard/test_metadata.py
```diff
@@ -1087,14 +1087,6 @@ def test_export_keyspace_schema_udts(self):
...
-        if PROTOCOL_VERSION < 3:
-            raise unittest.SkipTest(
-                "Protocol 3.0+ is required for UDT change events, currently testing against %r"
-                % (PROTOCOL_VERSION,))
-
-        if sys.version_info[0:2] != (2, 7):
-            raise unittest.SkipTest('This test compares static strings generated from dict items, which may change orders. Test with 2.7.')
-
         cluster = TestCluster()
         session = cluster.connect()
```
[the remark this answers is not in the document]
absurdfarce: This is kinda the point.  We no longer test with anything less than 3.x so _having_ this constraint makes no sense.

tests/unit/test_row_factories.py
```diff
@@ -24,12 +24,8 @@
 
 from unittest import TestCase
 
-
 log = logging.getLogger(__name__)
 
-
-NAMEDTUPLE_CREATION_BUG = sys.version_info >= (3,) and sys.version_info < (3, 7)
-
 class TestNamedTupleFactory(TestCase):
```
bschoening: Based on the Co-pilot comment, it seems like the whole test function can be dropped now (post 3.7)

tests/unit/test_row_factories.py
```diff
@@ -49,37 +45,16 @@ class TestNamedTupleFactory(TestCase):
...
-        self.assertEqual(len(w), 1)
-        warning = w[0]
-        self.assertIn('pseudo_namedtuple_factory', str(warning))
-        self.assertIn('3.7', str(warning))
-
-        for r in rows:
-            self.assertEqual(r.col0, self.long_rows[0][0])
+        named_tuple_factory(self.long_colnames, self.long_rows)
 
     def test_creation_no_warning_on_short_column_list(self):
```
bschoening: Can this be dropped also?

tests/unit/test_row_factories.py
```diff
@@ -24,12 +24,8 @@
 
 from unittest import TestCase
```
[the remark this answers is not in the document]
absurdfarce: Agreed, removed

tests/unit/test_row_factories.py
```diff
@@ -49,37 +45,16 @@ class TestNamedTupleFactory(TestCase):
...
-        with warnings.catch_warnings(record=True) as w:
-            rows = named_tuple_factory(self.long_colnames, self.long_rows)
-        self.assertEqual(len(w), 1)
-        warning = w[0]
-        self.assertIn('pseudo_namedtuple_factory', str(warning))
-        self.assertIn('3.7', str(warning))
-
-        for r in rows:
-            self.assertEqual(r.col0, self.long_rows[0][0])
+        named_tuple_factory(self.long_colnames, self.long_rows)
```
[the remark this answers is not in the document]
absurdfarce: I disagree that this is a no-op as it stands.  The underlying issue with [PYTHON-893](https://datastax-oss.atlassian.net/browse/PYTHON-893) was that namedtuple() would throw an exception if it were passed more than 255 parameters.  unittest will interpret an unhandled exception from a test case as a failure of that test case so just making the call (with appropriately contrived arguments) should be enough to trigger the error _if_ it occurs.

Now, I'm _perfectly_ willing to concede that the error _shouldn't_ occur for versions of Python we're actually testing against since this issue has been resolved in the Python runtime (as referenced in the comments).  So what we're guarding against here is a regression in how the Python runtime handles namedtuple() with lots of args.  If we decide we aren't worried about that we can absolutely remove this test... but if we are I don't think we need to go through the extra steps suggested by Copilot.

tests/unit/test_row_factories.py
absurdfarce: This is actually testing a different case; we're validating proper functionality for a shorter (arguably "normal") length column list.  I'd argue for something similar to what I said above; if we decide we aren't worried about regressions in the handling of namedtuple args we can delete this file in it's entirety.  I'm not arguing for or against it necessarily... I can see arguments both ways.

tests/unit/test_row_factories.py
absurdfarce: Discussed below.  Fundamentally this comes down to whether we want to test for regressions or not.

absurdfarce:
Mentioned to @bschoening in other communication but I'll make a note of it here too; if we remove the test cases above (because Python runtimes we support no longer have this issue) we should probably remove [the workaround](https://github.com/apache/cassandra-python-driver/commit/16ae36b52cad2bd9e96cd60c5ab5ce3ffb47f809) added for PYTHON-893 as well.

tests/unit/test_row_factories.py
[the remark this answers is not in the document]
bschoening: Since this Cython limitation is now explicitly tested upstream (per [Python issue18896](https://bugs.python.org/issue18896)) in **test_collections.py**, we probably don't need to duplicate it here. It might just become technical debt/clutter down the road. If you feel strongly about keeping it, though, I'm okay with it!

```
diff -r 79d52e7d472d -r 7454ca88aacb Lib/test/test_collections.py
--- a/Lib/test/test_collections.py	Fri Dec 16 19:06:51 2016 +0200
+++ b/Lib/test/test_collections.py	Fri Dec 16 19:19:02 2016 +0200
@@ -319,8 +319,7 @@
         self.assertEqual(Dot(1)._replace(d=999), (999,))
         self.assertEqual(Dot(1)._fields, ('d',))
 
-        # n = 5000
-        n = 254 # SyntaxError: more than 255 arguments:
+        n = 5000
         names = list(set(''.join([choice(string.ascii_letters)
                                   for j in range(10)]) for i in range(n)))
```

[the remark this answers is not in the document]
absurdfarce: An entirely reasonable point.

I also just realized that we're already doing something different _in this PR_!  The rationale for removing the warning about the thread pool impl when using eventlet was that eventlet had already fixed the underlying issue so new versions of the driver wouldn't need to worry about older versions.  It's a pretty straightforward parallel; if we do it there we probably should do the same in this case.

I have a _slight_ concern for folks upgrading existing applications which may be working off of a dated set of dependencies... that could potentially undermine the approach I argued for with eventlet.  But it's going away in the next release anyway (and isn't used much now) so... I'm not super-worried about it.

Regardless, I think we're in agreement on this test; it can just be removed.  I'm also going to remove the check logic that was put in place as part of PYTHON-893.

absurdfarce:
Okay, I believe I have all the changes we've agreed upon here @bschoening.  I've kicked off a Jenkins build for this PR just to make sure there aren't any unexpected regressions; I'll report status here when it's done!

Note that we're effectively reverting the change for PYTHON-893 in this PR... we prolly should make a note of that in the release notes when we get to that point.

absurdfarce (approved):

absurdfarce:
In separate communication confirmed @bschoening is 👍 to the current state of this PR.  We're calling this a wrap and moving forward with 3.30.1!

absurdfarce:
Jenkins run _after_ rebase-and-squash looks good, calling this done!

PR #1300, opened 2026-07-15
PEP-8 code style changes

bschoening:
Mostly whitespace fixes for PEP-8

### DIFF
--- a/cassandra/cluster.py
+++ b/cassandra/cluster.py
@@ -98,21 +98,24 @@ try:
 except ImportError:
     from cassandra.util import WeakSet  # NOQA
 
+
 def _try_libev_import():
     try:
         from cassandra.io.libevreactor import LibevConnection
-        return (LibevConnection,None)
+        return (LibevConnection, None)
     except DependencyException as e:
         return (None, e)
+
 
 def _try_asyncore_import():
     try:
         from cassandra.io.asyncorereactor import AsyncoreConnection
-        return (AsyncoreConnection,None)
+        return (AsyncoreConnection, None)
     except DependencyException as e:
         return (None, e)
 
-def _connection_reduce_fn(val,import_fn):
+
+def _connection_reduce_fn(val, import_fn):
     (rv, excs) = val
     # If we've already found a workable Connection class return immediately
     if rv:
@@ -122,10 +125,11 @@ def _connection_reduce_fn(val,import_fn):
         excs.append(exc)
     return (rv or import_result, excs)
 
+
 log = logging.getLogger(__name__)
 
 conn_fns = (_try_libev_import, _try_asyncore_import)
-(conn_class, excs) = reduce(_connection_reduce_fn, conn_fns, (None,[]))
+(conn_class, excs) = reduce(_connection_reduce_fn, conn_fns, (None, []))
 if not conn_class:
     raise DependencyException("Unable to load a default connection class", excs)
 DefaultConnection = conn_class
@@ -383,8 +387,8 @@ class ExecutionProfile(object):
 
         self.retry_policy = retry_policy or RetryPolicy()
 
-        if (serial_consistency_level is not None and
-                not ConsistencyLevel.is_serial(serial_consistency_level)):
+        if (serial_consistency_level is not None
+                and not ConsistencyLevel.is_serial(serial_consistency_level)):
             raise ValueError("serial_consistency_level must be either "
                              "ConsistencyLevel.SERIAL "
                              "or ConsistencyLevel.LOCAL_SERIAL.")
@@ -675,6 +679,7 @@ class Cluster(object):
         self._auth_provider = value
 
     _load_balancing_policy = None
+
     @property
     def load_balancing_policy(self):
         """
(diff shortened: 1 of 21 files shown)


cassandra/cluster.py
```diff
@@ -768,9 +774,8 @@ def default_retry_policy(self, policy):
     Using ssl_options without ssl_context is deprecated and will be removed in the
     next major release.
 
-    An optional dict which will be used as kwargs for ``ssl.SSLContext.wrap_socket`` 
-    when new sockets are created. This should be used when client encryption is enabled 
-    in Cassandra.
+    An optional dict which will be used as kwargs for ``ssl.SSLContext.wrap_socket``when new
+    sockets are created. This should be used when client encryption is enabled in Cassandra.
```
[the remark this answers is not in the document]
bschoening: fixed

cassandra/cqltypes.py
```diff
@@ -1006,9 +1020,11 @@ def evict_udt_class(cls, keyspace, udt_name):
 
     @classmethod
     def apply_parameters(cls, subtypes, names):
-        keyspace = subtypes[0].cass_parameterized_type()  # when parsed from cassandra type, the keyspace is created as an unrecognized cass type; This gets the name back
+        # when parsed from cassandra type, the keyspace is created as an unrecognized cass type; This resolves the name
```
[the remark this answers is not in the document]
absurdfarce: "This" is still capitalized as it stands right now which seems wrong when it's coming after the semi-colon.

cassandra/cqltypes.py
```diff
@@ -1006,9 +1020,11 @@ def evict_udt_class(cls, keyspace, udt_name):
...
+        keyspace = subtypes[0].cass_parameterized_type()
         udt_name = _name_from_hex_string(subtypes[1].cassname)
-        field_names = tuple(_name_from_hex_string(encoded_name) for encoded_name in names[2:])  # using tuple here to match what comes into make_udt_class from other sources (for caching equality test)
+        # tuple used to match what comes into make_udt_class from other sources (for caching equality test)
```
absurdfarce: Nit: perhaps just "tupled matches what comes into..."; seems like that makes things a bit more readable?

cassandra/metadata.py
```diff
@@ -789,16 +788,16 @@ def export_as_string(self):
...
+            + self.user_type_strings()
+            + [f.export_as_string() for f in self.functions.values()]
+            + [a.export_as_string() for a in self.aggregates.values()]
+            + [t.export_as_string() for t in tables_with_vertex + other_tables])
 
         if self._exc_info:
             import traceback
             ret = "/*\nWarning: Keyspace %s is incomplete because of an error processing metadata.\n" % \
-                  (self.name)
+                  self.name
```
absurdfarce, line 800: This... looked odd to me 'cause I was pretty sure the % operator required a tuple as it's arg... but apparently there's a special case for single values (based on [the docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) anyway).  I'm kinda wondering if maybe it isn't worth our while to move away from the % operator anyway but I'm very receptive to an argument that says this PR isn't the place to do that.

We can consider another ticket to move to f-strings (or perhaps some other option) uniformly throughout the code base but it's definitely not worth side-tracking this effort to do so here.

cassandra/util.py
```diff
@@ -1692,15 +1709,17 @@ def __repr__(self):
...
 
+
 @total_ordering
 class Version(object):
     """
     Representation of a Cassandra version.  Mostly follows the implementation of the same logic in the Java driver;
     see https://github.com/apache/cassandra-java-driver/blob/4.19.2/core/src/main/java/com/datastax/oss/driver/api/core/Version.java.
 
-    Cassandra versions are assumed to correspond to major.minor.patch with an optional additional numeric build field as well as a
+    Cassandra's versions are assumed to correspond to major.minor.patch with an optional additional numeric build field as well as a
```
absurdfarce: I'd actually argue the text above reads more clearly but I don't feel super-strong about it or anything

absurdfarce (approved):
A few small comments but all of them are nits.  I'm perfectly fine merging this exactly as it stands @bschoening ... I'll leave it up to you to take a look at the review and tell me what you want to do.

cassandra/cqltypes.py
bschoening: agreed, although the function name should stay 'tuple'

cassandra/util.py
bschoening: It was a little wordy, I reworded it "Versions correspond to major.minor.patch with an optional additional numeric build field and string prerelease field."

cassandra/metadata.py
bschoening: yes, f-strings would be a nice change for later

cassandra/util.py
absurdfarce: Sweet... I'm quite good with what you have in there now.

absurdfarce (approved):

PR #1301, opened 2026-07-15
CASSPYTHON-24 Remove Insights support

absurdfarce:


cassandra/cluster.py
```diff
@@ -31,10 +31,8 @@
 import logging
 from warnings import warn
 from random import random
-import re
 import queue
 import socket
-import sys
```
absurdfarce, line 37: A drive-by fix.  I noticed these two imports were unused while working on CASSPYTHON-13 with Brett, figured I'd clean them up next time I was in this code.

cassandra/cluster.py
```diff
@@ -1170,14 +1130,14 @@ def __init__(self,
...
-
         if contact_points and (not self.endpoints_resolved):
             # only want to raise here if the user specified CPs but resolution failed
-            raise UnresolvableContactPoints(self._endpoint_map_for_insights)
+            endpoint_map = {repr(ep): '{ip}:{port}'.format(ip=ep.address, port=ep.port) for ep in self.endpoints_resolved}
+            endpoint_map.update(
+                {key: ['{ip}:{port}'.format(ip=ip, port=port) for ip, port in value]
+                 for key, value in strs_resolved_map.items() if value is not None}
+            )
+            raise UnresolvableContactPoints(endpoint_map)
```
absurdfarce, line 1140: Seems weird to me that we were re-using an Insights map in the exception that's thrown here.. but apparently we were.  Moved the logic into this block (so that it's only executed if an exception actually occurs) and moved everything out of class state.

cassandra/cluster.py
```diff
@@ -2473,8 +2424,7 @@ def default_serial_consistency_level(self, cl):
 
     session_id = None
     """
-    A UUID that uniquely identifies this Session to Insights. This will be
-    generated automatically.
+    A UUID that uniquely identifies this Session. This will be generated automatically.
```
absurdfarce, line 2427: I had originally planned on removing this entirely but this ID is also used in column-level encryption (CLE)

absurdfarce:
Forgot to mention: Jenkins run on this change looks pretty good

bschoening (approved):

bschoening:
+1

